PR #2, opened 2018-02-25
Fixed some keyboard bindings

bitionaire:
* Arrow up/down won't switch focused card, while in edit mode or a modal is opened


src/components/Column/Column.container.ts
```diff
@@ -119,31 +119,37 @@ export const mapStateToProps = (
...
-          onFocusCard(null);
-          ev.preventDefault();
-        }
-
-        // on arrow up
-        if (ev.keyCode === Key.UpArrow) {
-          if (index > 0) {
-            onFocusCard(cards[index - 1].id as string);
+    if (state.app.keyboardNavigationEnabled) {
+      window.onkeydown = (ev: KeyboardEvent) => {
```
dotcs: Not sure, but if the global `window.onkeydown` method is set here, isn't it overwritten three times when there are three columns? Shouldn't this be defined the `ColumnView` component or higher in the hierarchy?

bitionaire: Just added `focused.type === ownProps.type` to the outer condition check. Key binding should only be set on focused column now. Good catch! 👍

PR #11, opened 2018-03-02
Refactored <Action /> component

bitionaire:


src/components/Column/FocusedCardComponent/FocusedCardComponent.tsx
```diff
@@ -62,7 +62,7 @@ class FocusedCardComponent extends React.Component<
   };
 
   render() {
-    let theme = 'dark';
+    let theme: 'dark' | 'light' = 'dark';
```
dotcs: I guess this should be of type `ActionTheme`, isn't it?

bitionaire: Ehm.. you must have had and outdated copy of the code.. :trollface:

dotcs: You don't say ... ;)

PR #117, opened 2020-04-09
Setup basic CI/CD

### LINKED ISSUE #105: Setup CI/CD
Auto build of master and deployment to Firebase STAGING

theexiile1305:
closes #105


theexiile1305:
@bitionaire Shall we change the node package manager to `yarn`. This would increase the building time and unexpected behaviours, because the `package-lock.json` file will not be crated on every installation. Instead of this, the yarn-lock.json file is will be kept with the command `yarn install --forzen-lockfile`.

theexiile1305:
fixes #113

theexiile1305:
@bitionaire The production build of the react app doesn't work right now. It seems that there is a false api-key given `{code: "auth/invalid-api-key", message: "Your API key is invalid, please check you have copied it correctly.", a: null}.` . 
This happens locally and on the CI, even though all environments variables are set.
Do you have an idea?

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM node:13.8-alpine3.11 as builder
+
+# Stage 1 - the build process
+WORKDIR /usr/src/app
+COPY ["package.json", "package-lock.json", "./"]
+RUN npm install
```
bitionaire, line 6: Dependencies shouldn't be installed within the Docker container. Instead the packaged version should be copied here from the CI tasks result. 
A `npm install` run might change the result on every run, since dependencies can be overwritten on npmjs

.github/workflows/staging.yml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          args: deploy --project ${{ secrets.REACT_APP_FIREBASE_PROJECT_ID }}
+        env:
+          FIREBASE_TOKEN: ${{ secrets.FIREBASE_TOKEN }}
+          REACT_APP_FIREBASE_API_KEY: ${{ secrets.REACT_APP_FIREBASE_API_KEY }}
+          REACT_APP_FIREBASE_APP_ID: ${{ secrets.REACT_APP_FIREBASE_APP_ID }}
+          REACT_APP_FIREBASE_MEASUREMENT_ID: ${{ secrets.REACT_APP_FIREBASE_MEASUREMENT_ID }}
+          REACT_APP_MESSAGING_SENDER_ID: ${{ secrets.REACT_APP_MESSAGING_SENDER_ID }}
+          REACT_APP_FIREBASE_PROJECT_ID: ${{ secrets.REACT_APP_FIREBASE_PROJECT_ID }}
+  deploy_github:
```
bitionaire, line 59: I'm not sure if we need a GitHub package. Do you see any benefits from this?

.firebaserc
```diff
@@ -0,0 +1,5 @@
+{
+  "projects": {
+    "default": "scrumlr-dev-530d9"
```
bitionaire, line 3: Maybe we should add a .firebaserc with a dummy project name like `<YOUR_FIREBASE_PROJECT>` and add this file to .gitignore, after it was initially added?

.github/workflows/staging.yml
```diff
@@ -0,0 +1,79 @@
+name: staging-CI
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
bitionaire, line 6: This pipeline should only execute on changes of the `master`. I'm haven't worked with GitHub actions yet, but I suspect this line triggers the CI on each PR on the master

.github/workflows/staging.yml
```diff
@@ -0,0 +1,79 @@
...
+        node: [ 13 ]
+        os: [ ubuntu-latest ]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: npm install
+      - name: Building produciton version
+        run: npm run build
```
bitionaire, line 47: I would add the `install` and `build` steps to the build pipeline. Seems more fitting, and these steps can also be executed for the `dev` CI. The tests might pass and yet the build might fail in some rare cases.
Yet I'm not an expert on GitHub actions and maybe you added this here, because resources (e.g. build artifacts) cannot be shared between different steps in the pipeline. The documentation isn't as great so I couldn't research this quickly.

bitionaire:
> @bitionaire The production build of the react app doesn't work right now. It seems that there is a false api-key given `{code: "auth/invalid-api-key", message: "Your API key is invalid, please check you have copied it correctly.", a: null}.` .
> This happens locally and on the CI, even though all environments variables are set.
> Do you have an idea?

Seems to be fixed, right?

bitionaire:
> @bitionaire Shall we change the node package manager to `yarn`. This would increase the building time and unexpected behaviours, because the `package-lock.json` file will not be crated on every installation. Instead of this, the yarn-lock.json file is will be kept with the command `yarn install --forzen-lockfile`.

You can use `npm ci` instead of `install` as this command will ignore the `package.json` and only install the dependencies from the `package-lock.json`. (-> https://docs.npmjs.com/cli/ci.html)
But I'm open to use `yarn` if you prefer it. In the old days we also used `yarn`, but I removed it since NPM 5 integrated many features of it. The performance of `yarn` is slightly better. As a previous Java developer I'm used to long running builds and don't care that much about it :D

.firebaserc
theexiile1305: I'v don it.

.github/workflows/staging.yml
theexiile1305: I only used it, to verify, that the pipe is successfully executed. I also removed it now.

.github/workflows/staging.yml
theexiile1305: The first approach on caching the dependencies from the documentation didn't work so well. It take much more time (15 min) than the clean installation. Maybe I wanna find some performance increase.

.github/workflows/staging.yml
theexiile1305: I just tried it out. Should I remove the publishing?

Issue #106, opened 2020-04-08
Add voting mechanism on cards


bitionaire:
- Add buttons to start, stop and reset voting on cards for board members
- Add configurable overall vote limit to board admin / moderator
- Add option to toggle visibility of votes before voting phase is completed (when voting is completed, all votes will be shown anyway)
- Add option to toggle multivotes (multiple votes on one card)

bitionaire:
As we are progressing towards the release of V2 further work on the current version will be stopped.

PR #120, opened 2020-04-10
#106 Voting

bitionaire:


firestore.rules
```diff
@@ -16,22 +22,30 @@ service cloud.firestore {
...
 
-            match /applicants/{uid} {
+            match /pending/{uid} {
                 allow read: if isAdmin()
                 allow create: if isAdmin() || request.auth.uid == uid
                 allow update: if isAdmin()
                 allow delete: if isAdmin() || request.auth.uid == uid
             }
 
             match /members/{uid} {
```
bitionaire, line 37: Do you think we should check the vote limits on the DB side or just leave this logic to the clients? 
These checks (and the necessary code) do have a negative effect on performance and costs. It would be much easier to use the Firestore functions 'increment' and 'decrement' instead and handle everything on the clients.

theexiile1305: Yea, it would be a nice approach! 👍

PR #157, opened 2020-09-02
Move Delete Button to Settings

alphapfote:


src/components/Modal/variant/SettingsModal.tsx
```diff
@@ -67,6 +77,23 @@ export class SettingsModal extends React.Component<SettingsModalProps, {}> {
           defaultValue={this.props.username}
           className="settings-modal__input"
         />
+
+        {this.props.isAdmin && (
+          <button
+            className="settings-modal__delete-button"
+            aria-label="Delete the scrumlr board"
```
bitionaire, line 84: `Delete this board`?

src/components/Modal/variant/SettingsModal.scss
```diff
@@ -35,4 +35,34 @@
...
+
+  font-size: $theme-font-size;
+  font-weight: bold;
+  text-transform: uppercase;
+
+  padding: $default-padding $large-padding;
+
+  margin-top: $large-margin;
+  color: $color-white;
+  background: $color-mint;
```
bitionaire, line 54: Maybe you can introduce another color, e.g. an orange to to indicate the danger of this action? There's an already defined `$color-orange` variable and used for the timer :)
Or you could change it in a grayish color tone and set the orange for `focus` and `hover` states as defined in line 66.

src/components/Modal/variant/SettingsModal.scss
```diff
@@ -35,4 +35,34 @@
...
+
+  transition: background .2s ease-in-out, box-shadow .2s ease-in-out;
+
+  .delete-button__icon {
```
bitionaire, line 58: The CSS guru Matthias Reuter taught me not to use nested selectors in SCSS, because it's easier to read when everything is defined in one line and also easier to `grep` / find those selectors. 
We can leave it here, as this version of the app will not be maintained in the future :)

bitionaire (changes requested):
Basically I would like another button background color to indicate the danger of the delete operation, as commented in the sass

src/components/Modal/variant/SettingsModal.tsx
alphapfote: Done

src/components/Modal/variant/SettingsModal.scss
alphapfote: Done

src/components/Modal/variant/SettingsModal.scss
alphapfote: Okay

PR #158, opened 2020-09-02
Custom radio buttons - public & private

alphapfote:


src/assets/icon-private.svg
```diff
@@ -0,0 +1,9 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="40" height="40" viewBox="0 0 40 40">
```
bitionaire, line 1: This file can be minified by almost 50%. Can you try running it through any cloud minifier like www.svgminify.com ?

src/assets/icon-public.svg
```diff
@@ -0,0 +1,8 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="40" height="40" viewBox="0 0 40 40">
```
bitionaire, line 1: This file should also be minified like the `icon-private.svg`

bitionaire (changes requested):
Made a few minor comments and I think the selection should look different on small screens. Either the buttons should be stretched or centered horizontally, so that they align with the buttons below in my opinion. Depending on the actual screen size it might look broken and misplaced in the current version 

Screen Shot 2020-09-08 at 13 46 16

src/assets/icon-private.svg
alphapfote: Done

src/assets/icon-public.svg
alphapfote: Done

src/components/AccessModeSelection/AccessModeSelection.scss
```diff
@@ -1,33 +1,69 @@
...
-  padding-left: 24px;
-  padding-right: 24px;
 }
 
 .access-mode-selection__radio + .access-mode-selection__radio {
-  border-left: 1px solid #eaeaea;
+  margin-left: $default-margin;
 }
 
+#access-mode-selection__private, #access-mode-selection__public {
```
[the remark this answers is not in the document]
alphapfote: This is the AccessModeSelectRadio (the id was already defined). Line 20 is there so that the circle of the radio button is not displayed. In line 38 the design of the selected button is defined. So both are needed.

alphapfote:
Requested changes done

PR #182, opened 2020-09-10
Add firestore security rules

brandstetterm:


firestore.rules
```diff
@@ -0,0 +1,90 @@
...
+                allow delete: if isAdmin() || isAuthor();
+
+                function isAuthor() {
+                    return request.auth.uid == resource.data.author;
+                }
+            }
+        }
+    }
+
+    match /templates/{template} {
```
brandstetterm, line 83: @bitionaire what do you think about the following rules for the templates? 

My intention was that anyone can read them, authenticated users can create them and only the author of a corresponding template can change & delete it.

bitionaire: Hey @brandstetterm, looks fine for me! 👍 
There might be a change necessary once we start implementing the template feature, as we intend to review new templates and toggle their public visibility (depending on the quality of the content and whether we think it uses explicit language or something else that would prohibit the release). 
But we can think about that later on. Atm everythings seems fine.

bitionaire (approved):

PR #186, opened 2020-09-24
Implement firebase function for deleting and obfuscating user data

orangehelicopter:
This is just the firebase function, but you can easily call it from the code - see https://firebase.google.com/docs/functions/callable

Still needs to be tested though

### DIFF
--- a/functions/index.js
+++ b/functions/index.js
@@ -1,3 +1,5 @@
+import * as Raven from "raven-js";
+
 const functions = require('firebase-functions');
 const admin = require("firebase-admin");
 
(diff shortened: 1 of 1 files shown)


functions/index.js
```diff
@@ -44,6 +46,42 @@ exports.scheduledDatabaseCleanup = functions.pubsub.schedule('30 3 * * *').onRun
   });
 });
 
+exports.deleteUserData = functions.https.onCall((data, context) => {
+  const userId = data.userId;
```
bitionaire, line 50: `data.userId` should be replaced with `context.auth.uid`, so that users can only delete their own accounts and are not able to delete the accounts of other users

bitionaire (changes requested):

PR #199, opened 2020-10-12
WIP: Add Store, Routes and NewBoard functionality 

brandstetterm:
Changes:
- Add .env template for private keys (.env.development.local with keys can be found in Slack)
- Add routing with predefined templates and protected routes
- Add AuthenticationManager for Firebase Auth functions
- Add NewBoard functionality (Login/SignIn & Create new Board in Firestore)

Functionality: 
- Anonymous Sign In with random animal name using Firebase Auth & Store current user in the redux store
- Create new Board in Firestore (with user as owner & participant)
- Redirect to the newly created Board

TODO:
- [ ] Implement the design as soon as it's available


src/utils/authentication/AuthenticationManager.ts
```diff
@@ -0,0 +1,18 @@
...
+function updateProfile(displayName?: string, photoURL?: string) {
+    firebase.auth().currentUser?.updateProfile({
+        displayName,
+        photoURL
+    });
+}
+
+function signInAnonymously(displayName?: string, photoURL?: string) {
+    return firebase.auth().signInAnonymously().then(_ => {
+        updateProfile(displayName, photoURL);
```
bitionaire, line 16: The `updateProfile` call might also throw an error which is uncaught / handled here. Do you intend to handle it somewhere else?

src/routes/PrivateRoute.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { useSelector } from 'react-redux';
+import { isLoaded, isEmpty } from 'react-redux-firebase';
+import {ApplicationState} from '../store/ApplicationState';
+import LoadingScreen from "../components/LoadingScreen/LoadingScreen";
+
+function PrivateRoute(props: any) {
+  const auth = useSelector((state: ApplicationState) => state.firebase.auth);
+  if (isLoaded(auth)) {
+     if (isEmpty(auth)) {
+       return <Redirect to='/new'/>;
```
bitionaire, line 12: Maybe it would be nice to redirect the user to a dedicated `/login` site and redirect him back to his initial request on successful login, instead of redirecting each user to `/new`. What do you think?

The `/join/:id` route could be ommitted with this change.

src/routes/NewBoard/NewBoard.tsx
```diff
@@ -0,0 +1,74 @@
...
+    return (
+        <div className='new-board'>
+            <Input
+                defaultValue={name}
+                type='text'
+                className='new-board__input'
+                onChange={handleChangeName}
+                onKeyDown={(e: React.KeyboardEvent<HTMLInputElement>) => {
+                    if (e.key === 'Enter') {
+                      handleLogin();
```
bitionaire, line 65: We did not specify any validation rules yet, but it would nice if we could set a char limit on the display name

src/routes/NewBoard/NewBoard.tsx
```diff
@@ -0,0 +1,74 @@
...
+                hideVotes: false
+            }
+        };
+
+        firestore.collection('boards').add(newBoard).then(success => {
+            firestore.collection('boards').doc(success.id).collection('participants').doc(auth!.uid).set({
+                name: auth!.displayName,
+                admin: true
+            });
+            props.history.push('/board/' + success.id);
```
bitionaire, line 52: I would prefer template literals and write `/board/${success.id}` here. In my opinion it is much easier to read (... with syntax highlighting in any IDE)

src/routes/NewBoard/NewBoard.tsx
```diff
@@ -0,0 +1,74 @@
...
+    let auth = firebase.auth().currentUser;
+
+    const [name, setName] = React.useState(getRandomName());
+    function handleChangeName(e: any) {
+        const name = (e.target as HTMLInputElement).value;
+        setName(name);
+    }
+
+    function handleLogin() {
+        AuthenticationManager.signInAnonymously(name).then(_ => {
```
bitionaire, line 27: `signInAnonymously` should return a `Task` of the `AuthResult`, and therefore you should be able to write ((task) => { task.getResult().getUser() }) without a second call on `firebase`. Dunno how this is handled internally, but the way with the param seems "safer" to me

src/routes/NewBoard/NewBoard.tsx
```diff
@@ -0,0 +1,74 @@
...
+            auth = firebase.auth().currentUser;
+            createBoard();
+        });
+    }
+
+    function createBoard() {
```
bitionaire, line 33: If you add the `User` as a parameter to this function (e.g. `boardCreator: User`) you can ensure that the `User` is not `null`  and no `auth!.uid` statements are necessary

.gitignore
```diff
@@ -21,3 +21,6 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+# ide
+/.idea
```
bitionaire, line 26: IntelliJ did produce `*.iml` files also in the past, are they gone in the newer versions?

package.json
```diff
@@ -13,16 +13,23 @@
...
     "react-router": "^5.2.0",
+    "react-router-dom": "^5.2.0",
     "react-scripts": "3.4.3",
     "redux-firestore": "^0.14.0",
-    "typescript": "~4.0.3"
+    "typescript": "~4.0.3",
+    "redux-thunk": "^2.3.0"
+  },
+  "devDependencies": {
+    "redux-devtools-extension": "^2.13.8"
```
bitionaire, line 32: Why is this the first `dev` depedency? As you might have noticed in my other comments I very much appreciate uniform solutions :D
If you introduce the `dev` part here I would also suggest to move all `@types` and other unneccessary packages for the production build here

bitionaire (changes requested):

src/components/LoadingScreen/LoadingScreen.css
```diff
@@ -0,0 +1,11 @@
+.loading-screen {
```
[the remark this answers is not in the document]
brandstetterm: Yes 👍 Thank you for noticing it

src/utils/authentication/AuthenticationManager.ts
brandstetterm: Yes it's very important to handle errors so the user knows if something went wrong. I just wasn't sure what a good solution would be. What do you think about Toasts? @bitionaire We could display a toast with an error message for example if the profile couldn't be updated.

src/routes/PrivateRoute.tsx
brandstetterm: I totally agree with you. Until now the PrivateRoute is only being used for accessing a board, so it would make perfectly sense to redirect to the 'join' site instead. I will change it as soon as the join site is implemented 👍

src/routes/NewBoard/NewBoard.tsx
brandstetterm: Good idea 👍

brandstetterm: What do you think about a maximum length of 15 characters? @bitionaire

.gitignore
brandstetterm: I'm using WebStorm right now and it only created the .idea directory

PR #206, opened 2020-10-12
Update public files

brandstetterm:
Changes:
- Change favicon.ico to use the scrumlr logo
- Change webpage title
- Remove unnecessary files etc.


public/index.html
```diff
@@ -7,37 +7,12 @@
...
-      Notice the use of %PUBLIC_URL% in the tags above.
-      It will be replaced with the URL of the `public` folder during the build.
-      Only files inside the `public` folder can be referenced from the HTML.
-
-      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
-      work correctly both with client-side routing and a non-root public URL.
-      Learn how to configure a non-root public URL by running `npm run build`.
-    -->
-    <title>React App</title>
+    <title>scrumlr</title>
```
bitionaire, line 12: We should use "Scrumlr", "scrumlr" or "scrumlr.io" consistently. 
I'm do not insist on any notation as long as it is uniform, but I would personally prefer "Scrumlr" as it is the notation within our logo.

public/index.html
```diff
@@ -7,37 +7,12 @@
     <meta name="theme-color" content="#000000" />
     <meta
       name="description"
-      content="Web site created using create-react-app"
+      content="Collaborative online tool"
     />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
```
bitionaire, line 12: Maybe @miiho can design a simple logo & favicon for the upcoming months?

bitionaire (changes requested):

public/index.html
bitionaire: +1 for the old logo

bitionaire (approved):

PR #269, opened 2021-01-25
Closes #218, Closes #220: Implement Admin- and UserMenu with feature stubs

### LINKED ISSUE #218: Add User Menü with Dark mode Feature stub
as a user ...

AC:

Tasks:

- [ ]  User Menü erstellen
- [ ] Icon für Dark mode im User menü anzeigen
- [ ] Mouse over "Dark mode coming soon.. "

### LINKED ISSUE #220: Admin menu is shown with time feature stub
as a user

AC:
- Timer Button is shown on the moderator menu (right hand side)

Task:

- [ ] Provide admin menu in frontend
- [ ] Provide timer button
- [ ] Show Mouse over text "Timer function coming soon"

brandstetterm:
🎉  **What has been changed?**
- Added Settings-Menu with feature stubs (buttons w/o function) for moderators
- Added Add-Menu with feature stubs (buttons w/o function) for participants
- Applied design to both menus
- Added multiple icons (SVG) for both menus

💅  **Visual changes**

🕶️ Dark theme:

https://user-images.githubusercontent.com/36969812/111623748-12f45980-87eb-11eb-9386-12726730f640.mov

💡 Light theme:

https://user-images.githubusercontent.com/36969812/111623755-14be1d00-87eb-11eb-93bc-76c4fd5b76f7.mov


src/index.scss
```diff
@@ -12,6 +12,10 @@ body {
   font-family: 'Raleway', sans-serif;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+
+  user-select: none;
```
bitionaire, line 16: Hhhm... you really want to set this globally? On which element do you want to apply this rule? I suggest to further specify the use

src/components/Menu/MenuItem/MenuToggle.tsx
```diff
@@ -0,0 +1,32 @@
...
+    toggleStartLabel: string;
+    toggleStopLabel: string;
+}
+
+function MenuToggle(props: MenuToggleProps) {
+
+    const [status, setStatus] = useState(false);
+
+    return (<button 
+                className={`menu-item menu-item${props.index} menu-item${status ? '--active' : '--disabled'} menu-item--${props.direction}`}
```
bitionaire, line 19: This looks complicated. Try to use https://www.npmjs.com/package/classnames instead and implement a rule such as `classNames('menu-item', 'menu-item${props.index}', { 'menu-item--active': status, 'menu-item--disabled': !status }, ...)`

src/components/Menu/MenuItem/MenuToggle.tsx
```diff
@@ -0,0 +1,32 @@
...
+    direction: 'left' | 'right';
+    index: number;
+    onToggle?: (active: boolean) => void;
+    toggleStartLabel: string;
+    toggleStopLabel: string;
+}
+
+function MenuToggle(props: MenuToggleProps) {
+
+    const [status, setStatus] = useState(false);
```
bitionaire, line 16: Please replace `status` with a better name which indicates the actual state and answers the boolean quetions, e.g. `isActive`, `isOpen` or whatever :)

src/components/Menu/MenuItem/MenuToggle.tsx
```diff
@@ -0,0 +1,32 @@
...
+    return (<button 
+                className={`menu-item menu-item${props.index} menu-item${status ? '--active' : '--disabled'} menu-item--${props.direction}`}
+                onClick={_ => setStatus(prevValue => !prevValue)}
+            >
+        <div className='menu-item__tooltip'>
+            <span className='tooltip__text'>
+                {status ? props.toggleStopLabel : props.toggleStartLabel}
+            </span>
+        </div>
+        <CloseIcon className='menu-item__icon menu-item__icon--end'/>
```
bitionaire, line 27: You can combine both css classes. I don't see where `<CloseIcon />` is used anywhere else without the modifier `menu-item__icon--end`

src/components/Menu/MenuItem/MenuToggle.tsx
```diff
@@ -0,0 +1,32 @@
...
+        {props.children}
```
bitionaire, line 28: I would suggest to pass the icon as a prop and not within the `children`. It makes it more explicit and maybe you can even start to restrict the usage of the prop with TypeScript

src/components/Menu/MenuItem/MenuButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return (<button 
+                className={`menu-item menu-item${props.index} menu-item--${props.direction}`}
+                onClick={props.onClick}
+            >
+        <div className='menu-item__tooltip'>
+            <span className='tooltip__text'>
+                {props.label}
+            </span>
+        </div>
+        {props.children}
```
bitionaire, line 23: Also here I would suggest to pass the icon as a prop instead of the children

src/components/Menu/MenuItem/MenuButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+type MenuButtonProps = {
+    children: React.ReactNode;
+    direction: 'left' | 'right';
+    index: number;
+    onClick?: () => void;
+    label: string;
+}
+
+function MenuButton(props: MenuButtonProps) {
```
bitionaire, line 12: Have you thought about combining the MenuButton and the ToggleButton into one component and just to extend the ToggleButton by the MenuButton? The structure is basically the same and I think it would be rather easy to reuse the code

src/components/Menu/MenuItem/MenuButton.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import './MenuItem.scss';
+
+type MenuButtonProps = {
+    children: React.ReactNode;
+    direction: 'left' | 'right';
+    index: number;
```
bitionaire, line 7: You can drop the index prop if you simply use another css selector in the generator: `menu-item:nth-child(...)`

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,70 @@
...
+import './Menu.scss';
+
+function Menu() {
+
+    const [showAdminMenu, toggleMenus] = useState(false); 
+
+    const currentUser = Parse.User.current();
+    const admins = useSelector((state: ApplicationState) => state.users.admins);
+
+    const isAdmin = admins.map(admin => admin.id).indexOf(currentUser!.id) !== -1;
```
bitionaire, line 26: I'm not sure if the result is always `-1` in each browser or if it's just a negative number by the specification. Can you look it up or are you sure that it's `-1` specified?

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    return (
+        <div className={`menu-bars menu-bars--${showAdminMenu ? 'admin' : 'user'} ${isAdmin && 'menu-bars--isAdmin'}`}>
+            <div className='menu user-menu'>
+                <div className='menu__items'>
+                    <MenuToggle index={1} direction='right' toggleStartLabel='MARK ME AS DONE' toggleStopLabel='UNMARK ME AS DONE'>
```
bitionaire, line 32: You can just enforce an uppercase label text by CSS rules and it should not be set in the label text here :)

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { ReactComponent as SettingsIcon } from 'assets/icon-settings.svg';
+import { ReactComponent as ColumnIcon } from 'assets/icon-column.svg';
+import { ReactComponent as FocusIcon } from 'assets/icon-focus.svg';
+import { ReactComponent as TimerIcon } from 'assets/icon-timer.svg';
+import { ReactComponent as VoteIcon } from 'assets/icon-vote.svg';
+import { ReactComponent as ToggleSettingsMenuIcon } from 'assets/icon-toggle-settings-menu.svg';
+import { ReactComponent as ToggleAddMenuIcon } from 'assets/icon-toggle-add-menu.svg';
+import './Menu.scss';
+
+function Menu() {
```
bitionaire, line 19: I would rather name this component `MenuBars` or something like that. Just naming it `Menu` is maybe to generic

src/components/Menu/Menu.scss
```diff
@@ -0,0 +1,202 @@
+@media screen and (max-width: 1343px) {
```
bitionaire, line 1: Please define some utility helper functions for the selection of the screen-sizes, so that can be reused in all other parts of the application

src/components/Menu/Menu.scss
```diff
@@ -0,0 +1,202 @@
...
+  .menu-bars {
+    position: fixed;
+    left: calc(50% - 134px);
```
bitionaire, line 4: This looks mysterious. Can you leave a comment here what the `50%-134px` are about?

src/components/Menu/Menu.scss
```diff
@@ -0,0 +1,202 @@
...
+    bottom: 8px;
+  }
+  .menu-bars--isAdmin {
+    left: calc(50% - 168px);
```
bitionaire, line 8: This looks mysterious. Can you leave a comment here what the `50%-168px` are about?

src/components/Menu/Menu.scss
```diff
@@ -0,0 +1,202 @@
...
+  }
+  .menu-bars--user {
+    @media (prefers-color-scheme: light) {
+      .menu-bars__switch {
+        background-color: #FFFFFF;
```
bitionaire, line 13: Please define global variables for the color codes that are used within this css file

src/components/Menu/Menu.scss
```diff
@@ -0,0 +1,202 @@
...
+  }
+  .menu-bars--user > .admin-menu {
+    visibility: hidden;
+    opacity: 0;
+    bottom: 72px;
+  }
+  .menu-bars--user > .menu-bars__switch > .switch__icon--add {
+    visibility: hidden;
+    opacity: 0;
+    transform: rotate(360deg);
```
bitionaire, line 43: Why not `transform: rotate(0)` isn't that just the same?

src/components/Menu/Menu.scss
```diff
@@ -0,0 +1,202 @@
...
+}
+
+.admin-menu {
+  --primaryColor: #ff0069;
+  --secondaryColor: #7c123e;
+}
+
+@media screen and (max-width: 1343px) {
+  .menu {
+    border-radius: 34.5px;
```
bitionaire, line 122: Is there such a thing as decimal px sizes? I think it might be relevant if you start zooming your browser view but it seems fishy somehow

src/components/Menu/MenuItem/MenuItem.scss
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  .menu-item {
+    background-color: transparent;
+    border: none;
+    border-radius: 18px;
+    padding: 0;
+    position: absolute;
+    width: 36px;
+    transition: left 0s;
```
bitionaire, line 20: Would this not be the same transition if you drop this line? `0s` indicates it for me

src/components/Menu/MenuItem/MenuItem.scss
```diff
@@ -0,0 +1,267 @@
...
+      }
+  
+    }
+    @media (prefers-color-scheme: dark) {
+      .menu-item__tooltip {
+        background-color: #FFFFFF;
+      }
+    }
+  }
+  // .menu-item:hover > .menu-item__icon {
```
bitionaire, line 47: This section should be probably deleted

src/components/Menu/MenuItem/MenuItem.scss
```diff
@@ -0,0 +1,267 @@
...
+  }
+  .menu-item--active:active > .menu-item__icon--start {
+    visibility: hidden;
+  }
+
+  .menu-item__tooltip {
+    box-sizing: border-box;
+    display: inline-flex;
+    align-items: center;
+    font-family: 'Raleway', serif;
```
bitionaire, line 73: The `font-family` should be defined globally and not for each element, except it really uses another font

src/components/Menu/MenuItem/MenuItem.scss
```diff
@@ -0,0 +1,267 @@
...
+  .menu-item--left {
+    .menu-item__tooltip {
+      direction: rtl;
+      right: 0;
+      top: 0;
+    }
+  }
+  .menu-item--right {
+    .menu-item__tooltip {
+      direction: ltr;
```
bitionaire, line 262: I guess that you did not intend to set the read direction `left-to-right` here and just wanted to position the element accordingly on the left or right side of the screen

bitionaire (changes requested):

src/index.scss
brandstetterm: The specific reason why I've set this property was to prevent random selections of areas/text on mobile phones while long-pressing/holding a button to see it's tooltip.

src/components/Menu/Menu.scss
brandstetterm: It's about positioning the menu at the center of the screen. You position the top left corner of the html-block at the center and go back 134px (half the width of the menu bar).

src/components/Menu/Menu.scss
brandstetterm: Basically you're moving the top left corner of the block to the middle of the screen and afterwards move it back half the width of the block. The end result is a centered block on the horizontal axis / width.

src/components/Menu/Menu.scss
brandstetterm: Because the default / initial situation is rotate(0). To perform a transformation you need to have a different radiant.

src/components/Menu/MenuItem/MenuButton.tsx
brandstetterm: Done ✅

src/components/Menu/MenuItem/MenuToggle.tsx
brandstetterm: Done ✅ Renamed it to ```isActive```

src/components/Menu/Menu.tsx
brandstetterm: Done ✅

src/components/Menu/MenuItem/MenuItem.scss
brandstetterm: ✅

src/components/Menu/MenuItem/MenuButton.tsx
brandstetterm: Done ✅

src/components/Menu/MenuItem/MenuToggle.tsx
brandstetterm: `menu-item__icon` is used for styling all icons and `menu-item__icon` `--start` `--end` are used for switching the visibility

src/components/Menu/MenuItem/MenuToggle.tsx
brandstetterm: Done ✅

src/components/Menu/Menu.tsx
brandstetterm: According to https://developer.mozilla.org it's always -1 in case the elements isn't present

src/components/Menu/MenuItem/MenuItem.scss
brandstetterm: Deleted it because made indeed no difference ✅

src/components/Menu/Menu.scss
brandstetterm: Done ✅

src/components/Menu/Menu.scss
brandstetterm: I changed it to 35px because I hasn't made a difference fully zoomed in

src/components/Menu/MenuItem/MenuItem.scss
brandstetterm: Done ✅

src/components/Menu/MenuItem/MenuToggle.tsx
brandstetterm: Done ✅

src/index.scss
bitionaire: Ok, got it. Then you should set this on the `.menu-item` class I guess so it won't affect any other area

src/components/Menu/MenuItem/MenuItem.scss
brandstetterm: Changed/Removed it ✅

src/components/Menu/Menu.scss
brandstetterm: Done ✅

src/components/Menu/MenuItem/MenuButton.tsx
brandstetterm: Yes, I thought about that and had a similar approach at the beginning of my implementation journey. The reason why I split them into two separate components were that both have different properties, methods and rendered components.

src/index.scss
brandstetterm: I removed it because the random selection hasn't happened anymore on my phone

bitionaire (approved):

PR #272, opened 2021-02-14
Add sticky note

alphapfote:


src/constants/style.scss
```diff
@@ -24,12 +23,34 @@ $color-pink: #F20574;
...
 $column__border-width: 12px;
 
 
+// margin and padding
+$small-padding: 16px;
+$medium-padding: 20px; 
+
+
+// font
+$small-letter-spacing: 0.25px;
```
bitionaire, line 37: I think inverting the names of the variables will be handy here. So writing `letter-spacing-...` instead of `...-letter-spacing` will make it easier to use the auto complete features of the IDE. Also I would suggest to use `--` to separate the modifier and write `letter-spacing--small` etc.

The same naming scheme should apply on the other variables.

src/routes/Board/Board.tsx
```diff
@@ -22,7 +22,7 @@ function Board(props: BoardProps) {
         }
     }, [ props.match.params.id ]);
 
-    const state = useSelector((state: ApplicationState) => ({
+    const state:any = useSelector((state: ApplicationState) => ({
```
bitionaire, line 25: We should have a look into this and replace the `any` with something meaningful (...if it is even required)

src/components/Note/Note.tsx
```diff
@@ -0,0 +1,34 @@
...
+        users: state.users
+    }));
+
+    return (
+        <li className={classNames('note', { 'note--own-card': Parse.User.current()?.id === authorId })}>      
+            <p className="note__text">{text}</p>
+            <footer>
+                <figure className="note__author">
+                    <img className="note__author-image" src={avatar} alt="User"/>
+                    <figcaption className="note__author-name">{state.users.all.filter((user) => user.id === authorId)[0]?.displayName}</figcaption>
```
bitionaire, line 28: We can add the attribute `aria-roledescription="author"` here, as it will provide more context for users with screenreaders

src/routes/Board/Board.tsx
```diff
@@ -32,15 +32,16 @@ function Board(props: BoardProps) {
...
-            <BoardComponent>
-                {
-                    state.board.data!.columns.map((column) => (<Column color="pink">{column.name}</Column>))
-                }
-            </BoardComponent>);
-        
+            <>        
+                <BoardComponent>
+                        {state.board.data!.columns.map((column:any) => (
+                            <Column color="pink" key={column.id} columnId={column.id}/>
```
bitionaire, line 38: We should move the `color` prop inside of the `<Column />` component. It can access the set color value by the provided id prop.

src/constants/style.scss
alphapfote: Done

src/components/Note/Note.tsx
alphapfote: Done

src/components/NoteInput/NoteInput.tsx
```diff
@@ -1,17 +1,96 @@
 import React from 'react';
 import './NoteInput.scss';
 import {ReactComponent as PlusIcon} from "assets/icon-add.svg";
+import store from "../../store";
```
brandstetterm, line 4:
```suggestion
import store from "store";
```

src/components/NoteInput/NoteInput.tsx
```diff
@@ -1,17 +1,96 @@
...
+import {ActionFactory} from "../../store/action";
```
brandstetterm, line 5:
```suggestion
import {ActionFactory} from "store/action";
```

src/components/NoteInput/NoteInput.tsx
```diff
@@ -1,17 +1,96 @@
...
+            },
+            input: {
+                color: 'white',
+            },
+        },
+    }),
+)(FilledInput);
+
+const NoteInput = ({columnId} : NoteInputProps) => {
+    const [value, setValue] = React.useState<string>();
```
brandstetterm, line 64: Wieso setzt du hier als default nicht den leeren String `""`? Dann brauchst du glaube ich bei value das Ausrufezeichen dahinter nicht

src/components/Column/Column.tsx
```diff
@@ -1,22 +1,39 @@
-import React from 'react';
 import './Column.scss';
 import {Color, getColorClassName} from "constants/colors";
 import NoteInput from "components/NoteInput/NoteInput";
+import Note from "components/Note/Note";
+import {ApplicationState} from "../../types/store";
```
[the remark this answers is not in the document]
alphapfote: Done

src/components/Note/Note.tsx
```diff
@@ -0,0 +1,33 @@
+import './Note.scss';
+import avatar from 'assets/avatar.png';
+import { useSelector } from 'react-redux';
+import {ApplicationState} from "../../types/store";
```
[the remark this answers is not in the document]
alphapfote: Done

src/components/NoteInput/NoteInput.tsx
alphapfote: Done

src/components/NoteInput/NoteInput.tsx
alphapfote: Done

src/components/NoteInput/NoteInput.tsx
alphapfote: Danke für den Tipp, wusste ich nicht.

PR #284, opened 2021-02-28
Closes #192: show active users

### LINKED ISSUE #192: Show active users
As a user, i wanr to see the name of the session and all active users, in order to see, in which meeting i am.

AC:

- The first 4 or 5 active users are shown with their avatar. The user own is the first one. On hover shows Name of the user. (<=5 show all avatars, >5 4 avatar + number)
 - The amount of the other active users is shown. On click shows list of names of the other user with avatar)

Tasks:
- [x] Check in Parse how to expose active user using websockets
- [x] Implement solution in BE
- [x] Show active users taking the number of users into account

busilina:
🎉 **What has been changed?** 
- Added BoardUsers, BoardUser and RestUserCount components;
- Added online/offline functionality;
- Added styles (see open issues below).

⚠️ **Open or to-be-discussed issues**
- Theming & positioning
- Since we don't have support for avatars yet I show user initials instead (i.e. AM for the user 'Admin Mustermann')

 💅 **Visual changes**

💡 _Light Theme_ 

image

🕶️ _Dark Theme_

image


src/components/Board/Board.tsx
```diff
@@ -1,125 +1,129 @@
...
+        <button
+          className={`board__navigation board__navigation-prev ${getColorClassName(columnColors[previousColumnIndex])}`}
+          onClick={handlePreviousClick}
+          aria-hidden={true}
+        >
+          <LeftArrowIcon className="board__navigation-arrow board__navigation-arrow-prev" />
+        </button>
+      )}
+
+      <BoardUsers numOfUsersToShow={NUM_OF_DISPLAYED_USERS} />
```
bitionaire, line 109: I think I would prefer to let the `BoardUsers` component deal with the question how many users should be displayed, as it is not related to the board.

Also I think the magical number `4` is only suitable for larger displays and should be set to a smaller number on mobile screens.

src/components/BoardUsers/BoardUser.tsx
```diff
@@ -0,0 +1,23 @@
+import './BoardUser.scss';
+import {getInitials} from 'constants/Name'
+export interface BoardUserProps {
+  id: string;
+  name: string;
+  avatar?: string;
+}
+
+const BoardUser = ({ name, avatar }: BoardUserProps) => {
```
bitionaire, line 9: I would rename the component to `UserAvatar`, as we probably will have more components related to a user (e.g. `UserProfile` or something like that if we want to display more information)

src/components/BoardUsers/BoardUser.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  return (
+    <li>
+      <div className="user">
```
bitionaire, line 13: I think this `div` can be removed and the styles can be applied on the parent `li` element. Also I like to name the CSS classes just like the components name, so that is easier to find on debug of the HTML code. In this case it would be `board-user` or `user-avatar` as I suggested a rename of the component. Whatever you prefer :)

src/components/BoardUsers/BoardUsers.tsx
```diff
@@ -0,0 +1,32 @@
...
+  numOfUsersToShow: number;
+}
+
+const BoardUsers = ({ numOfUsersToShow }: BoardUsersProps) => {
+  const users = useSelector((state: ApplicationState) => state.users.all);
+  const currentUser = Parse.User.current();
+
+  const me = users.find(user => user.id === currentUser!.id);
+  const connectedUsers = users.filter(user => user.online);
+  const them = connectedUsers.filter(user => user.id !== currentUser!.id);
```
bitionaire, line 18: I would put `me` and `them` next to each other, so that is easier to comprehend

src/components/BoardUsers/BoardUsers.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const usersToShow = them.splice(0, (them.length > numOfUsersToShow ? numOfUsersToShow - 1 : numOfUsersToShow));
+
+  return (
+    <div className="users">
+      <ul className="user-list">
```
bitionaire, line 23: I think you can also combine the parent `div` and the `ul` into one element here. Also I would prefer the name `board-users` as the class name here for easier debugging

src/components/BoardUsers/BoardUsers.tsx
```diff
@@ -0,0 +1,32 @@
...
+        {them.length > 0 && <RestUserCounter count={them.length} />}
```
bitionaire, line 24: I think you should use `connectedUsers` here instead. We can show a complete list of users (online & offline) in some other detail view

src/components/BoardUsers/BoardUsers.tsx
```diff
@@ -0,0 +1,32 @@
...
+        {usersToShow.map((user, i) => <BoardUser key={i} id={user.id} name={user.displayName} />)}
```
bitionaire, line 25: I think you should use the `user.id` as the key here, as `i` might change on every call or any order of the `usersToShow`

src/components/BoardUsers/BoardUsers.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+const BoardUsers = ({ numOfUsersToShow }: BoardUsersProps) => {
+  const users = useSelector((state: ApplicationState) => state.users.all);
+  const currentUser = Parse.User.current();
+
+  const me = users.find(user => user.id === currentUser!.id);
+  const connectedUsers = users.filter(user => user.online);
+  const them = connectedUsers.filter(user => user.id !== currentUser!.id);
+  const usersToShow = them.splice(0, (them.length > numOfUsersToShow ? numOfUsersToShow - 1 : numOfUsersToShow));
```
bitionaire, line 19: Also here you should use the `connectedUsers` instead of `them` and combine the conditional of both queries `(user.id !== currentUser.id && user.online)`

src/components/BoardUsers/BoardUsers.tsx
bitionaire, line 24: I think the `RestUserCounter` component is unnecessary. The overhead of a new component seems to me to be much more that those 4 lines of code put here

src/store/middleware/board.ts
```diff
@@ -1,127 +1,220 @@
...
+            for (const result of results) {
+              updateUsers(result as Parse.Role);
+            }
+          });
+        });
+      });
+
+      const userQuery = new Parse.Query(Parse.User);
+      userQuery.equalTo("online", true);
+      userQuery.subscribe().then((subscription) => {
```
bitionaire, line 132: I don't think that you need this query since the `update` of line 117 should also update by the user attributes? Not sure, but if not you should rather combine both queries.
Also if you combine the query the `setUserStatus` action becomes obsolete.

bitionaire (changes requested):
Very nice PR! I suggested a few minor tweaks on the code

src/components/BoardUsers/BoardUsers.tsx
busilina: done

src/components/BoardUsers/BoardUsers.tsx
busilina: done

src/components/BoardUsers/BoardUsers.tsx
busilina: done

src/components/BoardUsers/BoardUsers.tsx
busilina: done `RestUserCounter` removed

src/components/BoardUsers/BoardUsers.tsx
busilina: done

src/components/BoardUsers/BoardUser.tsx
busilina: done. I liked `user-avatar` more :)

src/components/BoardUsers/BoardUser.tsx
busilina: done

src/components/Board/Board.tsx
busilina: done

src/store/middleware/board.ts
busilina: I guess we do need them but maybe we can combine them. Let's do this together since I still need some help with Parse 😅

busilina: finally done. Please check it again.

bitionaire (approved):

PR #288, opened 2021-03-17
Closes #277: Column scrolling, if there are more notes then fit on the screen vertically

### LINKED ISSUE #277: If there are more notes then fit on the screen vertically, there should be a column scrolling, so you can see all notes

alphapfote:


src/components/Column/Column.scss
```diff
@@ -58,10 +63,15 @@
...
+  overflow: auto;
+}
+
 .column__note-list {
-  padding: 0;
+  margin: 0;
+  padding: 0 $padding--large;
   display: grid;
-    grid-template-columns: repeat(auto-fill,minmax(260px,1fr));
+    grid-template-columns: repeat(auto-fill);
```
brandstetterm, line 74: Regarding to my Chrome DevTools that's a `Invalid property value`

src/components/Column/Column.tsx
```diff
@@ -29,11 +29,13 @@ const Column = ({ color, columnId } : ColumnProps) => {
...
                     <NoteInput columnId={columnId}/>
                 </header>
-                <ul className="column__note-list">
-                    {notesInColumn.map((note:any, index:number) =>
-                        <Note key={note.id} text={note.text} authorId={note.author}/>
-                    )}
-                </ul>
+                <div className="column__notes-wrapper">
+                    <ul className="column__note-list">
+                        {notesInColumn.map((note:any, index:number) =>
```
brandstetterm, line 34: Personally I really dislike the usage of `any`. Maybe you could use the `NoteClientModel` type here

src/components/Column/Column.scss
alphapfote: I reversed the changes, they have no effect

bitionaire (approved):

src/components/Column/Column.scss
bitionaire: Problem seems to be back

bitionaire (changes requested):

bitionaire (approved):

PR #292, opened 2021-03-30
Closes #291: Color names

### LINKED ISSUE #291: Enable colored columns
Implement correct color variables:

Backlog Blue #0057FF
Planning Pink #FF0069
Lean Lila #C000FF
Retro Red #DF1313
Grooming Green #17C072
Poker Purple #5E00FF
Outcome Orange #FF3C19

and add colors to the board model within the backend.

bitionaire:
localhost_3000_ (1)


server/src/cloud/board.ts
```diff
@@ -67,6 +67,7 @@ const respondToJoinRequest = async (
 export interface CreateBoardRequest {
   columns: {
     name: string;
+    color: string;
```
brandstetterm, line 70: Why don't you use the Color type from `src/constants/colors.ts` instead of string?

server/src/cloud/column.ts
```diff
@@ -65,6 +68,10 @@ export const initializeColumnFunctions = () => {
                 columns[request.columnId].name = request.name;
             }
 
+            if (request.color) {
```
brandstetterm, line 71: Maybe it would be reasonable to check if the given color is an actual color of the Color type in src/constants/colors.ts

server/src/cloud/column.ts
```diff
@@ -6,6 +6,7 @@ import {serverConfig} from "../index";
 export interface AddColumnRequest {
     boardId: string;
     name: string;
+    color: string;
```
brandstetterm, line 9: Color type instead of string?

server/src/cloud/column.ts
```diff
@@ -18,6 +19,7 @@ export interface EditColumnRequest {
     boardId: string;
     columnId: string;
     name?: string;
+    color?: string;
```
brandstetterm, line 22: Color type instead of string?

src/routes/Board/Board.tsx
```diff
@@ -6,11 +6,13 @@ import {ApplicationState} from "types/store";
 import store from "store";
 import {ActionFactory} from "store/action";
 import Parse from 'parse';
+import Note from "../../components/Note/Note";
```
brandstetterm, line 9:
```suggestion
import Note from "components/Note/Note";
```

src/routes/NewBoard/NewBoard.tsx
```diff
@@ -2,10 +2,11 @@ import * as React from 'react';
 import Input from '@material-ui/core/Input';
 import Button from '@material-ui/core/Button';
 import {AuthenticationManager} from 'utils/authentication/AuthenticationManager';
-import getRandomName from '../../constants/Name';
+import getRandomName from 'constants/Name';
 import {RouteComponentProps} from "react-router";
 import Parse from "parse";
 import {API} from "../../api";
```
brandstetterm, line 8:
```suggestion
import {API} from "api";
```

src/routes/Board/Board.tsx
```diff
@@ -40,8 +42,12 @@ function Board() {
             <>    
                 {waitingUser}
                 <BoardComponent name={state.board.data!.name} boardstatus={boardstatus}>
-                        {state.board.data!.columns.map((column:any) => (
-                            <Column color="pink" key={column.id} columnId={column.id}/>
+                        {state.board.data!.columns.map((column) => (
```
brandstetterm, line 45: Maybe we should create a Column type

src/routes/Board/Board.tsx
```diff
@@ -40,8 +42,12 @@ function Board() {
...
+                            <Column key={column.id} id={column.id!} name={column.name} color={column.color}>
+                                {state.notes.filter((note) => note.columnId === column.id).map((note) =>
```
brandstetterm, line 47: You could use the NoteClientModel type

src/routes/Board/Board.tsx
bitionaire: Column type is used here since `board.data.columns` is already typed

src/routes/Board/Board.tsx
bitionaire: The `state.notes` already have the correct type, so there's no need to write it more explicit?

brandstetterm (approved):

server/src/cloud/board.ts
```diff
@@ -97,16 +86,16 @@ export interface JoinBoardRequest {
 }
 
 export const initializeBoardFunctions = () => {
-  Parse.Cloud["onLiveQueryEvent"](({event, sessionToken, ...other}) => {
+  Parse.Cloud.onLiveQueryEvent(({event, sessionToken, ...other}) => {
```
brandstetterm, line 89: Property 'onLiveQueryEvent' does not exist on type 'typeof Cloud'

PR #293, opened 2021-04-01
Closes #188: Linting

### LINKED ISSUE #188: Add Prettier & ESLint
- [ ] Configure Prettier & ESLint incl. GitHub Actions / CI

bitionaire:
Added linting, in the first step auto-formatting and linting will be applied on staged files. We should roll-out this feature to the CI once we fixed a significant amount of files in order to prevent a large number of merge conflicts.


.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
brandstetterm, line 1: Is that correct?

server/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "printWidth": 180,
+  "arrowParens": "always",
+  "semi": false,
```
brandstetterm, line 4: Ich would prefer `semi: true`

.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "printWidth": 180,
+  "arrowParens": "always",
+  "semi": false,
```
brandstetterm, line 4: Ich would prefer `semi: true`

brandstetterm (changes requested):

.husky/.gitignore
bitionaire: This is auto-generated by husky and ignores `husky.sh` within the directory `_`, so seems correct for me 🤷

brandstetterm (approved):

PR #602, opened 2021-08-12
Closes #165: Implement Cookie Consent

### LINKED ISSUE #165: As a user I want to be informed about cookies before accepting them via the cookie consent
Option a) Implement Cookie Policy yourself
- [ ] Build component
- [ ] Create frontend design
- [ ] Implement frontend

Option b) Use library
- [ ] Search for library
- [ ] Implement library
- [ ] Create frontend design
- [ ] Implement frontend

Idea for library: https://www.osano.com/features/consent-management

andiKandi:
<h2> Description </h2>
Implement basic cookie notice. Content was taken from Scrumlr V1.

# TODO:
- [ ] Get approval of content
- [ ] Get design
- [ ] Do Not Track - Header Funktion hinzufügen


brandstetterm:
Blocked -> Waiting for design

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,132 @@
...
+import {useSelector} from "react-redux";
+import {ApplicationState} from "types/store";
+import {ActionFactory} from "store/action";
+import CookiePolicy from "./CookiePolicy";
+
+/**
+ * Converts string value to boolean value.
+ * @param input string that needs conversion to boolean value.
+ */
+const stringToBoolean = (input: string): boolean => {
```
bitionaire, line 14: This function seems overly complicated to me, since you're storing the value `true` within your code and my guess is that noone would manually edit this value. I would rather suggest to replace that function with `cookieConsent.value === 'true'`

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const [showCookieNotice, setShowCookieNotice] = React.useState(true); // state: states whether cookie notice is shown.
+  const [open, setOpen] = React.useState(true); // state: states whether backdrop is open.
+
+  const [showDialog, setShowDialog] = React.useState(false);
+  const handleShowDialog = () => {
+    setShowDialog(!showDialog);
+  };
+
+  const shouldShowCookieNotice = () => !localStorage.getItem(scrumlrCookieName); // function: checks if scrumlrCookieName is present in localStorage.
```
bitionaire, line 51: I don't get why you implemented this as a function? `const shouldShowCookieNotice = !localStorage.getItem(scrumlrCookieName)` should do the trick

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,132 @@
...
+    document.cookie = `${key}=${value};`;
+  };
+
+  /**
+   * Sets cookie consent value i.e. scrumlrCookieName true.
+   */
+  const acceptFunction = () => {
+    if (shouldShowCookieNotice()) {
+      saveToStorage(scrumlrCookieName, "true");
+      setShowCookieNotice(!showCookieNotice);
```
bitionaire, line 69: Since the accept should close the cookie notice you should write `setShowCookieNotice(false);` here and make it more explicit

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const openPolicy = () => {
+    setShowDialog(true);
+  };
+
+  if (!shouldShowCookieNotice()) return null; // prevents cookie notice being shown if scrumlrCookieName is set.
+
+  // cookie notice rendering
+  return (
+    <div className={showCookieNotice ? "cookie-notice--show" : "cookie-notice--hidden"}>
```
bitionaire, line 93: You should not toggle the visibilty via CSS an rather `return null;` if `showCookieNotice` is `false` or the component otherwise

src/components/CookieNotice/CookieNotice.scss
```diff
@@ -0,0 +1,68 @@
+@import "constants/style.scss";
+
+$cookie-consent-width: 100%;
+$cookie-consent-top: 100%;
+$cookie-consent-position: sticky;
```
bitionaire, line 5: - I think the position should be `fixed` here, since scrolling on the page wont change the position. Or I just don't get it
- Since you're only using this variable once in `.cookie-notice--show` you don't need the explicit variable here or won't change it

src/components/CookieNotice/CookieNotice.scss
```diff
@@ -0,0 +1,68 @@
+@import "constants/style.scss";
+
+$cookie-consent-width: 100%;
+$cookie-consent-top: 100%;
```
bitionaire, line 4: I dont get the `100%` and it should be `0`, right?

src/components/CookieNotice/CookieNotice.scss
```diff
@@ -0,0 +1,68 @@
+@import "constants/style.scss";
+
+$cookie-consent-width: 100%;
```
bitionaire, line 3: That is the default for a `<div />` so you don't need it

src/components/CookieNotice/CookieNotice.scss
```diff
@@ -0,0 +1,68 @@
...
+$cookie-consent-button-text-size: 18px;
+
+.cookie-notice--show {
+  background: $color-light-gray;
+  color: $color-black;
+  position: $cookie-consent-position;
+  top: $cookie-consent-top;
+  width: $cookie-consent-width;
+  padding: $padding--small;
+  display: block;
```
bitionaire, line 17: `display: block` ist the default for `<div />` -> delete :)

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,132 @@
...
+    case null:
+      return false;
+    default:
+      return false;
+  }
+};
+
+const CookieNotice = () => {
+  /** cookie name */
+  const scrumlrCookieName = "scrumlr_cookieConsent";
```
bitionaire, line 32: This should be a constant outside of the component, e.g. `const COOKIE_CONSENT_NAME='...'` on top of the file

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,132 @@
...
+    }
+  };
+
+  /**
+   * Sets cookie consent value i.e. scrumlrCookieName false.
+   */
+  const declineFunction = () => {
+    if (shouldShowCookieNotice()) {
+      saveToStorage(scrumlrCookieName, "false");
+      setShowCookieNotice(!showCookieNotice);
```
bitionaire, line 80: see `false` @ `acceptFunction`

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+        <button className="cookie-notice__button-decline" type="button" onClick={declineFunction}>
+          Decline
+        </button>
+
+        <button className="cookie-notice__button-accept" type="button" color="primary" onClick={acceptFunction}>
+          Accept
+        </button>
+      </div>
+      <CookiePolicy scrumlrCookieName={scrumlrCookieName} acceptFunction={acceptFunction} onClose={handleShowDialog} show={showDialog} />
```
bitionaire, line 127: If you export the `COOKIE_CONSTENT_NAME` you can just access it as a variable instead of passing it as a prop for the CookiePolicy component here. _qed._ :D

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,132 @@
...
+  if (localStorage.getItem(scrumlrCookieName)) {
+    cookieState.cookieConsent.name = scrumlrCookieName;
+    cookieState.cookieConsent.value = stringToBoolean(localStorage[scrumlrCookieName]);
+  }
+
+  const [showCookieNotice, setShowCookieNotice] = React.useState(true); // state: states whether cookie notice is shown.
+  const [open, setOpen] = React.useState(true); // state: states whether backdrop is open.
+
+  const [showDialog, setShowDialog] = React.useState(false);
+  const handleShowDialog = () => {
```
bitionaire, line 47: I would name this function `toggleShowDialog` to make it more explicit. Might be just me, but `handle` implies that i probably can pass an argument

src/components/CookieNotice/CookiePolicy.scss
```diff
@@ -0,0 +1,50 @@
...
+.portal {
+  border-radius: 5px;
+}
+
+.cookie-policy {
+  margin: $margin--small;
+}
+
+.cookie-policy__title {
+  display: block;
```
bitionaire, line 14: ... you should not set `display: block` explicitly if it is the default of the HTML elements. It just bloats the file size of the CSS and will not change the look & feel

src/components/CookieNotice/CookiePolicy.tsx
```diff
@@ -0,0 +1,101 @@
...
+  };
+
+  return (
+    <Portal onClose={onClose}>
+      <div className="cookie-policy">
+        <div className="cookie-policy__title">
+          <h3>Cookie Policy of scrumlr.io</h3>
+        </div>
+        <div className="cookie-policy__body">
+          <p>
```
bitionaire, line 30: This text is rather lengthy and it might be exhausting to change something here. I would suggest to import the text as markdown and maybe you can research some different ways to do it? https://stackoverflow.com/questions/42928530/how-do-i-load-a-markdown-file-into-a-react-component suggests `marked` but maybe there are even simpler ways by now to do it in `create-react-app`. Dunno

src/store/action/cookie.ts
```diff
@@ -0,0 +1,43 @@
+import {CookieClientModel} from "types/cookie";
+
+/** This object lists cookie object specific internal Redux Action types. */
+export const CookieActionType = {
```
bitionaire, line 4: You can just delete everything redux related as we already discussed

src/types/cookie.ts
```diff
@@ -0,0 +1,7 @@
+/** The representation of a note on the client. */
+export interface CookieClientModel {
```
bitionaire, line 2: You can just name it `CookieConsent` or something like that. `Client` and `Server` in the name were just used to easier differentiate the models of the server and the client. The cookie consent is something that will only be stored on the client so the name is not necessary. Also the term `consent`, which is defined by the `value` is missing in the name.

bitionaire: ... i guess this interface will be deleted anyway if you delete the redux stuff. So the comment above is just jFYI :)

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  if (localStorage.getItem(scrumlrCookieName)) {
+    cookieState.cookieConsent.name = scrumlrCookieName;
+    cookieState.cookieConsent.value = stringToBoolean(localStorage[scrumlrCookieName]);
+  }
+
+  const [showCookieNotice, setShowCookieNotice] = React.useState(true); // state: states whether cookie notice is shown.
+  const [open, setOpen] = React.useState(true); // state: states whether backdrop is open.
+
+  const [showDialog, setShowDialog] = React.useState(false);
```
bitionaire, line 46: Maybe you can rename this to `showCookiePolicy` here instead, which would make it more explicit

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,132 @@
...
+    cookieConsent: applicationState.cookieConsent,
+  }));
+
+  if (localStorage.getItem(scrumlrCookieName)) {
+    cookieState.cookieConsent.name = scrumlrCookieName;
+    cookieState.cookieConsent.value = stringToBoolean(localStorage[scrumlrCookieName]);
+  }
+
+  const [showCookieNotice, setShowCookieNotice] = React.useState(true); // state: states whether cookie notice is shown.
+  const [open, setOpen] = React.useState(true); // state: states whether backdrop is open.
```
bitionaire, line 44: I guess `open` and `showCookieNotice` share the same state, so only one state is necessary here.

bitionaire (changes requested):
I left some comments and change suggestions

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,97 @@
+import "./CookieNotice.scss";
+import React from "react";
+import Backdrop from "@material-ui/core/Backdrop";
```
brandstetterm, line 3: @bitionaire and me are not really keen on using Googles material-ui. Maybe you could do it yourself?

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,97 @@
...
+      setShowCookieNotice(false);
+    }
+  };
+
+  const openPolicy = () => {
+    setShowCookiePolicy(true);
+  };
+  // prevents cookie notice being shown if scrumlrCookieName is set in localStorage
+  if (!shouldShowCookieNotice) return null;
+  if (!showCookieNotice) return null;
```
brandstetterm, line 55:
```suggestion
  if (!shouldShowCookieNotice || !showCookieNotice) return null;
```

src/components/CookieNotice/CookiePolicy.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const CookiePolicy = ({acceptFunction, onClose, show, darkBackground}: CookiePolicyProps) => {
+  const [policy, setPolicy] = useState({markdown: ""});
+
+  useEffect(() => {
+    fetch(policyText)
+      .then((response) => response.text())
+      .then((text) => setPolicy({markdown: marked(text)}));
+  }, []);
```
brandstetterm, line 21:
```suggestion
  });
```

src/components/CookieNotice/CookieNotice.scss
```diff
@@ -0,0 +1,59 @@
...
+  background: $color-backlog-blue;
+  color: $color-white;
+  margin: 0 0 $margin--default $margin--default;
+  border: none;
+  padding: 5px 15px;
+  border-radius: 5px;
+  font-family: inherit;
+  font-size: $cookie-consent-button-text-size;
+}
+.cookie-notice__button-accept:hover {
```
brandstetterm, line 29: Is it really the preferred behaviour that a button is less visible on hover? 
I think I'd prefer something like `transform: scale(1.05)`

brandstetterm: Also I'd give `cookie-notice__button-accept`, `cookie-notice__button-decline` and  `cookie-notice__button-cookie-policy` an additional class `cookie-notice__button` and summarise all common css attributes there. 

For example all there buttons have the same hover effect, padding and border-radius.

src/components/CookieNotice/CookiePolicy.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <Portal onClose={onClose} darkBackground={darkBackground}>
+      <div className="cookie-policy">
+        <div className="cookie-policy__title">
+          <h1>Cookie Policy of scrumlr.io</h1>
+        </div>
+        <div className="cookie-policy__body" dangerouslySetInnerHTML={{__html: policy.markdown}} />
+        <div className="cookie-policy__footer">
+          <button className="cookie-policy__button-accept" type="button" onClick={handleAccept}>
```
brandstetterm, line 39: Why's the Decline-Button missing?

src/components/CookieNotice/CookiePolicy.tsx
```diff
@@ -0,0 +1,48 @@
...
+  if (!show) {
+    return null;
+  }
+
+  const handleAccept = () => {
+    acceptFunction();
+  };
+
+  return (
+    <Portal onClose={onClose} darkBackground={darkBackground}>
```
brandstetterm, line 32: The portal has some margin at the sides because it's designed for looking good at our board with the coloured border
Screenshot 2021-09-17 at 13 39 04
l

src/components/CookieNotice/CookiePolicy.scss
```diff
@@ -0,0 +1,48 @@
...
+  color: $color-white;
+  margin: 0 0 0 auto;
+  border: none;
+  padding: 5px 15px;
+  border-radius: 5px;
+  font-family: inherit;
+  font-size: $cookie-consent-button-text-size;
+  justify-content: flex-end;
+  display: flex;
+  position: sticky;
```
brandstetterm, line 47: I don't think this is doing anything here

src/components/CookieNotice/CookiePolicy.scss
```diff
@@ -0,0 +1,48 @@
...
+  background: $color-backlog-blue;
+  color: $color-white;
+  margin: 0 0 0 auto;
+  border: none;
+  padding: 5px 15px;
+  border-radius: 5px;
+  font-family: inherit;
+  font-size: $cookie-consent-button-text-size;
+  justify-content: flex-end;
+  display: flex;
```
brandstetterm, line 46: `justify-content: flex-end` doesn't do anything here because you do not position any element inside of the button. Also, just for placing the button at the end of the footer I'd use `display: block` or `display: inline-block`

src/components/CookieNotice/CookiePolicy.scss
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.cookie-policy__button-accept {
+  background: $color-backlog-blue;
+  color: $color-white;
+  margin: 0 0 0 auto;
+  border: none;
+  padding: 5px 15px;
+  border-radius: 5px;
+  font-family: inherit;
```
brandstetterm, line 43: Don't think this does anything because inherit only inherits from the parent, not from greater ancestors

src/components/CookieNotice/CookieNotice.scss
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.cookie-notice__button-accept {
+  background: $color-backlog-blue;
+  color: $color-white;
+  margin: 0 0 $margin--default $margin--default;
+  border: none;
+  padding: 5px 15px;
+  border-radius: 5px;
+  font-family: inherit;
```
brandstetterm, line 26: Don't think this does anything because inherit only inherits from the parent, not from greater ancestors

brandstetterm (changes requested):

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,97 @@
...
+  // states whether cookie notice is shown.
+  const [showCookieNotice, setShowCookieNotice] = React.useState<boolean>(true);
+
+  const [showCookiePolicy, setShowCookiePolicy] = React.useState(false);
+  const toggleShowCookiePolicy = () => {
+    setShowCookiePolicy(!showCookiePolicy);
+  };
+
+  // show cookie notice if there's no cookie in local storage
+  const shouldShowCookieNotice = !localStorage.getItem(COOKIE_CONSENT_NAME);
```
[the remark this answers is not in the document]
andiKandi: tried it, but if .getItem(...) is removed and you refresh the page, the cookie notice will show again although the cookie is set. -> Seems to be needed.

src/components/CookieNotice/CookieNotice.tsx
```diff
@@ -0,0 +1,97 @@
+import "./CookieNotice.scss";
+import React from "react";
+import Backdrop from "@material-ui/core/Backdrop";
+import CookiePolicy from "./CookiePolicy";
+
+const COOKIE_CONSENT_NAME = "scrumlr_cookieConsent";
+
+const CookieNotice = () => {
+  // states whether cookie notice is shown.
+  const [showCookieNotice, setShowCookieNotice] = React.useState<boolean>(true);
```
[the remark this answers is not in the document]
andiKandi: isn't <boolean>(true) cleaner, i.e. more explicit?

src/components/CookieNotice/CookiePolicy.tsx
andiKandi: because it wasn't there in scrumlr V1 and I copied the "design" logic. But you're right: it should be there.

src/components/CookieNotice/CookieNotice.tsx
andiKandi: replaced material-ui with portal

src/components/CookieNotice/CookieNotice.tsx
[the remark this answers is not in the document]
brandstetterm: 1. You could import useState so you don't have to write `React.useState(...)`, but that's up to you
2. You don't need to write `useState<boolean>(true)` because the initialValue already applies the boolean type to the function

src/components/CookieNotice/CookieNotice.tsx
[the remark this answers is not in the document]
brandstetterm: 👍🏼

brandstetterm (approved):
I've noted one thing. But it's up to you if you want to change it. It's a approve anyways. 

The Godfather sends his regards.

bitionaire (approved):

Issue #586, opened 2021-08-11
As a participant, I want to be able to unstack notes from stacks


brandstetterm:
AC:
- [ ] Participants can unstack stacked notes 
- [ ] Unstacked notes will be displayed in the board again
- [ ] Tests have been added/updated

Design: See v1

<img width="713" alt="Screenshot 2021-09-03 at 10 07 41" src="https://user-images.githubusercontent.com/36969812/131972680-8532ed52-febf-42e5-8512-56a84ad9e5cf.png">

Issue #585, opened 2021-08-11
As a participant, I want to be able to create a cover for stacks


brandstetterm:
AC: 
- [ ] Participants can create covers for stacked notes 
- [ ] Covers can be updated/deleted
- [ ] All participants see the cover instead of the actual note content of the parent 
- [ ] Tests have been added/updated

PR #619, opened 2021-08-20
Closes #584: Stacking of Notes 

### LINKED ISSUE #584: As a participant, I want to create stacks of notes if these are identical in content
Tasks:
- Implement Stacks as Schema, etc. in the parse cloud
- Implement Stacks in the front end (Drag & Drop notes over other notes to create a stack)

brandstetterm:
# Description

Participants are able to create stacks by moving notes on top of other notes. 

Issue: #584

## Changes

- Add `parent` attribute to notes 
- Use `react-dnd` for creating stacks by moving notes on top of each other
- Adjust `NoteDialog` to show stacks
- Adjust `editNote` cloud function so that every participant has the permission to move ALL notes and stacks to different columns
- Adjust `deleteNote` cloud function so that childNotes unstack if the parent of a stack has been deleted

## Follow up tasks:

- Ability to create stack covers #585
- Ability to unstack notes #586


bitionaire (approved):
LGTM!

Issue #595, opened 2021-08-11
Conclude voting phase once all users are marked as "ready"


brandstetterm:
BLOCKED by #593, #592 -> Must be implemented first

Issue #594, opened 2021-08-11
Allow moderator to conclude voting and display vote results to all participants


brandstetterm:
BLOCKED by #593, #592 -> Must be implemented first

PR #624, opened 2021-08-23
Closes #592, closes #593, closes #596: Voting

### LINKED ISSUE #592: Allow users to vote on notes
BLOCKED by #593 -> Must be implemented first

### LINKED ISSUE #593: Allow moderator to start vote phase
As a first step, phases must be defined and implemented

This could be done for example by adding a "phase" field to our board scheme which can hold our predefined phases (e.g. "voting", ...)

### LINKED ISSUE #596: Allow users to delete/revert a vote
BLOCKED by #593, #592 -> Must be implemented first

brandstetterm:
# Description

As a participant I want to vote for notes or stacks. As a moderator I want to toggle if voting is enabled or not. 

Issues: 
- #592 
- #593 
- #596 

## Changes

- Change default vote limit to 10 votes per user
- Update `addVote` and `removeVote` cloud functions for error handling
- Update `editBoard` cloud function so that `voting` is toggleable
- Update `deleteBoard` and `deleteNote` cloud functions to also delete related votes
- Add api functions for votes
- Add redux actions, middleware and reducer for votes
- Update redux actions and middleware for the board
- Create `Votes` and `DotButton` component 
- Update `Note` and `NoteDialog` component to display votes for every note
- Update `MenuBars` to make `voting` toggleable
- Add new tests and update old tests

## Follow up tasks

- At the moment all participants see every vote during voting phase. Maybe we should disable that and only show all votes after the voting has been ended #594 
- #595 

---
<details>
<summary>Visual Changes</summary>

Screenshot 2021-09-01 at 11 42 06
Screenshot 2021-09-01 at 11 18 11
Screenshot 2021-09-01 at 11 41 47
Screenshot 2021-09-01 at 11 18 39

</details>


server/src/cloud/board.ts
```diff
@@ -73,6 +73,7 @@ export type EditableBoardAttributes = {
   timerUTCEndTime?: Date;
   expirationUTCTime?: Date;
   joinConfirmationRequired?: boolean;
+  voting?: "active" | "disabled";
```
bitionaire, line 76: Aren't there 3 states: actice, disabled and e.g. concluded?

server/src/cloud/vote.ts
```diff
@@ -2,34 +2,58 @@ import {requireValidBoardMember, getMemberRoleName} from "./permission";
...
+    }
+    // Check if 'voting' is active
+    if (board.get("voting") === "disabled") {
+      return {status: "Error", description: "You cannot vote while voting is disabled"};
+    }
+    // Check if user will exceed vote limit
+    const voteQuery = new Parse.Query("Vote");
+    voteQuery.equalTo("board", board);
+    voteQuery.equalTo("user", user);
+    if ((await voteQuery.count({useMasterKey: true})) >= board.get("voteLimit")) {
```
bitionaire, line 26: I guess you have to double check this with a `beforeSave` hook or something like that, because this might lead to multiple saved votes if the cloud function gets called multiple times in a short period. Did you try to spam adding votes?

bitionaire (approved):

server/src/cloud/vote.ts
brandstetterm: Yes I've tried it and it never worked 😄

server/src/cloud/board.ts
brandstetterm: Good point! But I guess the concluded state will be added after finishing #594 or #595.

PR #692, opened 2021-09-09
Closes #661: Voting phase iterations

### LINKED ISSUE #661: As a moderator I want to reset/clear the result of the previous voting phase

Dominik-Weinzierl:
<h2> Description </h2>

Store and show votes based on voting iteration.

Issue: #661 

## Changelog 
- Add `votingIteration` as `Board` and `Vote` attribute
- Change voting limit to voting limit per iteration
- Filter displayed votes based on the `votingIteration`


server/src/model/init.ts
```diff
@@ -35,6 +36,7 @@ const addInitialVoteSchema = async () => {
   schema.addPointer("board", "Board", {required: true});
   schema.addPointer("note", "Note", {required: true});
   schema.addPointer("user", "_User", {required: true});
+  schema.addNumber("votingIteration", {required: true, defaultValue: 0});
```
brandstetterm, line 39: I think I wouldn't specify a defaultValue here because when creating a vote without specifying the `votingIteration` it's wrong to assume that the board is in the 0th iteration.

brandstetterm (commented):
Perhaps it's also a good idea to show a warning before starting the nth + 1 voting iteration. Something like "Warning: You are starting a new voting phase. Old votes will not be displayed"

server/src/model/init.ts
Dominik-Weinzierl: If we specify a default value, we don't need to handle the undefined case separately (incremented by one each voting iteration).

brandstetterm (approved):

PR #713, opened 2021-09-16
Fix voting notification bug

Dominik-Weinzierl:
<h2> Description </h2>

Instead of displaying the start of the voting phase (or other voting notifications) only when the voting phase started, it was displayed for every board edit event.

## Changelog 
- Add undefined check for `action.board.voting` in `boardReducer`

### DIFF
--- a/src/store/reducer/board.ts
+++ b/src/store/reducer/board.ts
@@ -13,13 +13,11 @@ export const boardReducer = (state: BoardState = {status: "unknown"}, action: Re
       };
     }
     case ActionType.EditBoard: {
-      // Moderator started voting phase - notification to moderator (user w″ho started the voting)
-      if (state.data?.voting !== action.board.voting && action.board.voting) {
-        if (action.board.voting === "active") {
-          Toast.success("You started the voting phase!");
-        } else {
-          Toast.error("You ended the voting phase!");
-        }
+      // Moderator started voting phase - notifica″tion to moderator (user who started the voting)
+      if (action.board.voting === "active") {
+        Toast.success("You started the voting phase!");
+      } else if (action.board.voting === "disabled") {
+        Toast.error("You ended the voting phase!");
       }
 
       return {
@@ -86,7 +84,7 @@ export const boardReducer = (state: BoardState = {status: "unknown"}, action: Re
     }
     case ActionType.UpdatedBoard: {
       // User notification
-      if (state.data?.voting !== action.board.voting && action.board.voting) {
+      if (state.data?.voting !== action.board.voting) {
         if (action.board.voting === "active") {
           Toast.success("Voting phase started! You can vote now!");
         } else {
--- a/src/store/reducer/board.ts
+++ b/src/store/reducer/board.ts
@@ -13,8 +13,8 @@ export const boardReducer = (state: BoardState = {status: "unknown"}, action: Re
       };
     }
     case ActionType.EditBoard: {
-      // Moderator started voting phase - notification to moderator (user who started the voting)
-      if (state.data?.voting !== action.board.voting) {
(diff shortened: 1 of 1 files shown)


src/store/reducer/board.ts
```diff
@@ -13,8 +13,8 @@ export const boardReducer = (state: BoardState = {status: "unknown"}, action: Re
       };
     }
     case ActionType.EditBoard: {
-      // Moderator started voting phase - notification to moderator (user who started the voting)
-      if (state.data?.voting !== action.board.voting) {
+      // Moderator started voting phase - notification to moderator (user w″ho started the voting)
```
brandstetterm, line 16: Typo

src/store/reducer/board.ts
```diff
@@ -13,8 +13,8 @@ export const boardReducer = (state: BoardState = {status: "unknown"}, action: Re
...
+      if (state.data?.voting !== action.board.voting && action.board.voting) {
```
brandstetterm, line 17: In theory `if (action.board.voting) {...}` should be enough. `action.board.voting` is undefined if it's not getting changed

src/store/reducer/board.ts
```diff
@@ -86,7 +86,7 @@ export const boardReducer = (state: BoardState = {status: "unknown"}, action: Re
     }
     case ActionType.UpdatedBoard: {
       // User notification
-      if (state.data?.voting !== action.board.voting) {
+      if (state.data?.voting !== action.board.voting && action.board.voting) {
```
brandstetterm, line 89: `action.board.voting` is always defined

src/store/reducer/board.ts
Dominik-Weinzierl: Thank you

src/store/reducer/board.ts
Dominik-Weinzierl: True

src/store/reducer/board.ts
Dominik-Weinzierl: Yes, that works

src/store/reducer/board.ts
```diff
@@ -13,13 +13,11 @@ export const boardReducer = (state: BoardState = {status: "unknown"}, action: Re
...
     }
     case ActionType.EditBoard: {
-      // Moderator started voting phase - notification to moderator (user who started the voting)
-      if (state.data?.voting !== action.board.voting) {
-        if (action.board.voting === "active") {
-          Toast.success("You started the voting phase!");
-        } else {
-          Toast.error("You ended the voting phase!");
-        }
+      // Moderator started voting phase - notifica″tion to moderator (user who started the voting)
```
brandstetterm, line 16:
```suggestion
      // Moderator started voting phase - notification to moderator (user who started the voting)
```

src/store/reducer/board.ts
```diff
@@ -13,13 +13,11 @@ export const boardReducer = (state: BoardState = {status: "unknown"}, action: Re
...
+      if (action.board.voting === "active") {
+        Toast.success("You started the voting phase!");
+      } else if (action.board.voting === "disabled") {
+        Toast.error("You ended the voting phase!");
```
brandstetterm, line 20: Wie wärs damit? :P

```suggestion
      if (action.board.voting) {
        Toast.success(`You ${action.board.voting === "active" ? "started" : "ended"} the voting phase!`);
      } 
```

src/store/reducer/board.ts
Dominik-Weinzierl: Interesting. Thank you.

src/store/reducer/board.ts
Dominik-Weinzierl: :P

brandstetterm (approved):

PR #727, opened 2021-09-17
Closes 660: Hide columns as moderator

Dominik-Weinzierl:
<h2> Description </h2>

Moderators are now able to toggle the visibility of columns and hide/show hidden columns.

Issue: #660 

## Checklist
- [ ] `ToggleButtonProps` is not updated by Redux update
- [ ] Add design
- [ ] Update/Add tests

---

<details>
<summary>Visual Changes - will be changed (only for testing)</summary>

Admin view
User view
</details>

### DIFF
--- a/server/src/cloud/board.ts
+++ b/server/src/cloud/board.ts
@@ -11,6 +11,9 @@ interface JoinBoardResponse {
 
 const goOnline = (user: Parse.User, boardId: string) => {
   user.add("boards", boardId);
+  if (!user.get("showHiddenColumns")) {
+    user.set("showHiddenColumns", false);
+  }
   user.save(null, {useMasterKey: true});
 };
 
--- a/server/src/cloud/column.ts
+++ b/server/src/cloud/column.ts
@@ -6,89 +6,99 @@ import Color, {isOfTypeColor} from "../util/Color";
 
 export interface AddColumnRequest {
   boardId: string;
-  name: string;
-  color: Color;
-  hidden: boolean;
+  addColumnRequest: {
+    name: string;
+    color: Color;
+    hidden: boolean;
+  };
 }
 
 export interface DeleteColumnRequest {
   boardId: string;
-  columnId: string;
+  id: string;
 }
 
 export interface EditColumnRequest {
   boardId: string;
-  columnId: string;
-  name?: string;
-  color?: Color;
-  hidden?: boolean;
+  editColumnRequest: {
+    id: string;
+    name?: string;
+    color?: Color;
+    hidden?: boolean;
+  };
 }
 
 export const initializeColumnFunctions = () => {
-  api<AddColumnRequest, boolean>("addColumn", async (user, request) => {
+  api<AddColumnRequest, {status: string; description: string}>("addColumn", async (user, request) => {
     await requireValidBoardAdmin(user, request.boardId);
 
     const board = await new Parse.Query("Board").get(request.boardId, {useMasterKey: true});
-    if (board) {
-      const columns = board.get("columns");
-      if (!isOfTypeColor(request.color)) {
-        throw new Error(`color ${request.color} is not allowed for columns`);
-      }
 
-      columns[newObjectId(serverConfig.objectIdSize)] = {
-        name: request.name,
-        color: request.color,
-        hidden: request.hidden,
-      };
-      await board.save({columns}, {useMasterKey: true});
-      return true;
+    // Check if the board exists
+    if (!board) {
+      return {status: "Error", description: `Board '${request.boardId}' does not exist`};
     }
 
-    throw new Error(`Board '${request.boardId}' not found`);
+    const columns = board.get("columns");
+    if (!isOfTypeColor(request.addColumnRequest.color)) {
+      throw new Error(`color ${request.addColumnRequest.color} is not allowed for columns`);
+    }
+
+    columns[newObjectId(serverConfig.objectIdSize)] = {
+      name: request.addColumnRequest.name,
+      color: request.addColumnRequest.color,
+      hidden: request.addColumnRequest.hidden,
+    };
+
+    await board.save(null, {useMasterKey: true});
+    return {status: "Success", description: `New column added`};
   });
 
-  api<DeleteColumnRequest, boolean>("deleteColumn", async (user, request) => {
+  api<DeleteColumnRequest, {status: string; description: string}>("deleteColumn", async (user, request) => {
     await requireValidBoardAdmin(user, request.boardId);
 
     // TODO delete notes & votes
 
     const board = await new Parse.Query("Board").get(request.boardId, {useMasterKey: true});
-    if (board) {
-      const columns = board.get("columns");
-      delete columns[request.columnId];
-      await board.save({columns}, {useMasterKey: true});
-      return true;
+    // Check if the board exists
+    if (!board) {
+      return {status: "Error", description: `Board '${request.boardId}' does not exist`};
     }
 
-    return true;
+    const columns = board.get("columns");
+    delete columns[request.id];
+    await board.save(null, {useMasterKey: true});
+
+    return {status: "Success", description: `Column ${request.id} deleted`};
   });
 
-  api<EditColumnRequest, boolean>("editColumn", async (user, request) => {
+  api<EditColumnRequest, {status: string; description: string}>("editColumn", async (user, request) => {
     await requireValidBoardAdmin(user, request.boardId);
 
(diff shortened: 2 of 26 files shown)


server/src/cloud/user.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export const initializeUserFunctions = () => {
+  /**
+   * Allows users to configure their settings.
+   */
+  api<EditUserConfigurationRequest, {status: string; description: string}>("editUserConfiguration", async (user, request) => {
+    await requireValidBoardMember(user, request.boardId);
+
+    if (request.editUserConfigurationRequest.showHiddenColumns !== undefined) {
```
brandstetterm, line 22: This setting should only be editable for moderators, shouldn't it?

src/types/note.ts
```diff
@@ -50,6 +50,7 @@ type EditableNoteAttributes = {
   columnId: string;
   parentId: string;
   text: string;
+  hidden: boolean;
```
brandstetterm, line 53: I guess that's a leftover from hide notes

src/store/reducer/__tests__/board.test.ts
```diff
@@ -0,0 +1,49 @@
+import {BoardState} from "types/store";
+import {AddColumnRequest, EditColumnRequest} from "types/column";
+import {boardReducer} from "../board";
+import {ActionFactory} from "../../action";
```
brandstetterm, line 4: pls no relative paths

server/src/cloud/board.ts
```diff
@@ -11,6 +11,11 @@ interface JoinBoardResponse {
 
 const goOnline = (user: Parse.User, boardId: string) => {
   user.add("boards", boardId);
+
+  // Add default user configuration if not already available in database
```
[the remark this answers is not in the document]
Dominik-Weinzierl: Yes, this is something we can think about in a next PR (store user configurations in local storage or create a new table for user configurations).

server/src/cloud/user.ts
Dominik-Weinzierl: This is implicit done since only moderators have access to the `HeaderMenu`. But we can add an additional check only for this configuration to be sure.

src/types/note.ts
Dominik-Weinzierl: Right. Thank you!

src/store/reducer/__tests__/board.test.ts
Dominik-Weinzierl: Fixed. (changed the imports also in `reducer/__tests__/users.test.ts`)

server/src/cloud/column.ts
```diff
@@ -18,77 +20,85 @@ export interface DeleteColumnRequest {
...
+      if (isOfTypeColor(request.column.color)) {
+        columns[request.column.columnId].color = request.column.color;
+      } else {
+        throw new Error(`specified column color '${request.column.color}' is not allowed`);
       }
+    }
 
-      await board.save({columns}, {useMasterKey: true});
-      return true;
+    if (request.column.hidden !== undefined) {
```
brandstetterm, line 96: `!=` to check for both null and undefined

server/src/cloud/user.ts
```diff
@@ -0,0 +1,31 @@
...
+    await requireValidBoardMember(user, request.boardId);
+
+    if (request.editUserConfigurationRequest.showHiddenColumns != undefined) {
+      await requireValidBoardAdmin(user, request.boardId);
+      user.set("showHiddenColumns", request.editUserConfigurationRequest.showHiddenColumns);
+    }
+
+    user.save(null, {useMasterKey: true});
+
+    return {status: "Success", description: "User was successfully removed from the list of moderators"};
```
brandstetterm, line 29: Copy Pasta

src/components/BoardHeader/HeaderMenu/HeaderMenu.tsx
```diff
@@ -18,7 +19,9 @@ type HeaderMenuProps = {
 const HeaderMenu = (props: HeaderMenuProps) => {
   const state = useSelector((applicationState: ApplicationState) => ({
     board: applicationState.board.data,
+    user: applicationState.users.all.find((user) => user.id === Parse.User.current()!.id),
```
brandstetterm, line 22: In theory, you could also get `showHiddenColumns` with `Parse.User.current().fetch()`

src/components/BoardHeader/HeaderMenu/HeaderMenu.tsx
```diff
@@ -112,6 +115,25 @@ const HeaderMenu = (props: HeaderMenuProps) => {
             <label className="item-button__label">{state.board!.showNotesOfOtherUsers ? "Hide" : "Show"} notes of other users</label>
           </button>
         </li>
+        <li className="header-menu__item">
+          <button
+            className="menu__item-button"
+            onClick={() => {
+              store.dispatch(ActionFactory.editUserConfiguration(Parse.User.current()?.id || "", {showHiddenColumns: !state.user?.showHiddenColumns}));
```
brandstetterm, line 122: As we said

Btw: `Parse.User.current()?.id ?? ""` would be the more beautiful alternative to your solution

src/components/BoardHeader/HeaderMenu/HeaderMenu.tsx
Dominik-Weinzierl: Yes, there are several solutions -> e.g. `Parse.User.current()?.toJSON() as unknown as UserClientModel` should also work, but is not supported by our tests. I think we should leave it this way.

brandstetterm:
Last friday I've talked to Benji about his PR. We came up with a solution that could be useful for this issue too. We'll add a map structure to the board scheme where for all participants the ready state will be saved. Maybe you could use this map in the future as well. We could extend it to also have your `showHiddenColumns` flag. 
So it would look like that:
`["user-1": {ready: false, showHiddenColumns: true}, "user-2": {ready: true, showHiddenColumns: false}, ...]`

server/src/cloud/column.ts
```diff
@@ -18,77 +20,85 @@ export interface DeleteColumnRequest {
...
+    columnId: string;
+    name?: string;
+    color?: Color;
+    hidden?: boolean;
+  };
 }
 
 export const initializeColumnFunctions = () => {
-  api<AddColumnRequest, boolean>("addColumn", async (user, request) => {
+  api<AddColumnRequest, {status: string; description: string}>("addColumn", async (user, request) => {
```
bitionaire, line 32: I think you should put the `{status: string; description: string}` in a general response type since it's used on multiple occasions.

server/src/cloud/user.ts
```diff
@@ -0,0 +1,31 @@
+import {requireValidBoardAdmin, requireValidBoardMember} from "./permission";
+import {api} from "./util";
+
+/**
+ * @param showHiddenColumns Allows moderators to toggle the display of hidden columns (visible only to themselves)
+ */
+type EditUserConfigurationRequest = {
+  userId: string;
+  boardId: string;
+  editUserConfigurationRequest: {
```
bitionaire, line 10: I would name this one just `userConfiguration`, since you'll probably want to override all previous settings if new settings are defined and it's not just a request? The edit and request aspect itself is already defined by the type and the name of the endpoint so it seems redundant. Just my 2 cent on the naming

src/components/Note/Note.test.tsx
```diff
@@ -183,7 +183,7 @@ describe("Note", () => {
       expect((container.querySelector(".dot-button")?.lastChild as HTMLSpanElement).innerHTML).toEqual("3");
     });
 
-    test("test-user-2 has two votes during vote phase", () => {
+    test("test-user-2 hast two votes during vote phase", () => {
```
bitionaire, line 186: Adding `t` seems unintentional

bitionaire (changes requested):

server/src/cloud/user.ts
Dominik-Weinzierl: Right. Thank you. But it will be also updated/refactored soon when #778 is done.

src/components/Note/Note.test.tsx
Dominik-Weinzierl: Yes, that was not intended ^^

server/src/cloud/column.ts
Dominik-Weinzierl: Good idea. Replaced with `StatusResponse` 👍

bitionaire (approved):

PR #728, opened 2021-09-20
Fix toggle button to always show the correct state

brandstetterm:
We've had the problem that the ToggleButton did not always show the correct state even tho the component recieved the updated props. The problem was that the internal state hasn't updated on prop change. One solution would have been to use the react hook 'useEffect' to update the state on changed props. But I've decided to remove the internal state completely because it isn't really necessary. The ToggleButton will always be in a form or something similar which will control the state correctly, so we do not need the ToggleButton to control the state itself. 

I've also decided to add some additional callback functions and some comments for better documentation and reusability.


Dominik-Weinzierl (approved):
Tested. Now works correctly 👍

Dominik-Weinzierl (approved):
Good changes. Provides us better usability. LGTM

PR #733, opened 2021-09-20
Improve tooltip animations of the ActionBar

brandstetterm:
## Description

This pull requests changes the way tooltips are shown if the action bars are at the bottom of the screen (so when not all columns are fully visible). It also fixes a bug where the action bars were still shown on the sides even tho columns were partly hidden.

The tooltip animation on screens without hover is pretty cool. But I've always thought that it doesn't feel that good on desktop screens. So I've implemented a new hover animation where the tooltip is shown directly above the button with a little arrow underneath it. 

## Changelog

- Add className to action bars container if not all columns are fully visible
- Reduce CSS to avoid code duplicates
- Update snapshots 

## Visual changes
Screenshot 2021-09-20 at 13 42 51
Screenshot 2021-09-20 at 13 42 30
Screenshot 2021-09-20 at 13 42 19


Dominik-Weinzierl (commented):
Looks good.

src/components/MenuBars/MenuBars.scss
```diff
@@ -109,7 +115,7 @@
   }
 }
 
-@media #{$menu-desktop} {
+.menu-bars:not(.menu-bars--bottom) {
```
Dominik-Weinzierl, line 118: Can be inserted in line 167 to avoid duplicate code.

src/components/MenuBars/MenuItem/MenuItem.scss
```diff
@@ -1,225 +1,229 @@
 @import "constants/style.scss";
 
+// DISABLED
+.menu-item:disabled {
+  cursor: not-allowed;
+}
+.menu-item:disabled .menu-item__icon {
+  color: #adadad;
```
Dominik-Weinzierl, line 8: We use #a3a6aa as a disabled color in several components. Should it be the same?

src/components/MenuBars/MenuItem/MenuItem.scss
```diff
@@ -1,225 +1,229 @@
...
+  padding: 0;
+  position: relative;
+  width: 36px;
+}
+
+.menu-bars--bottom {
   $menu-item__first-position-left: 28px;
 
-  @for $i from 1 through 6 {
+  @for $i from 1 through 4 {
```
Dominik-Weinzierl, line 29: Should we replace hard coded numbers with variables?

src/components/MenuBars/MenuItem/MenuItem.scss
```diff
@@ -1,225 +1,229 @@
...
-    left: 0;
-    position: absolute;
-    top: 14px;
-    border-radius: 100%;
-    transition: left 0s;
+    color: $color-white;
+    background: var(--primaryColor);
   }
+}
+.menu-item:not(:disabled):hover {
```
Dominik-Weinzierl, line 80: Can we replace `not(:disabled)` with `:enabled`? This occurs in several lines.

Dominik-Weinzierl (commented):
Some questions and changes.

brandstetterm:
@Dominik-Weinzierl Thanks for your review! I've implemented all of your suggestions

Dominik-Weinzierl (approved):

PR #754, opened 2021-09-23
Closes #675, #676: Moderationphase

### LINKED ISSUE #675: As a moderator I want to start and end the moderation phase via the admin menu

Dominik-Weinzierl:
<h2> Description </h2>

A moderator can now start or end a moderation phase and set the focus on a specific note.

Issue: #675 #676 

## Changelog
- Add `moderation` as board attribute
- Add `focus` as note attribute
- Show `NoteDialog` (for all users) if the moderator clicked on it during the moderation phase (hide other open `NoteDialog`s)

### DIFF
--- a/src/components/MenuBars/MenuBars.tsx
+++ b/src/components/MenuBars/MenuBars.tsx
@@ -30,8 +30,8 @@ function MenuBars() {
     store.dispatch(ActionFactory.editBoard({id: boardId, voting: active ? "active" : "disabled"}));
   };
 
-  const toggleModerationPhase = (active: boolean) => {
-    store.dispatch(ActionFactory.editBoard({id: boardId, moderationPhase: active ? "active" : "disabled"}));
+  const toggleModeration = (active: boolean) => {
+    store.dispatch(ActionFactory.editBoard({id: boardId, moderation: active ? "active" : "disabled"}));
   };
 
   return (
@@ -58,7 +58,7 @@ function MenuBars() {
             <MenuToggle disabled direction="left" toggleStartLabel="Start column mode" toggleStopLabel="End column mode" icon={ColumnIcon} onToggle={() => null} />
             <MenuToggle disabled direction="left" toggleStartLabel="Start timer" toggleStopLabel="Stop timer" icon={TimerIcon} onToggle={() => null} />
             <MenuToggle direction="left" toggleStartLabel="Start voting phase" toggleStopLabel="End voting phase" icon={VoteIcon} onToggle={toggleVoting} />
-            <MenuToggle direction="left" toggleStartLabel="Start focused mode" toggleStopLabel="End focused mode" icon={FocusIcon} onToggle={toggleModerationPhase} />
+            <MenuToggle direction="left" toggleStartLabel="Start focused mode" toggleStopLabel="End focused mode" icon={FocusIcon} onToggle={toggleModeration} />
           </div>
         </section>
       )}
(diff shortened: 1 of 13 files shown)


src/components/NoteDialog/NoteDialog.test.tsx
```diff
@@ -128,5 +132,27 @@ describe("NoteDialog", () => {
...
+
+    describe("Moderation phase", () => {
+      test("moderation: note-dialog__options note three isn't present", () => {
+        // @ts-ignore
+        Parse.User.current = jest.fn(() => ({id: "test-user-2"}));
+        const {container} = render(createNoteDialog("Test Text", "test-user-1", true, {userId: "test-user-1", status: true}), {
+          container: global.document.querySelector("#portal")!,
+        });
+        expect(container.querySelector(".note-dialog")?.children.item(3)?.children.length).toEqual(2);
+        expect(container.querySelector(".note-dialog")?.children.item(3)?.lastChild).not.toHaveClass("note-dialog__options");
```
bitionaire, line 144: Same as https://github.com/inovex/scrumlr.io/pull/754/files#r723622834

Please write a more explicit query. If you struggle with the selection of a specific element you can also use the `data-testid` prop end select it via the testing library https://testing-library.com/docs/queries/bytestid/
But you should use that as your last option.

src/components/NoteDialog/NoteDialog.test.tsx
```diff
@@ -128,5 +132,27 @@ describe("NoteDialog", () => {
...
+      });
+
+      test("moderation: note-dialog__options note three is present", () => {
+        // @ts-ignore
+        Parse.User.current = jest.fn(() => ({id: "test-user-2"}));
+        const {container} = render(createNoteDialog("Test Text", "test-user-1", true, {userId: "test-user-1", status: true}, true), {
+          container: global.document.querySelector("#portal")!,
+        });
+        expect(container.querySelector(".note-dialog")?.children.item(3)?.children.length).toEqual(3);
```
bitionaire, line 153: Same as https://github.com/inovex/scrumlr.io/pull/754/files#r723622834 -> more explicit query

src/components/MenuBars/MenuBars.tsx
```diff
@@ -75,7 +80,14 @@ function MenuBars() {
             <MenuToggle disabled direction="left" toggleStartLabel="Start column mode" toggleStopLabel="End column mode" icon={ColumnIcon} onToggle={() => null} />
             <MenuToggle value={state.timer != null} direction="left" toggleStartLabel="Start timer" toggleStopLabel="Stop timer" icon={TimerIcon} onToggle={toggleTimer} />
             <MenuToggle direction="left" toggleStartLabel="Start voting phase" toggleStopLabel="End voting phase" icon={VoteIcon} onToggle={toggleVoting} />
-            <MenuToggle disabled direction="left" toggleStartLabel="Start focused mode" toggleStopLabel="End focused mode" icon={FocusIcon} onToggle={() => null} />
+            <MenuToggle
```
bitionaire, line 83: I think there's another icon required here or maybe even another background, to make it more explicit to the user that the focus mode is in active state.
Also we need some sort of tooltip what "focus" means or you could probably describe it in the toast? I don't know what would be the best option. We can discuss this together

bitionaire: ... also the mouse cursor might change to `pointer` on note hover while in focus mode? Or another cursor... maybe `zoom` or `copy`?  What do you think?

PR #771, opened 2021-09-27
Closes #759: `MenuBars` animation bug

### LINKED ISSUE #759: Css `MenuBars` bug when switching between side and bottom bar
## Description
When switching from the sidebar to the bottom bar, the admin panel is displayed above the user menu for a short time instead of only the user menu.

## Expected behavior
Displaying only the user menu.

## How to reproduce
Select the user menu. Switch from the sidebar to the bottom bar.

Dominik-Weinzierl:
<h2> Description </h2>

When switching from the sidebar to the bottom bar, the admin/user panel is displayed above/below the user/admin menu for a short time instead of only the user/admin menu.

Issue: #759 

## Changelog
- Extract transition into `.menu-animation`
- Add `onTransitionEnd()` to toggle end of animation and remove class
- Filter events based on `attributes` (`classNames: menu-animation`)
- Start animation when pressing the switch button (instead of `ease-in-out`)


src/components/MenuBars/MenuBars.tsx
```diff
@@ -54,7 +55,7 @@ function MenuBars() {
         {"menu-bars--bottom": document.getElementById("menu-bars")?.classList.contains("menu-bars--bottom")}
       )}
     >
-      <section className="menu user-menu">
+      <section className={classNames("menu", "user-menu", {"menu-animation": animate})} onTransitionEnd={() => setAnimate(false)}>
```
bitionaire, line 58: Since onTransitionEnd will be called upon each CSS property `setAnimate(false)` will be called after `0.4s`. You can check for the CSS property name here and read it from the `event` which will be passed to the inner function before calling `setAnimate` or use the React animations addon here https://reactjs.org/docs/animation.html

Seems like a bit of overhead but would be the proper solution here I guess

bitionaire (changes requested):

src/components/MenuBars/MenuBars.tsx
bitionaire: ... or you can try to unify the transition durations

Head explodes

bitionaire (changes requested):

src/components/MenuBars/MenuBars.tsx
Dominik-Weinzierl: Good point. Thank you. Didn't thought about `event bubbling`. I will consider all the suggestions and choose the most promising one 😄

Dominik-Weinzierl: I decided to filter the events based on their `classNames`. But we can keep the second approach (`ReactTransitionGroup`) in mind and replace the current solution if needed in the future.

src/components/MenuBars/MenuBars.tsx
```diff
@@ -54,7 +61,7 @@ function MenuBars() {
         {"menu-bars--bottom": document.getElementById("menu-bars")?.classList.contains("menu-bars--bottom")}
       )}
     >
-      <section className="menu user-menu">
+      <section className={classNames("menu", "user-menu", {"menu-animation": animate})} onTransitionEnd={(event) => handleAnimate(event)}>
```
andiKandi, line 64: onTransitionEnd={() => setAnimate(false)} does the same as handleAnimate. Or is the check for event.currentTarget.attributes.getNamedItem("class")?.nodeValue?.includes("menu-animation") necessary?

src/components/MenuBars/MenuBars.tsx
```diff
@@ -63,7 +70,7 @@ function MenuBars() {
         </div>
       </section>
       {isAdmin && (
-        <section className="menu admin-menu">
+        <section className={classNames("menu", "admin-menu", {"menu-animation": animate})} onTransitionEnd={(event) => handleAnimate(event)}>
```
andiKandi, line 73: see above

andiKandi (approved):

src/components/MenuBars/MenuBars.tsx
Dominik-Weinzierl: Yes, as mentioned above in the conversation with Johann, this check is needed to react only to events from the component with the class "menu-animation" (`event bubbling`).

PR #772, opened 2021-09-27
Closes #760: Css offset bug

### LINKED ISSUE #760: Css scrollbar issue with columns and notes
## Description
There seems to be some css offsets when the board is minimized.

## How to reproduce
1: Make the board as small as possible, you you need to see at least the scrollbar of the columns
2: If the scroll bar exceeds the border (outside the window), you can scroll down and see a white screen (offset error)

## Screenshots
1: Here you can see how the scroll bar extends beyond the border.
Overlapping scrollbar

2: White space at the bottom of the board.
White boarder

Dominik-Weinzierl:
<h2> Description </h2>

There seems to be some css offsets when the board is minimized. This should be fixed now.

Issue: #760

## Changelog

- Add `overflow: hidden` to Board.scss (`.board`)
- Replace `padding` with `margin-top`


bitionaire (approved):
Haven't tested it but this changes is so small that it LGTM :)

PR #785, opened 2021-09-30
Change the touchscreen behaviour of the action bar buttons

brandstetterm:
I've implemented the 2-click behaviour for our action bar buttons. 
1st click: Hover
2nd click: Actual click 

https://user-images.githubusercontent.com/36969812/135413270-99c67783-d177-4ce2-880c-d4bd12152915.mov


src/components/MenuBars/MenuItem/MenuButton.tsx
```diff
@@ -10,10 +11,29 @@ type MenuButtonProps = {
 };
 
 function MenuButton(props: MenuButtonProps) {
+  const [clickCount, setClickCount] = useState(0);
   const Icon = props.icon;
 
+  useEffect(() => {
+    if (clickCount > 0) {
```
Dominik-Weinzierl, line 18: Works probably not as expected 🤔

Dominik-Weinzierl (commented):
I am not sure if this was a bug in firefox, but it seems to be possible to display both tool tips.

Screenshot 2021-09-30 at 11 47 51

src/components/MenuBars/MenuItem/MenuButton.tsx
brandstetterm: Yes, that's not the intended behaviour and I've already located the issue. The problem is that currently I prevent the default behaviour in the `touchstart` event so that the `click` event doesn't get fired. But this also prevents the normal event bubbling and therefore a touch event never reaches the window event listener. 

I have to think about a possible workaround / alternative.

PR #790, opened 2021-10-02
Overflow ellipsis on board header

bitionaire:
<h2> Overflow ellipsis on board header</h2>

Board header title was not limited to max-width and could wrap before. I added the overflow settings and restricted the width

---
<details>
<summary>Visual Changes</summary>

Before
After

</details>

### DIFF
--- a/src/components/BoardHeader/BoardHeader.scss
+++ b/src/components/BoardHeader/BoardHeader.scss
@@ -59,7 +59,13 @@
   font-weight: bold;
   letter-spacing: 0.5px;
   line-height: 40px;
-  margin: 0px;
+  margin: 0;
+
+  text-overflow: ellipsis;
+  overflow: hidden;
+  white-space: nowrap;
+  width: calc(100vw - 260px);
+  max-width: 28vw;
 }
 @media (prefers-color-scheme: dark) {
   .board-header__title {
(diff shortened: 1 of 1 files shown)


src/components/BoardHeader/BoardHeader.scss
```diff
@@ -59,7 +59,13 @@
   font-weight: bold;
   letter-spacing: 0.5px;
   line-height: 40px;
-  margin: 0px;
+  margin: 0;
+
+  text-overflow: ellipsis;
+  overflow: hidden;
```
Dominik-Weinzierl, line 65: You could add `text-align: center;` here. And to avoid some offsets with the icon, we can add padding in line 84.

```
  .board-header__title {
    padding-right: 24px;
  }
 ```
 
Screenshot 2021-10-03 at 17 32 38

Dominik-Weinzierl (commented):
Seems like the board header name is not centered anymore.
Screenshot 2021-10-03 at 17 16 49

In comparison the main branch:
Screenshot 2021-10-03 at 17 18 05

src/components/BoardHeader/BoardHeader.scss
bitionaire: Good catch! I forgot to remove the `width` prop and it wasn't centered anymore 👍

Dominik-Weinzierl (approved):
LGTM

PR #791, opened 2021-10-02
Removed redundant font-family

bitionaire:
<h2> Removed redundant font-family</h2>

`font-family` is already set on `body` so it's redundant in the places I removed it.


Dominik-Weinzierl (commented):
Is
```
body {
  font-family: "Raleway", sans-serif;
}
```
it still needed even though we are already using the universal selector🤔 
```
*,
*:before,
*:after {
  font-family: "Raleway", sans-serif;
}
```

andiKandi (approved):

PR #891, opened 2021-10-14
Add root attribute for column visibility

brandstetterm:
## Description

I've came up with the idea to add an attribute for the column visibility to our root element. This way we don't need to toggle classes or something else if the visibility of columns change. There has also been a problem with toggling classes and optional classes from classNames.


Dominik-Weinzierl:
Good catch @andiKandi. I think this is not intended since the moderator can still hide columns for other users, but can't toggle the visibility of hidden columns for himself.

PR #894, opened 2021-10-14
Closes #893: Remove multiple occurrences of footer/header/main

### LINKED ISSUE #893: Remove multiple occurrences of footer/header/main

Dominik-Weinzierl:
<h2> Description </h2>

Removed multiple occurrences of footer/header/main as they should only occur once.

Issue: Closes #893 

## Changelog
- Replace multiple occurrences of `<footer/>`/`<header/>`/`<main/>` with `<div/>`

### DIFF
--- a/src/components/BoardHeader/HeaderMenu/HeaderMenu.scss
+++ b/src/components/BoardHeader/HeaderMenu/HeaderMenu.scss
@@ -6,11 +6,10 @@ $header-menu-width: 288px;
 $header-menu__item-height: 48px;
 
 .header-menu {
-  position: fixed;
   width: 288px;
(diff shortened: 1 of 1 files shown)


andiKandi:
Hi, the way I understand this documentation is that you are allowed to have several header/ footers but should use them appropriately: https://html.spec.whatwg.org/multipage/sections.html#the-header-element

Dominik-Weinzierl:
Yes, this is possible, but only wrapped in an article. Otherwise it is still convention to use it only once on the complete page. But correct me if i am wrong - e.g. Johanns comment https://github.com/inovex/scrumlr.io/pull/777#discussion_r727161898

PR #974, opened 2021-10-28
Alignment of vote buttons 

bitionaire:
Fixes issue where vote count has a large margin to the right side


Dominik-Weinzierl (approved):
Code looks good to me, but i am not sure if this behavior is intended 🤔 

https://user-images.githubusercontent.com/60005702/139578974-daf83762-2080-4864-a98f-ae35c1a999c9.mov

PR #975, opened 2021-10-31
Fixed menu bars visibility for small screens

### LINKED ISSUE #969: Entering a board on mobile device initializes wrong layout of menu bars
I will attach a screenshot

localhost_3000_board_z7XZHLjCOEGnUZVMH4MEoAmZp2haOjsQ(iPhone 6_7_8)

This happens  in lean coffee mode. Since a single column will not split on multiple views the `column-visibility` is always set to `visible` and therefore the bars and the timer will be rendered in the wrong place

bitionaire:
Closes #969 (checkout for more details).

This is just a suggestion and a quick and almost clean fix, it's not perfect. 
Perfect would be a rewrite of parts of the css but it's rather complicated.

### DIFF
--- a/src/components/MenuBars/MenuBars.scss
+++ b/src/components/MenuBars/MenuBars.scss
@@ -5,7 +5,7 @@
   z-index: $menu-z-index;
 }
 
-[column-visibility="collapsed"] .menu-bars {
+@include collapsedColumns(".menu-bars") {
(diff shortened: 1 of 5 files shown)


Dominik-Weinzierl (approved):
LGTM 👍 - works as expected and is a good current solution

PR #1664, opened 2022-05-12
New columns should be added as hidden

timengel:
## Changelog

- New columns are hidden by default
- If visibility of hidden columns is turned OFF, it is set to ON upon adding a column (when this happens, a toast is shown notifying the user that the visibility of the hidden columns has changed)
- Off-topic: Hidden columns now have a different header color to distinguish them more easily from visible columns (see below)

image

## Checklist


src/components/Column/ColumnSettings.tsx
```diff
@@ -10,6 +10,9 @@ import {Color, getColorClassName} from "constants/colors";
 import {useTranslation} from "react-i18next";
 import {useDispatch} from "react-redux";
 import "./ColumnSettings.scss";
+import i18n from "i18next";
+import {useAppSelector} from "../../store";
+import {Toast} from "../../utils/Toast";
```
brandstetterm, line 15: Please try to use absolute instead of relativ paths

brandstetterm (approved):

PR #1673, opened 2022-05-15
Closes #1672: Added release pipeline

### LINKED ISSUE #1672: Add support for releases
As a user I want to be able to use specific application images instead of using the `latest` tag.

We have to add a GitHub action which will trigger on a manual release and set the docker images tag upon execution of the pipeline. See https://github.com/docker/build-push-action/blob/master/docs/advanced/tags-labels.md for reference on how to use the action plugin.

bitionaire:
Closes #1672

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,57 +1,14 @@
-name: Continuous Integration
+name: Package release
 
 on:
-  push:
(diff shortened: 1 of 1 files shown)


louiskroener (approved):
lgtm

PR #1786, opened 2022-06-08
Fix voting sorting

brandstetterm:
## Description

The problem was that the sorting did not include the votes on the stacked notes. So the sorting was basically just dependent on the visible notes without the stacked notes.

## Changelist

- Change SQL to calculate the number of votes over the note itself and its stacked notes (For every note on the board, the number of rows is counted from the notes inner joined with the votes of the current voting).
- Add unit test to check correct note rank after voting has been closed.
- Add fixture test data (User, Board, Column, Voting, Notes and Votes)


bitionaire (changes requested):
LUTM (Looks ugly to me) 😅 

Joking aside, can you add a unit test for that please? I know that it's a pain in the ass to add a test fixture (and I'm sorry that I request it) but nevertheless I think it's a good idea for this code in case we have to touch it in the future.

bitionaire (approved):

PR #1787, opened 2022-06-08
Add docker-compose file which includes all dev profile services for lima

brandstetterm:
Since lima doesn't support the `--profile dev` option with their `lima nerdctl compose` command, I've added an additional compose file which can be used with `lima nerdctl compose --file dev-docker-compose.yml up`


CronJorian (approved):
First line is empty and can be removed.

PR #1793, opened 2022-06-09
Fix column scroll on mobile

### LINKED ISSUE #1725: Column header should not scroll on mobile
image

CronJorian:
## Description

Closes #1725 
Change height for board and columns to 100% instead of 100vh and handle vertical overflow.

## Changelog

Updated 2 CSS selectors and added 1 more.

## Checklist

### DIFF
--- a/src/components/Board/Board.scss
+++ b/src/components/Board/Board.scss
@@ -16,7 +16,7 @@
   scroll-padding: $board__side-panel-width;
 
   width: 100vw;
-  height: 100vh;
+  height: 100%;
 
   -ms-overflow-style: none;
   scrollbar-width: none;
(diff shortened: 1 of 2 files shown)


bitionaire (approved):

bitionaire:
You can merge it but it does not change the behaviour. You can still scroll the background while the header stays sticky :D

CronJorian:
@bitionaire but this seems to be an iOS "feature". Couldn't reproduce it on Android, Chrome, Safari or Firefox. Or is there a misunderstanding?

bitionaire:
@CronJorian yes, this is a special iOS feature. In order to fix it we would need to position the header `relative` instead of `fixed`. Also the columns should stretch the page vertically. Currently you must target the note area in order to scroll and cant scroll the whole page vertically, which seems very weird.

PR #1794, opened 2022-06-09
Change timer sounds

timengel:
## Changelog

- [x] Added the `use-sound` package (supposedly only "1kb bytes (gzip) in your bundle! ~10kb loaded async"). Its `useSound` hook enables the following changes.
- [x] Timer sounds are now unaffected by hardware media keys (afaik)
- [x] Countdown sound starts 6 seconds before the timer ends. Afterwards, the times-up sound is played. The scheme is as follows: `tik - tok - tik - tok - tik - tok - boom`. This felt more natural than `tik - tok - tik - boom`.
- [x] Timer sounds dont re-play after the page is reloaded without closing a finished timer
- [x] Sound volume reduced

## Checklist


CronJorian (commented):
Works but needs some finetuning for the sound imo. Briefly before the `boom` an additional `tik` is audible.

CronJorian:
This might be resolved by not triggering the second sound by useEffect (due to its asynchrone nature)

brandstetterm:
Btw. yesterday during the AD exchange retro perspective, I heard that Miro only plays a sound once the timer has ended. To be honest I really like that.

timengel:
> Btw. yesterday during the AD exchange retro perspective, I heard that Miro only plays a sound once the timer has ended. To be honest I really liked that.

Yes, I support that. The `tik-tok`-sound's initial purpose of indicating that board members should hurry up is gone anyways due to the short tik-tok-period. Therefore, one sound at the end is enough.

CronJorian (approved):

bitionaire (approved):

PR #1873, opened 2022-06-21
Fix json export

### LINKED ISSUE #1861: JSON export with file content [object Object]

brandstetterm:
Closes #1861 

- Wrap json object in `JSON.stringify()`
- Add `scrumlr.io` to name if board name is undefined


CronJorian (approved):
Nice. 👏 

FYI as member you can export the board to see hidden columns and their notes. IDK if there should be a permission check and only the available content should be output.
If so, please add it as issue (i can do it too)

PR #1880, opened 2022-06-22
Remove edit icon from note hover, add outline to note focus

brandstetterm:
- Remove edit icon on hover
- Add coloured outline to focus and focus-within (50% opacity just as in the other PR)

Screenshot 2022-06-22 at 15 15 20


src/components/Note/Note.scss
```diff
@@ -58,7 +58,9 @@ $note__author-image-size: 26px;
 }
 
 .note:focus-within,
-.note:focus,
+.note:focus {
+  outline: 2px solid rgba(var(--accent-color-rgb), 0.5);
+}
 .note:hover {
   transform: scale(1.02);
   outline: none;
```
Resaki1, line 66: is it intended behavior that the outline of a focused card is removed on hover? I think it looks weird

src/components/Note/Note.scss
```diff
@@ -58,7 +58,9 @@ $note__author-image-size: 26px;
 }
 
 .note:focus-within,
-.note:focus,
+.note:focus {
+  outline: 2px solid rgba(var(--accent-color-rgb), 0.5);
```
Resaki1, line 62: full opacity looks better imo, because it matches the small colored strip on the left side of the cards. What do you think?

src/components/Note/Note.scss
brandstetterm: I aggree, it looks better if it stays

Resaki1 (approved):

PR #1916, opened 2022-07-03
Fix default language in settings dialog

### LINKED ISSUE #1862: Wrong (default) language selection
English is selected in settings dropdown though the default interface language is set to "German".

seschi98:
## Description

This fixes #1862. Because the actual language detected by i18n is `de-DE` and not `de`, the language dropdown in the settings dialog will fallback to `en` (in the else-Branch). But if we use `i18n.resolvedLanguage` instead of `i18n.language`, it will work.

## Changelog

- Switched from `i18n.language` to `i18n.resolvedLanguage` for the default dropdown value

## Checklist


CronJorian (approved):
Thank you for contributing 🥳🚀

Issue #1722, opened 2022-05-22
Change number of touches for action bar


bitionaire:
While a button is focused/active in the action bar only one tap should be necessary to toggle it. Currently one need two taps to disable eg the ready state though it seems active

- change to long-press to show tooltip, single press to activate action

![image](https://user-images.githubusercontent.com/1539948/169691075-8422716c-9836-4bc6-8f72-4607cdfd1d19.png)

brandstetterm:
Remove tooltips for mobile for now (but think about a setting where you can reactive them)

Resaki1:
closed as tooltips won't be shown on mobile anymore

PR #2000, opened 2022-07-21
Implement new design of actions bars on mobile

timengel:
## Description

Implementation of the redesigned menu bars for mobile screens according to [this](https://xd.adobe.com/view/6e4e8128-3c24-4e24-8409-56c192e0b25b-3b1d/screen/b162f71e-e943-43a3-a337-c8f24b14437f/) design template.

## Changelog

- [x] Menu bars transform into expandable Fab on smaller screens
- [x] Change the color of disabled icons for better visibility

## ToDo

- [x] ~~Disable tooltips for action bar on mobile screens or enable them by long press (https://github.com/inovex/scrumlr.io/issues/1722)~~
- [x] Layout the bars vertically and horizontally on mobile screens hidden in a burger menu
- [x] Add activated buttons to fab in contracted state
- [x] Find a better way to animate the expanding Fab options
- [x] Remove old menu switch
- [x] Fix Voting & Timer Menu on screens between smartphone and tablet size

## Checklist


src/components/MenuBars/MenuBars.scss
```diff
@@ -60,77 +203,10 @@ $box-shadow: 0px 10px 20px #00000026;
...
-}
-.switch__icon {
-  position: absolute;
-  top: 0;
-  left: 0;
-  transform: rotate(0deg);
-  transition: opacity 0.5s, visibility 0.5s, transform 1s;
-  color: $color-black;
-}
 .menu-animation {
```
Resaki1, line 206: Is this still needed? If not, there's also some dead code in `MenuBars.tsx`, such as the `animate` state, the `handleAnimate` function and some imports

src/components/MenuBars/MenuItem/MenuItem.scss
```diff
@@ -161,14 +158,14 @@ $icon-background-color-dark: #333948;
 [theme="dark"] {
   .tooltip__text {
     color: $color-white !important;
-    background-color: $tooltip-background-color-dark !important;
+    background-color: $tooltip-background-color--dark !important;
   }
 }
 
 .menu-item__icon {
```
Resaki1, line 165: https://user-images.githubusercontent.com/35272402/181040213-58c8c8f9-e49f-493a-9f51-3705c0e6eebe.mov

The icons for focus mode, ready, and raised hand don't change back from the cross to the initial icon as fast as the background changes, which feels weird imo. Is that fixable?

src/constants/style.scss
```diff
@@ -7,15 +7,23 @@ $color-white-one: #f9fafb;
...
+
+// element colors
 $color-dark-mode: #272f41;
+$color-dark-mode--disabled: #1b202d;
 $color-dark-mode-note: #313949;
 $color-dark-mode-note--hover: #3d4555;
+$menu-icon-background-color--dark: #333948;
+$menu-icon-background-color--light: $color-white;
+$tooltip-background-color--light: #d3daf0;
+$tooltip-background-color--dark: #485064;
```
Resaki1, line 26: Do we need all those extra colors? We already have many different hues, can't we just re-use them? Would make for a cleaner look overall imo

src/components/MenuBars/MenuBars.tsx
```diff
@@ -110,74 +111,151 @@ export const MenuBars = ({showPreviousColumn, showNextColumn, onPreviousColumn,
...
+            onClick={onPreviousColumn}
+            aria-hidden
+          >
+            <LeftArrowIcon className="menu-bars__navigation-icon" />
+          </button>
+        </section>
+
+        <section className={classNames("menu", "admin-menu", {"admin-menu--empty": !isAdmin, "menu-animation": animate})} onTransitionEnd={(event) => handleAnimate(event)}>
+          {isAdmin && (
+            <div className="menu__items">
```
Resaki1, line 151: can this be a `ul` with each Button/Toggle being an `li`, just like the other menu? Would be better for a11y

src/components/MenuBars/MenuItem/MenuItem.scss
```diff
@@ -15,6 +11,7 @@ $icon-background-color-dark: #333948;
   padding: 0;
   position: relative;
   cursor: pointer;
+  pointer-events: auto;
```
Resaki1, line 14: isn't it `auto` by default?

src/components/MenuBars/MenuBars.tsx
```diff
@@ -110,74 +111,151 @@ export const MenuBars = ({showPreviousColumn, showNextColumn, onPreviousColumn,
...
-          onClick={onNextColumn}
-          aria-hidden
+          className={classNames("menu-bars-mobile__fab menu-bars-mobile__fab-main", {"menu-bars-mobile__fab-main--isExpanded": fabIsExpanded})}
+          onClick={() => {
+            setFabIsExpanded(!fabIsExpanded);
+          }}
+          aria-label={t("MenuBars.openMenu")}
         >
-          <RightArrowIcon className="menu-bars__navigation-icon" />
+          {fabIsExpanded ? <CloseIcon aria-hidden /> : <MenuIcon className="menu-bars-mobile__fab-main-icon" aria-hidden />}
```
Resaki1, line 184: that class (`fab-main-icon`) isn't used anywhere as far as I can tell

Resaki1, line 184: Bildschirmfoto 2022-07-26 um 17 23 12

On WebKit (all iOS Browsers), the close Icon is the same color as the background and therefore not visible (both light and dark mode)

Resaki1: Also, what do you think about the menu closing automatically when pressing outside of it? We have an clickOutside event handler already implemented in `ColumSettings.tsx:37`

Resaki1 (changes requested):
I like the overall look and feel it's a good improvement of the mobile UX 👌🏼

src/components/MenuBars/MenuBars.tsx
timengel: Agreed, I have refactored it.

src/components/MenuBars/MenuItem/MenuItem.scss
timengel: No, somehow the element inherited `none` from the container. But I have changed the grid container anyways, so `pointer-events` is gone :)

src/components/MenuBars/MenuBars.tsx
timengel: The icon had the correct color but somehow the width/height `auto` sizing was buggy in WebKit. I set the size manually and now it works :)

src/constants/style.scss
timengel: Mhm I just compared the colors with the existing ones and they are different i.e. unique. E.g. `$menu-icon-background-color--dark` is darker than `$color-darker-gray` but visibly lighter than `$color-darkest-gray` (and also a little blue-ish).

Btw, I just moved the constants from `MenuBars.scss` to `styles.scss` bc some were used in multiple files.

Didnt someone refactor the MenuBar colors for better visibility just recently?

timengel: I just talked to @bitionaire. We need those colors in the future, so we cant replace them but we should try to fit them in our current color scheme. Since they dont fit easily (see example above), this should be handled in a separate PR.

src/components/MenuBars/MenuItem/MenuItem.scss
timengel: There was rogue property of `transition: visibility 150ms linear;` that got overwritten on larger screens but was active on smaller ones. Now it works :)

Resaki1 (approved):
Looks good, thanks for the fixes.It'd be nice if

timengel:
> Looks good, thanks for the fixes.
> 
> It'd be nice if the menu would close automatically if the users press outside of it, and maybe even when pressing one of the menu buttons (i.e., the ready button). But we can implement that later (maybe even with a nice animation of the menu/close icons?) and merge this branch now

Oh sorry, I forgot about that 🙈  I have implemented a close-on-click-outside with the hook you mentioned.

Issue #1713, opened 2022-05-21
Replace use of tabindex with better document structure


bitionaire:
As a developer I want to get rid of the usage of `tabindex` and instead replace it by a better structure of the document, which will lead to the current tab order.

The usage of `tabindex` might be misleading for users and also it tends to break once we start changing up elements.

PR #2008, opened 2022-07-26
Cr/focus states

Resaki1:
## Description

Many interactive elements didn't have proper focus states and some couldn't be navigated to via keyboard at all. This has been fixed.

## Changelog

- added missing focus states
- improved some existing focus states
- made some interactive elements focusable (e.g., settings dropdown)
- moved from `.focus` to `.focus-visible` for better UX
- improved print icon (not in scope but I just had to...)
- fixed board templates not being checked anymore after navigating back from extended config

## Checklist

## Visual Changes (examples)
Options of SettingsDropdown are focusable now:
Bildschirmfoto 2022-07-26 um 16 26 51

Before (focus is on Participants):
Bildschirmfoto 2022-07-26 um 16 25 17

After:
Bildschirmfoto 2022-07-26 um 16 26 09

Before (hover and focus states are identical):
https://user-images.githubusercontent.com/35272402/181032888-3dc02673-d446-4ab8-9aea-a9154f5e50cd.mov

After (separate states):
https://user-images.githubusercontent.com/35272402/181033352-8ffb10ff-07c1-4813-b181-4517be31e573.mov

### DIFF
--- a/src/components/SettingsDialog/Components/LanguageSettingsDropdown.scss
+++ b/src/components/SettingsDialog/Components/LanguageSettingsDropdown.scss
@@ -1,4 +1,5 @@
 @import "constants/style";
+@import "./SettingsButton";
 
 .settings-dropdown {
   display: flex;
@@ -29,7 +30,7 @@
 .settings-dropdown-content {
   position: absolute;
   width: calc(100% + 20px);
-  top: 48px;
+  top: calc($settings__item-height + 1px);
   left: -16px;
   border-radius: 8px 8px 8px 8px;
   padding: 0;
@@ -39,6 +40,7 @@
 .settings-dropdown-content li {
   display: flex;
   align-items: center;
+  outline: none;
   width: 100%;
   color: $color-black;
   padding: $padding--default $padding--medium $padding--default $padding--default;
@@ -46,27 +48,34 @@
   font-weight: bold;
   background-color: $color-white;
   gap: 12px;
+  border: 2px solid transparent;
   border-radius: 8px 8px 8px 8px;
+
+  &:last-child {
+    border-radius: 0px 0px 8px 8px;
+  }
+
+  &:focus-visible {
+    border: 2px solid rgba(var(--accent-color-rgb), 0.3);
+  }
+
+  &:hover {
+    background-color: rgba(var(--accent-color-rgb), 0.1);
+  }
 }
 
 [theme="dark"] {
   .settings-dropdown-content li {
     background-color: $color-dark-mode-note;
     color: $color-white;
-  }
-}
 
-.settings-dropdown-content li:last-child {
-  border-radius: 0px 0px 8px 8px;
-}
+    &:focus-visible {
+      border: 2px solid rgba($color-white, 0.6);
+    }
 
-.settings-dropdown-content li:hover {
-  background-color: rgba(var(--accent-color-rgb), 0.1);
-}
-
-[theme="dark"] {
-  .settings-dropdown-content li:hover {
-    background-color: $color-dark-mode-note--hover;
+    &:hover {
+      background-color: $color-dark-mode-note--hover;
+    }
   }
 }
 
--- a/src/components/SettingsDialog/Components/SettingsCarousel.scss
+++ b/src/components/SettingsDialog/Components/SettingsCarousel.scss
@@ -29,12 +29,17 @@
     justify-items: center;
     align-items: center;
     border: none;
+    outline: none;
     color: var(--accent-color);
     transition: all 0.08s ease-out;
 
     &:hover {
       filter: brightness(1.4);
       transform: scale(1.1);
+    }
+
+    &:focus-visible {
+      filter: brightness(1.4);
     }
 
     svg {
--- a/src/components/SettingsDialog/Components/SettingsInput.scss
+++ b/src/components/SettingsDialog/Components/SettingsInput.scss
@@ -17,7 +17,7 @@
     font-weight: bold;
     padding-top: 10px;
 
-    &:focus,
+    &:focus-visible,
     &:hover {
       background-color: rgba(var(--accent-color-rgb), 0.1);
     }
@@ -39,7 +39,6 @@
 
   label {
     margin: $margin--default $margin--medium;
-    z-index: 1;
     position: absolute;
     transition: all 0.08s ease-out;
     cursor: text;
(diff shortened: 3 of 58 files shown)


timengel:
Is there a way to navigate the settings options on the left side using only the keyboard? Rn `tab/shift+tab` loops through the content on the right and back but afterwards it doesnt progress the options list.
image

timengel:
The keyboard navigation on the homepage works fine in Blink/Chromium browsers but in Firefox it only targets the header (languages, login, etc.) and skips the footer (app-info), and in Safari it doesnt target the page itself at all and jumps right into the URL field.

Resaki1:
> The keyboard navigation on the homepage works fine in Blink/Chromium browsers but in Firefox it only targets the header (languages, login, etc.) and skips the footer (app-info), and in Safari it doesnt target the page itself at all and jumps right into the URL field.
> 
> Edit: On MacOS this problem can be fixed by checking this box in System Preferences > Keyboard Settings: image

Thanks for the info. This seems to be a firefox on MacOS specific issue though, I don't think there's anything we can do about it: https://support.mozilla.org/en-US/questions/1278793

src/components/SettingsDialog/Components/SettingsCarousel.scss
```diff
@@ -37,6 +38,10 @@
       transform: scale(1.1);
     }
 
+    &:focus-visible {
+      filter: brightness(1.4);
```
CronJorian, line 42: See brightness comment

src/components/SettingsDialog/Components/SettingsInput.scss
```diff
@@ -17,7 +17,7 @@
     font-weight: bold;
```
CronJorian, line 17: `position: absolute;` can to be removed. ATM the component has a height of 0px in the `ProfileSettings` component. Although the `SettingsInput` in the `BoardSettings` is fine and has the correct height, this is not the case on the `ProfileSettings`.
If there is a reason for using `absolute` positioning, this is fine, but imo it's not needed.

src/components/SettingsDialog/Components/ThemeSettingsDropdown.tsx
```diff
@@ -7,18 +7,38 @@ import {ReactComponent as DarkIcon} from "assets/icon-darkmode.svg";
...
+  const [theme, setTheme] = useState<Theme>((document.documentElement.getAttribute("theme") as Theme) ?? Theme.LIGHT);
   useEffect(() => {
     document.documentElement.setAttribute("theme", theme!);
     localStorage.setItem("theme", theme!);
   }, [theme]);
 
+  const handleKeyDown = (newTheme: Theme) => (e: KeyboardEvent<HTMLLIElement>) => {
+    switch (e.key) {
+      case " ":
+      case "SpaceBar":
```
CronJorian, line 28: See previous comment to this.
```suggestion
      case "Spacebar":
```

src/routes/Homepage/Homepage.scss
```diff
@@ -31,26 +31,18 @@
...
 .homepage__language {
-  background: $color-lighter-gray;
-}
-
-html:not([theme="dark"]) {
-  .homepage__language {
-    box-shadow: 0 0 0 0 rgba($color-middle-gray, 0) !important;
-  }
+  background: $color-light-gray;
+  box-shadow: 0 0 0 0 rgba($color-middle-gray, 0);
```
CronJorian, line 35:
```suggestion
  box-shadow: 0 0 rgba($color-middle-gray, 0);
```
`blur-radius` and `spread-radius` default to 0, so the values can be omitted

src/routes/Legal/Legal.scss
```diff
@@ -11,10 +11,24 @@
...
+      transform: scale(1.08);
+    }
+  }
+
+  &:hover,
+  &:focus-visible {
     text-decoration: underline;
+
+    > svg {
+      filter: brightness(1.4);
```
CronJorian, line 30: See brightness comment.

src/routes/LoginBoard/LoginBoard.scss
```diff
@@ -13,6 +13,14 @@
   padding: $padding--default;
 }
 
+.login-board__form a {
+  outline: none;
+
+  &:focus-visible svg {
+    filter: brightness(1.6);
```
CronJorian, line 20: See brightness comment

src/routes/NewBoard/NewBoard.scss
```diff
@@ -13,6 +13,18 @@
...
+
+  a {
+    outline: none;
+
+    &:hover svg {
+      filter: drop-shadow(4px 4px 4px $color-middle-gray);
+    }
+
+    &:focus-visible > svg {
+      filter: brightness(1.6);
```
CronJorian, line 25: See brightness comment.

CronJorian (changes requested):
Not sure if this is intended, but if I go through the components by using TAB, at first all the column inputs come and then all the notes. If this is intended its fine, I was just confused and expected to go through every column, first input and then the notes of each column.

Additionally it is not possible by keyboard only to add notes. The input field is focussed, but no text input is possible (also not by pressing enter etc).

On the homepage a new element is focussed which is not visible at all. (Between the URL bar and the flags).

Maybe it's reasonable to create values for scale, brightness and some other things in the `style.scss` and reuse these values. It seems you use many different values for these functions and I don't know if this is always justified.

CronJorian:
Additionally it might be more pleasant to add a border around the scrumlr logo instead of just highlighting it.

src/components/SettingsDialog/Components/LanguageSettingsDropdown.scss
```diff
@@ -29,7 +30,7 @@
 .settings-dropdown-content {
   position: absolute;
   width: calc(100% + 20px);
-  top: 48px;
+  top: calc($settings__item-height + 1px);
```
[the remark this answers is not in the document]
Resaki1: because it overlaps by 1px otherwise

Resaki1:
> Not sure if this is intended, but if I go through the components by using TAB, at first all the column inputs come and then all the notes. If this is intended its fine, I was just confused and expected to go through every column, first input and then the notes of each column.
> 
> Additionally it is not possible by keyboard only to add notes. The input field is focussed, but no text input is possible (also not by pressing enter etc).
> 
> On the homepage a new element is focussed which is not visible at all. (Between the URL bar and the flags).
> 
> Maybe it's reasonable to create values for scale, brightness and some other things in the `style.scss` and reuse these values. It seems you use many different values for these functions and I don't know if this is always justified.

We already have a ticket for removing the tabIndex with a better document structure ([#1713](https://github.com/inovex/scrumlr.io/issues/1713)). I think those issues should better be resolved there

CronJorian (approved):
There is definitely more to do (e.g. atm I cannot scroll through the tabs in the settings dialog), but this is a huge improvement. LGTM 🙏

PR #2070, opened 2022-08-18
Fixed page reload after setting the language

Resaki1:
## Changelog

Removed unnecessary code that would reload the page after setting the application language.


CronJorian (approved):
I did not test this, but it looks reasonable.

PR #2071, opened 2022-08-18
#1966 Reset ready state

### LINKED ISSUE #1966: Reset ready state
As a moderator I want all users to be automatically marked as not ready once I conclude (abort or close) a voting session.

As a moderator I want to be notified by a toast if the timer is up with the message that I can reset the ready state of all users at once.

```
## Technical perspective @ Backend

* Add method to set ready to false on all board sessions for a board id
* Call method on conclusion of a voting
* Add API endpoint to change ready state for all users
* Add websocket message once ready state was changed for all users

## Technical perspective @ Frontend

* Receive websocket update and reset ready state for all users
* Show toast to moderators once timer is up with a button to trigger API call
```

BenedictHomuth:
fixes #1966


server/src/api/board_sessions.go
```diff
@@ -101,3 +101,67 @@ func (s *Server) updateBoardSessions(w http.ResponseWriter, r *http.Request) {
...
+	filter := database.BoardSessionFilterTypeFromQueryString(r.URL.Query())
+	sessions, err := s.sessions.List(r.Context(), currentBoard, filter)
+	if err != nil {
+		log.Errorw("unable to get board sessions", "err", err)
+		common.Throw(w, r, common.InternalServerError)
+	}
+
+	for _, session := range sessions {
+		board := currentBoard
+		caller := caller
```
cdreier, line 120: not sure if these assignments are necessary :)

BenedictHomuth: Done. :)

src/components/Timer/Timer.tsx
```diff
@@ -37,6 +39,9 @@ export const Timer = (props: TimerProps) => {
   const allReady = useAppSelector((state) => state.participants!.others.filter((p) => p.connected && p.role === "PARTICIPANT").every((participant) => participant.ready));
   const isModerator = useAppSelector((state) => state.participants?.self.role === "OWNER" || state.participants?.self.role === "MODERATOR");
 
+  // Better way to get boardId? Please help :)
+  const boardId = `${useAppSelector((state) => state.board.data?.id)}`;
```
BenedictHomuth, line 43: Is there a better way to get the current board id?

public/locales/en/translation.json
```diff
@@ -223,7 +223,9 @@
     "hiddenColumnsVisible": "Hidden columns are now visible to you!",
     "passwordCopied": "Board protected and password copied!",
     "boardMadePublic": "The board is now public!",
-    "allParticipantsDone": "All participants are ready."
+    "allParticipantsDone": "All participants are ready.",
+    "moderatorResetReadyStates": "Set all participants to not ready?",
```
bitionaire, line 227:
```suggestion
    "moderatorResetReadyStates": "Do you want to reset the readiness of the participants?",
```

public/locales/de/translation.json
```diff
@@ -223,7 +223,10 @@
     "hiddenColumnsVisible": "Ausgeblendete Spalten werden dir nun angezeigt!",
     "passwordCopied": "Board geschützt und Passwort kopiert!",
     "boardMadePublic": "Das Board ist nun öffentlich!",
-    "allParticipantsDone": "Alle Teilnehmer sind fertig."
+    "allParticipantsDone": "Alle Teilnehmer sind fertig.",
+    "moderatorResetReadyStates": "Alle Teilnehmer auf nicht fertig setzen?",
```
bitionaire, line 227:
```suggestion
    "moderatorResetReadyStates": "Markierung der Teilnehmer als fertig zurücksetzen?",
```

server/src/api/router.go
```diff
@@ -147,9 +148,9 @@ func (s *Server) protectedRoutes(r chi.Router) {
 			r.With(s.BoardParticipantContext).Get("/export", s.exportBoard)
 			r.With(s.BoardParticipantContext).Post("/timer", s.setTimer)
 			r.With(s.BoardParticipantContext).Delete("/timer", s.deleteTimer)
-
 			r.With(s.BoardModeratorContext).Put("/", s.updateBoard)
 			r.With(s.BoardModeratorContext).Delete("/", s.deleteBoard)
+			r.With(s.BoardModeratorContext).Put("/resetAllReadyStates", s.resetAllReadyStates)
```
bitionaire, line 153: In the REST paradigm you rather manipulate resources than executing functions (e.g. in comparison to SOAP), what means to you could probably either create an endpoint in order to set the ready state of the users by `PUT` or you'll have an endpoint to maybe `DELETE` the ready state. Not sure what would be best but I hope you get the point.

My suggestions would be that you move the endpoint to `PUT /boards/<id>/participants/ready` or you could even include the readiness within the participants. Maybe we'll come up with other use cases where we want to change some flag of all participants.

src/components/Timer/Timer.tsx
bitionaire, line 43:
```suggestion
  const boardId = useAppSelector((state) => state.board.data!.id);
```

This will do the trick. The board data is always available for the timer component.

src/components/Timer/Timer.tsx
```diff
@@ -54,6 +59,15 @@ export const Timer = (props: TimerProps) => {
     if (!previousPlayTimesUpState && playTimesUp) {
       timesUpSoundObject.on("end", () => setPlayTimesUp(false));
       playTimesUpSound();
+      if (isModerator) {
+        Toast.info(
+          <div>
+            <div>{t("Toast.moderatorResetReadyStates")}</div>
+            <Button onClick={() => API.resetAllReadyStates(boardId)}>{t("Toast.moderatorResetReadyStatesButton")}</Button>
```
bitionaire, line 66: Like mentioned in our daily you could increase the margin top of the button here, so that there's at least a gap of 1 line in between the label and the button.

bitionaire (changes requested):

src/components/Timer/Timer.tsx
BenedictHomuth: Works great! :-)

src/components/Timer/Timer.tsx
BenedictHomuth: Done. :-)

bitionaire (approved):

server/src/api/board_sessions.go
```diff
@@ -101,3 +101,42 @@ func (s *Server) updateBoardSessions(w http.ResponseWriter, r *http.Request) {
 	render.Status(r, http.StatusOK)
 	render.Respond(w, r, sessions)
 }
+
+func (s *Server) updateReadyStates(w http.ResponseWriter, r *http.Request) {
```
bitionaire, line 105: Hey @BenedictHomuth I just checked the code and saw that we already have this method implemented in `updateBoardSessions`. Why did you choose to add an additional endpoint, was there an error with the previous implementation?

server/src/api/router.go
```diff
@@ -198,6 +198,8 @@ func (s *Server) initBoardSessionResources(r chi.Router) {
 
 		r.With(s.BoardModeratorContext).Put("/", s.updateBoardSessions)
```
bitionaire, line 199: This endpoint just quite the same to the newly addition, except `/ready` only changes the ready state and this resource is also capable of e.g. changing the raised hand state for all participants. If it wasn't working before we should probably fix that :)

bitionaire (changes requested):
Just discovered that the endpoint was already implemented, can you check it?

bitionaire (changes requested):
Discussed in meet

server/src/api/board_sessions.go
migerh: There is indeed an issue with the existing endpoint that can handle more than just the ready state.

This new endpoint sends out a bunch of `PARTICIPANT_UPDATED` event from the backend to the frontend for each participant. The existing endpoint sends *one* `PARTICIPANTS_UPDATED` event for all participants of a board. Note the missing `S` in the event of the new endpoint.

The problem with the `PARTICIPANTS_UPDATED` event is, that it is triggered by the [BoardSessionsObserver.AfterScanRow](https://github.com/inovex/scrumlr.io/blob/ea560aa1d9815a02e5118839a5c6434c2845019b/server/src/database/board_sessions_observer.go#L50) hook into the `bun` DB client. This hook is fired multiple times, not just once. It is triggered by [UpdateBoardSessions](https://github.com/inovex/scrumlr.io/blob/ea560aa1d9815a02e5118839a5c6434c2845019b/server/src/database/board_sessions.go#L163). And it is missing participants in its event payloads. Let's assume we have three participants on our board, P1, P2, P3. Then the `AfterScanRow` hook will be triggered thrice by `bun`. The first trigger sends an update with only `P1` in the participants list. Update 2 will have `P1` and `P2` in its payload. Update 3 finally sends out an update with all participants in the list.

This behavior is unexpected for the frontend that expects the `PARTICIPANTS_UPDATED` event to have a list of _all_ participants. The reducer for that event will look for the current session's user and if it can't find it, delete the `participants.self` entry from the global app state. A few components rely on that info to be present, though, basically throwing those users that are missing in the event's participant list from the board.

migerh: There doesn't seem to be a way to `bun` to not trigger the hook multiple times. One solutino could be to switch the `PARTICIPANTS_UPDATED` event to a bunch of `PARTICIPANT_UPDATED` events. The number of events sent to the frontend by the backend is still the same.

bitionaire: I tried to lookup a way to fix it with `bun` but it seems that there's no way. Instead I moved to call of the update into the `board_sessions.go` and left a comment there. Also I added a simple assertion in the tests that the update will be called only once. @BenedictHomuth and @migerh what do you think?
I tested it locally and everything seems to work fine.

bitionaire (approved):

PR #2104, opened 2022-08-29
fix: show correct toast when starting a timer

### LINKED ISSUE #1954: Wrong notification when starting a timer
I think the notification is not correct. I would expect something like: "Der Timer wurde gestartet"

https://user-images.githubusercontent.com/60005702/178580283-d903da00-305c-400b-bb5d-6b19c7c3e7ec.mp4

Schwehn42:
## Description

👉 Resolves #1954

Previously, when starting a timer, a toast with the message along the lines of "All participants are ready" would appear. This is changed to "Timer started!"

## Changelog

- `Timer.tsx`: Change translation key to `Timer.timerStarted` in order to display appropiate text

## Notes to discuss:
- Change German translation from "Alarm" to "Timer"?
- Some usused translation keys, e.g. `timerEnded` and `timerCancelled` are in the locale file. Will these be used at some point or are they legacy?


CronJorian (approved):
Shame on me :(

PR #2134, opened 2022-09-06
`General`: Update toggle padding and increase consistency

Dominik-Weinzierl:
## Description

During local testing i noticed that some toggles have the wrong right padding.

## Changelog

- Enforced the correct right padding for the toggles

## Checklist

## (Optional) Visual Changes

Screenshot 2022-09-06 at 13 52 59

### DIFF
--- a/src/components/SettingsDialog/BoardSettings/BoardSettings.scss
+++ b/src/components/SettingsDialog/BoardSettings/BoardSettings.scss
@@ -258,22 +258,22 @@
 
 .board-settings__show-author-button {
   justify-content: space-between;
-  padding-right: 24px;
+  padding-right: $padding--default !important;
 }
 
 .board-settings__show-notes-button {
   justify-content: space-between;
-  padding-right: 24px;
+  padding-right: $padding--default !important;
 }
 
 .board-settings__show-columns-button {
   justify-content: space-between;
-  padding-right: 24px;
+  padding-right: $padding--default !important;
 }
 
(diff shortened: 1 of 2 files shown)


CronJorian (commented):
Using `!important` is generally discouraged and `$padding-default` translates to 16px instead of 24px. For 24px use `$padding-medium`.
The reason why the padding here does not work is due to the `settingsButton` component which has a padding of 0.

CronJorian:
Oh and it would be nice if you could provide a before and after image if possible, so that we can see the changes directly.

CronJorian:
image

image

The buttons should align with the other settings and paddings imo. These settings have a padding of 24px, so I'd rather go with that and adjust the toggle buttons

Dominik-Weinzierl:
>  The reason why the padding here does not work is due to the settingsButton component which has a padding of 0.

Yes, I am aware of this fact. But we have to be careful when removing this value, so I decided to use `!important`.

Dominik-Weinzierl:
> The buttons should align with the other settings and paddings imo. These settings have a padding of 24px, so I'd rather go with that and adjust the toggle buttons

Sure, we can change that 👍

CronJorian:
I wrapped the buttons in divs and gave them padding, so nothing will break.

CronJorian (approved):

Dominik-Weinzierl:
Thanks for your suggestions! @CronJorian

PR #2142, opened 2022-09-08
Fix: moderator board crash

### LINKED ISSUE #2100: Moderator board crashes when user deletes his own card in focus mode
When your card as a participant is being shown in the focus mode, you can still delete it. 

When you do so it crashes the board of the moderator, resulting in a white screen.

BenedictHomuth:
## Description

Closes #2100 

When a card gets deleted, the code now checks whether it is currently shared. If so, only a moderator in focus mode can delete it.

Every other participant will get an error toast saying, that it is not possible to delete a card while it is being shared. 

## Checklist


CronJorian:
Just tested it:

The problem here is not that the card is being shared, but being open as dialog (`NoteDialog`) by **any** participant.
Right now, with this fix the board of the viewer either crashes or the one who wants to delete it, is not allowed to remove it, despite the card being shared or not.

https://user-images.githubusercontent.com/44020029/189097775-be322319-072c-4172-9e43-672d68dea112.mov

public/locales/de/translation.json
```diff
@@ -315,6 +315,7 @@
     "deleteColumn": "Es tut uns leid, aber wir haben ein Problem beim Löschen der Spalte. Bitte versuche es erneut.",
     "addNote": "Es tut uns leid, aber wir haben ein Problem beim Hinzufügen des Kärtchens. Bitte versuche es erneut.",
     "deleteNote": "Es tut uns leid, aber wir haben ein Problem beim Löschen des Kärtchens. Bitte versuche es erneut.",
+    "deleteNoteWhenShared": "Kärtchen kann, während es geteilt wird, nicht von dir gelöscht werden.",
```
CronJorian, line 318:
```suggestion
    "deleteNoteWhenShared": "Das Kärtchen kann, während es geteilt wird, nicht von dir gelöscht werden.",
```

BenedictHomuth:
The error seems to be in the following lines. 

https://github.com/inovex/scrumlr.io/blob/fc2de743d205d1d8c3b9cfa5d8129ddc753454c0/src/routes/StackView/StackView.tsx#L39-L41

The expected behavior is, that if the note is undefined (e.g. because the note id is wrong / deleted) it navigates back to the board.

In fact this line gets passed through, but the navigation is ignored. (why???!)


On the other hand and in my eyes, a broken route should also be intercepted by the router in line 69... 

https://github.com/inovex/scrumlr.io/blob/fc2de743d205d1d8c3b9cfa5d8129ddc753454c0/src/routes/Router.tsx#L56-L71

but that doesn't happen either. So the user gets a whitescreen and has to reload the page.

CronJorian (approved):

BenedictHomuth:
This PR is from now on just focusing on checking whether a moderator is sharing the currently active note. If so the creator can not delete it, except when the moderator is also the creator.

The described whitescreen issue(#2100) is fixed in (#2161)

PR #2160, opened 2022-09-13
Implement button to return to focused note

brandstetterm:


public/locales/de/translation.json
```diff
@@ -1,4 +1,7 @@
 {
+  "InfoBar": {
+    "ReturnToFocusedNote": "Zurück zur Fokuskarte"
```
CronJorian, line 3: Too much hokus ~~pokus~~ fokus :P 
```suggestion
    "ReturnToFocusedNote": "Zurück zur fokussierten Karte"
```
Or do you think it's too long? I think the word "Fokuskarte" is not selfexplanatory for the most users. If you agree and have a better translation, go for it. If not, then it's fine too.

CronJorian (approved):
lgtm 👏

public/locales/de/translation.json
brandstetterm: It's too long, that's why I've used 'Fokuskarte'

PR #2167, opened 2022-09-14
Fix manual approvals

brandstetterm:


CronJorian (changes requested):
This solves the problem with two users being either both rejected or approved.
What it does not solve is the user being redirected to the board.

If I have 2 users requesting at the same time and approve one of them and reject the other, both of them get the "You have been rejected message." and only after reloading the page, one of them has access to the board.

Edit: This still has some issues, which might be frontend related. If you want I can approve this MR, but there are definitely still some problems

CronJorian:
@brandstetterm do you want to fix this (see below) in this PR or later on?

https://user-images.githubusercontent.com/44020029/190585393-608b3791-7cb7-4e80-b1b1-d07ab04243a4.mov

CronJorian (approved):

PR #2189, opened 2022-09-22
Fix #2089: Timer offset on different clients

### LINKED ISSUE #2089: Analyse timer offset on different clients
Some users reported that the timer can have a larger offset (+ 30 sec.) and we should look into this issue and fix it.

BenedictHomuth:
## Description
Timer is now working independently from timezones. This is done via calculating an offset.

## Changelog
Changed how the offset is calculated between client and server.

## Checklist


BenedictHomuth:
There is still a bug / unintended behaviour.

Whenever a updatedBoard event is fired, the "timerEnd" value of our timer gets send to the database.... from whoever is currently triggering the event.

This is problematic, since we overwrite the initial "timerEnd" value when calculating the offset
- so a 04:05 PM alarm is set to 06:05 PM  for a user who is two hours ahead


As a consequence we face the following issue:
> (Serverside) timerEnd = 04:05 
>>  (Clientside / Different timezone (+2 hours)): timerEnd = 06:05 
>>>  Client triggers board update
>>>> timerEnd gets overwritten for everybody to 06:05
>>>>> all frontends get notified via websocket message
>>>>>> when another event is triggered the offset continues to 08:05...

CronJorian (approved):
LGTM, but I did not test this.

PR #2271, opened 2022-10-18
redesign request notification dialog

### LINKED ISSUE #2085: Implement redesign of raised hand notification dialog
To further clarify some design details that aren't self explanatory by the exported design:

- Width of the notifications should be defined by the length of the usernames

### LINKED ISSUE #2192: Implement redesign of join request dialog

Schwehn42:
## Description

**This PR implements new designs for both Join -and RaiseHand request notifications.**
- resolves #2085 
- resolves #2192 

## Changelog

- refactored the Request component by splitting it into two files, `Requests.tsx` and `Request.tsx`
- former is a container for all requests, latter is the request itself.
- implemented design and transistions according to design

## Checklist

## (Optional) Visual Changes

[//]: <> (If available, please provide a before and after screenshot of your UI related changes.)

https://user-images.githubusercontent.com/7889564/196410330-8a7995ac-7100-493f-ae2f-2708dd7adb56.mov


src/components/Requests/Request/Request.scss
```diff
@@ -0,0 +1,117 @@
...
+  animation: fly-in-right $fly-in-time ease-out;
+
+  background-color: $color-white;
+  box-shadow: 0 3px 5px rgba($menu-icon-background-color--dark, 30%);
+
+  transition: margin-right $transition-time ease-out, padding-right $transition-time ease-out;
+
+  &:hover {
+    padding-right: 10px;
+    margin-right: 10px;
```
CronJorian, line 29: Why padding and margin?
Wouldn't margin be enough? And could you use $padding-default or $padding-small?

src/components/Requests/Request/Request.scss
```diff
@@ -0,0 +1,117 @@
...
+  left: 50%;
+  transform: translateX(-50%);
+}
+
+.request__button {
+  height: 40px;
+  width: 100%;
+
+  border: none;
+  border-radius: 21px;
```
CronJorian, line 88:
```suggestion
  border-radius: $border-radius--round;
```

src/components/Requests/Request/Request.tsx
```diff
@@ -0,0 +1,58 @@
+import "./Request.scss";
+import {ReactComponent as HandIcon} from "assets/icon-hand.svg";
+import {useTranslation} from "react-i18next";
+import {UserAvatar} from "../../BoardUsers";
+import {Auth} from "../../../types/auth";
+
+type RequestType = "JOIN" | "RAISE_HAND";
+
+export type RequestProps = {
```
CronJorian, line 9: no need for exporting, when only used in `Request.tsx`

```suggestion
type RequestProps = {
```

src/components/Requests/Request/Request.tsx
```diff
@@ -0,0 +1,58 @@
...
+    );
+  };
+
+  return (
+    <div className="request__container">
+      <figure className="request__request-figure">
+        <UserAvatar id={participant.id} avatar={participant.avatar} name={participant.name} />
+      </figure>
+
+      {/* either `request__text-container` or `request__button` is displayed */}
```
CronJorian, line 44: Could not find `request__text-container`. Is this comment still relevant?

src/components/Requests/Request/Request.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <div className="request__middle-wrapper">
```
CronJorian, line 45: imho there is no need to call it "middle-wrapper" instead of "wrapper".

src/components/Requests/Requests.tsx
```diff
@@ -0,0 +1,43 @@
...
+    } else {
+      dispatch(Actions.rejectJoinRequests([user]));
+    }
+  };
+
+  const lowerHand = (user: string) => {
+    dispatch(Actions.setRaisedHand(user, false));
+  };
+
+  return requests.length !== 0 || participantsWithRaisedHand.length !== 0 ? (
```
CronJorian, line 28: should also work as: `return requests.length || participantsWithRaisedHand.length ? (` because 0 is a falsy value

CronJorian (commented):
All in all well done, just some small nitpicks. :)

src/components/Requests/Request/Request.scss
CronJorian: Not sure whether I get what you mean. The :hover pseudo class had only the two right spacings and I just wondered why you need to push the outer _and_ inner box instead of just the outer (`margin-right: 20px;`) or the inner (`padding-right: 20px;`). That was why I asked :)
The effect per se is nice :)

src/components/Requests/Requests.scss
```diff
@@ -0,0 +1,23 @@
+@import "src/constants/style";
+
+$wrapper-position__top: 110px;
+$wrapper-position__right: 30px;
+$gap-between-requests: 15px;
+
+/* fixed wrapper in the top right containing all request */
+.requests__wrapper {
```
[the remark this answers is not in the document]
CronJorian: Would be an option, you can check if theres an option to add an shadow if the list is scrollable (e.g. https://css-scroll-shadows.vercel.app/ resp. https://lea.verou.me/2012/04/background-attachment-local/) so the scrollable is clear.

Though more easily is including the scrollbar mixin (`@include scrollbar();`) (and maybe changing the color) The scrollbar is only shown if theres scrollable content :)

CronJorian (changes requested):
Sorry for the late review, there are still some things I noticed:
- The request window is blocking the elements behind:

https://user-images.githubusercontent.com/44020029/201945655-18e53332-d33d-48e1-9398-34c388af2596.mov

- Not sure if the div should always be the full height, but I guess its fine as long it doesn't block elements below from being accessed
  - you can use `pointer-events: none;` on `requests__wrapper` and `pointer-events: all;` on `requests__main` for a quickfix, but i am not sure if this is a bad practice or not.

- The request is above the settings portal which leads to ugly moments

image

- The box shadow to the left and right is cut off due to the width of the parent element

image

One time after processing a join request the backend crashed, but I am not sure whether it's because of any of these changes, just wanted to let you know.
These were the issues I found, I think after addressing these, we'll just merge it and change other stuff later on (if there is more to change)

Oh and I think while having 10-20 requests at a time the browser slows down while hovering. This might be linked to the cut of sides of the box shadow. Would be nice if you could test this too (just open multiple private windows in safari) and check the performance.

src/components/Requests/Requests.scss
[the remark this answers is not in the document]
Schwehn42: i've made some further adjustments, the requests container is now no longer full height. 

Also, I'm not certain if you would like the scrollbar to be shown when hovering (and the list is scrollable) or don't show it at all times?

Schwehn42:
@CronJorian shadows should be fixed now.  
Not sure what caused the backend crash, I'm just gonna claim it didn't happen because of any of my changes 👀

src/constants/style.scss
```diff
@@ -14,6 +14,7 @@ $color-dark-gray: #859093;
 $color-darker-gray: #4f5253;
 $color-darkest-gray: #232323;
 $color-black: #111111;
+// $transparent: rgba(0, 0, 0, 0);
```
CronJorian, line 17: Is this comment necessary? :)

CronJorian:
Looks really good now :)

Is it intended that the request container has just a light theme? If so, i'll approve, if not can you add these changes and then we're done here 😅

src/constants/style.scss
scja1069: nope ^^

Schwehn42 (commented):
@CronJorian I think we should be good to go now.  
At most one could add a second icon for the notification type "Join", but I couldn't find a fitting one in our exisiting assets, maybe we can add that later?

CronJorian (approved):

PR #2324, opened 2022-11-04
Add hotkey combination as tooltip to action buttons

NNKKOO:
Added tooltips, displaying the corresponding hotkey, when hovering over an action button.


NNKKOO:
Please tell me if you think the show-delay is okay. I'm afraid I got a little impatient during the testing process and might have set it too short.

src/components/Board/Board.tsx
```diff
@@ -147,7 +148,7 @@ export const BoardComponent = ({children, currentUserIsModerator, moderating}: B
   return (
     <>
       <style>{`.board { --board__columns: ${columnsCount} }`}</style>
-
+      <ReactTooltip type={localStorage.getItem("theme") === "dark" ? "dark" : "light"} place="bottom" delayShow={500} multiline />
```
Resaki1, line 151: localStorage doesn't properly reflect the current theme if it is set to "auto", which leads to the tooltips being white when theme is set to auto and the OS setting is set to dark. Maybe you can access the "theme" property on the <html> element?

src/components/Column/Column.tsx
```diff
@@ -166,7 +168,14 @@ export const Column = ({id, name, color, visible, index}: ColumnProps) => {
...
       <div className="column__content">
         <div className="column__header">
-          <NoteInput columnIndex={index} columnId={id} maxNoteLength={MAX_NOTE_LENGTH} columnIsVisible={visible} toggleColumnVisibility={toggleVisibilityHandler} />
+          <NoteInput
+            columnIndex={index}
+            columnId={id}
+            maxNoteLength={MAX_NOTE_LENGTH}
+            columnIsVisible={visible}
+            toggleColumnVisibility={toggleVisibilityHandler}
+            hotkeyHint={`${SELECT_NOTE_INPUT_FIRST_KEY.map((key, i) => (i === 0 ? `${key}/` : key)).join("")} + ${index + 1}`}
```
Resaki1, line 177: maybe the keys toUpperKeys as well, just like the others?

src/components/MenuBars/MenuBars.tsx
```diff
@@ -108,10 +109,11 @@ export const MenuBars = ({showPreviousColumn, showNextColumn, onPreviousColumn,
                 icon={RaiseHand}
                 onClick={toggleRaiseHand}
                 active={raisedHand}
+                hotkeyHint={TOGGLE_RAISED_HAND.toUpperCase()}
               />
             </li>
             <li>
-              <TooltipButton direction="right" label={t("MenuBars.settings")} onClick={showSettings} icon={SettingsIcon} />
+              <TooltipButton direction="right" label={t("MenuBars.settings")} onClick={showSettings} icon={SettingsIcon} hotkeyHint={SHOW_SETTINGS.toUpperCase()} />
```
Resaki1, line 116: imo it looks really weird that the TooltipButtons now have two tooltips. Maybe we can add the hotkey hint right after the label (in a less visible font-color). What do you think?

src/components/MenuBars/MenuBars.tsx
```diff
@@ -129,10 +131,21 @@ export const MenuBars = ({showPreviousColumn, showNextColumn, onPreviousColumn,
           {isAdmin && (
             <ul className="menu__items">
               <li>
-                <TooltipButton direction="left" label="Timer" onClick={showTimerMenu} icon={TimerIcon} />
+                <TooltipButton
+                  direction="left"
+                  label="Timer"
+                  onClick={showTimerMenu}
+                  icon={TimerIcon}
+                  hotkeyHint={
```
Resaki1, line 139: this one even overlaps the buttons tooltip

Resaki1 (changes requested):

src/components/MenuBars/MenuBars.tsx
NNKKOO: It should look like this. But I'll look into that
Bildschirmfoto 2022-11-04 um 17 29 18

src/components/MenuBars/MenuBars.tsx
NNKKOO: Good idea. This would also make some of the other requests obsolete.

Resaki1 (approved):

PR #2325, opened 2022-11-07
"Create board" button in fixed position at bottom

NNKKOO:


src/routes/NewBoard/NewBoard.scss
```diff
@@ -112,7 +113,13 @@
 
 .new-board__actions {
   display: flex;
-  margin-top: 2em;
+  position: -webkit-sticky;
```
Schwehn42, line 116: LGTM, only thing I'm wondering isn't the `-webkit-sticky` unnecessary?

Schwehn42 (approved):
I tested it on different devices/browsers and both themes, seems to be working fine. code looks good too afaict 👍

Issue #2356, opened 2022-11-11
Tooltip button shows undefined if no hotkey is provided


Resaki1:
![Image](https://user-images.githubusercontent.com/35272402/201307447-9b5c5f41-ef7b-4911-90f3-96721082eac1.png)

PR #2358, opened 2022-11-11
TooltipButton hotkey bug fix

NNKKOO:
tooltip buttons no longer show an 'undefined' hotkey

### DIFF
--- a/src/components/TooltipButton/TooltipButton.tsx
+++ b/src/components/TooltipButton/TooltipButton.tsx
@@ -26,7 +26,7 @@ export const TooltipButton = (props: TooltipButtonProps) => {
       <div className="tooltip-button__tooltip" aria-hidden>
         <span className="tooltip-button__tooltip-text">
           {props.label}
-          <span className="tooltip-button__hotkey">{` [${props.hotkeyKey}]`}</span>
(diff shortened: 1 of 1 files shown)


src/components/TooltipButton/TooltipButton.tsx
```diff
@@ -26,7 +26,7 @@ export const TooltipButton = (props: TooltipButtonProps) => {
       <div className="tooltip-button__tooltip" aria-hidden>
         <span className="tooltip-button__tooltip-text">
           {props.label}
-          <span className="tooltip-button__hotkey">{` [${props.hotkeyKey}]`}</span>
+          {props.hotkeyKey !== undefined && <span className="tooltip-button__hotkey">{` [${props.hotkeyKey}]`}</span>}
```
Resaki1, line 29: Just fyi, this could be written shorter:
```suggestion
          {props.hotkeyKey && <span className="tooltip-button__hotkey">{` [${props.hotkeyKey}]`}</span>}
```

But your way is more explicit and thus better to read 👍🏼

Resaki1 (approved):

Resaki1:
closes #2356

PR #2359, opened 2022-11-11
"back to shared note" - button is no longer displayed to the moderator who activated the focus mode

NNKKOO:


Resaki1 (changes requested):
Works like a charm 👍 

Since we basically only store the initiator, which is a participant, in the store, could we move the code to the participants part of the store (as in, `action/participants.ts`, `reducer/participants.ts` and so on)? I think this would make thinks easier to understand in the future

CronJorian:
Sorry to bring the bad news but this is not working :(
You are right that this works when clicking the focus-mode button, but using the hotkey still breaks it.

https://user-images.githubusercontent.com/44020029/201960053-d67fa682-d4b3-4c3f-899b-587375f4fab4.mov

src/store/reducer/participants.ts
```diff
@@ -55,6 +60,26 @@ export const participantsReducer = (state: ParticipantsState = null, action: Red
           user: action.user,
         },
         others: [...state!.others],
+        focusInitiator: state!.focusInitiator,
```
CronJorian, line 63: If I get this right, this is not needed. `...state` returns the whole state and the elements afterwards are the modified ones, so if the EditSelf action has something to do with the focusInitiator, this would be right, but I dont think it does. You are welcome to correct me if i'm wrong :)

CronJorian: I'm not sure about the rest of the `focusInitiator` additions some lines above, I think it's worth checking them too

CronJorian (changes requested):

CronJorian:
The problem shown in the video should be rechecked after #2360 is merged. The problem might disappear afterwards.

CronJorian:
BTW, we should also check what happens when the presenter reloads the site. At the moment (main branch) from what I'm seeing, it doesn't stop the presentation, but looses the presenter reference aka. someone is still presenting, but not the presenter from before

NNKKOO:
I checked it and can confirm. The original presenter looses his presenting status. It can easily be fixed by just starting the focus mode again, but yeah, this really shouldn't happen.

NNKKOO:
But shouldn't this be a new issue?

CronJorian:
Yeah sure, just wanted to write it down somewhere :) I'll create an issue 👍

NNKKOO:
should i wait with the merge request then?

CronJorian:
What do you mean? Wait until #2393 is solved? Nah :D Although the issue is related, they don't depend on each other, so there is no conflict 👍

Resaki1 (approved):

CronJorian (approved):

CronJorian:
Didn't check this, but if @Resaki1 approves, it should be fine 👍 sorry it took so long.

PR #2360, opened 2022-11-13
Fix focus button bugs

### LINKED ISSUE #2342: Toggling focus by hotkey while detail view of note is opened
As a moderator I expect that the focus mode will be triggered once I press the focus mode hotkey and I have opened the detail view of note.
Currently the opened note won't be shared to other participants in this case.

### LINKED ISSUE #2345: Shared note button does not disappear if focus mode will be finished by hotkey
The button to open a focused note does not disappear if the moderator will end the focus mode by hotkey while the detail view of a note is opened.

Resaki1:
## Description
If the focus mode was ended using the hotkey, the return-to-focus-button would still be visible for users. Vice versa, if the focus mode was startet using the hotkey while already being in the stack view, the current card was not shared with the other users. Both bugs have been fix. Closes #2342 and #2345


CronJorian (approved):
Not bad, thought this would be way more complex.
Didn't find any issues, so looks pretty f good

PR #2443, opened 2022-12-05
Change presenter mode hotkey

brandstetterm:
Since we've changed the naming from `focus mode` to `presenter mode`, changing the hotkey from `F` to `P` also makes sense. The hotkey PDF also has been updated.


Lennart01 (approved):
Looks good :shipit:

PR #2468, opened 2022-12-07
Added new menu bar icons

### LINKED ISSUE #2304: Adjust icons
Anbei die aktualisierten Icons in gleicher Linienstärke für dark- und lightmodus jeweils als png & svg.

[Scruml_icons_update_27102022.zip](https://github.com/inovex/scrumlr.io/files/9879830/Scruml_icons_update_27102022.zip)

bitionaire:
Closes #2304 

**Known issue, please check it locally**
The icons on the admin bar flicker when I quickly hover between the different buttons. Dunno whether it's related to my browser (Chrome 108.0.5359.96, Windows 11, 1920x1080) and I couldn't debug it. I guess it may be related to the point decimal precision within the svg 🤷‍♂️ 
EDIT: Seems to be related to precision and aliasing of SVGs in Chrome. It's fixed once I apply `shape-rendering: crispedges` but then the icons look even worse 🤮 

**Learnings**
- Icons can't be separated in light/dark mode since they often are styled by CSS e.g. on hover and change their color
- Icons can't be imported dynamically as inline-html (... there are hacky workarounds, but they suck)
- Regardsless of which minifier you'll find there will be still many things left you can optimize manually (e.g. wrap `stroke="#fff"` in a wrapping `<g />` element or stuff like that 

I will close #2457 therefore, even though there are some improvements for the `<Html />` component but I guess we can work on that a later stage.


brandstetterm:
The icons are now getting displayed slightly larger than before. Is that intended? The timer icon looks decentered now 

Before:
Screenshot 2022-12-08 at 06 09 08

Now: 
Screenshot 2022-12-08 at 06 09 33

bitionaire:
> The icons are now getting displayed slightly larger than before. Is that intended? The timer icon looks decentered now
> 
> Before: Screenshot 2022-12-08 at 06 09 08
> 
> Now: Screenshot 2022-12-08 at 06 09 33

I'll ask our experts :)

CronJorian:
Additionally to @brandstetterm s comment:
-  I think the `icon-check` isn't centered either and should be 0.5px to the right.
image
- depending on what should be centered is the `icon-focus` 1.x px too far to the left (if the rectangle should be in the center), or ~0.75 px too far to the right
- the `icon-timer` is definitely off too.

The remaining icons should also be checked probably.

PR #2475, opened 2022-12-08
improvement: increase distinct look of hidden columns

brandstetterm:
## Description

After it is repeatedly noted that it is not quite clear whether a column is hidden or not, I've tried to make some improvements.

## Changelog
- Add crosshatched background to hidden columns
- Change the color from the hidden icon from grey to red

## (Optional) Visual Changes

Screen Shot 2022-12-08 at 10 22 32
Screen Shot 2022-12-08 at 10 22 25
Screen Shot 2022-12-08 at 10 21 57
Screen Shot 2022-12-08 at 10 21 41

### DIFF
--- a/src/components/Column/Column.scss
+++ b/src/components/Column/Column.scss
@@ -27,6 +27,14 @@
     background-color: $color-dark-two;
   }
 }
+.column--hidden {
+  background-image: repeating-linear-gradient(45deg, $color-white-one, $color-white-one 20px, $color-white-two 20px, $color-white-two 40px);
+}
+[theme="dark"] {
+  .column--hidden {
+    background-image: repeating-linear-gradient(45deg, $color-dark-one, $color-dark-one 20px, $color-dark-two 20px, $color-dark-two 40px);
+  }
+}
 
 .column__content {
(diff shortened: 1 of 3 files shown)


src/components/Column/Column.scss
```diff
@@ -172,7 +180,7 @@
   min-width: $icon--medium;
   min-height: $icon--medium;
   margin: 0 $margin--small 0 0;
-  color: $color-middle-gray;
+  color: $color-warning-red;
```
CronJorian, line 183: image

What do you think about a border around the icon?

CronJorian:
```suggestion
  color: $color-warning-red;
  border: 2px solid $color-warning-red;
  box-sizing: content-box;
  border-radius: $border-radius--round;
```

CronJorian: Oh and in case you want to use this ^, you should probably add the border-color to the hover effect, too :)

brandstetterm: Ah, I think it look's better and more minimalistic without the border 😄

PR #2477, opened 2022-12-08
feat: add assignees to notes

NNKKOO:


server/src/database/notes.go
```diff
@@ -128,6 +137,15 @@ func (d *Database) updateNoteText(update NoteUpdate) (Note, error) {
 	return note, nil
 }
 
+func (d *Database) updateNoteAssignee(update NoteUpdate) (Note, error) {
+	var note Note
+	_, err := d.db.NewUpdate().Model(&update).Column("assignee").Where("id = ?", update.ID).Where("board = ?", update.Board).Where("id = ?", update.ID).Returning("*").Exec(common.ContextWithValues(context.Background(), "Database", d, "Board", update.Board), &note)
```
NNKKOO, line 142: wouldn't it make sense to use Scan() instead of Exec()? since Scan() does the same as Exec() but omits the sql result .

NNKKOO: nvm Scan only works for Selects..

NNKKOO:
TODO:
- [ ] provide translation
- [ ] make list appear below corresponding note
- [ ] prevent the note from opening after assigning someone
- [ ] display name as tooltip on hover
- [ ] give external assignees different background colors
- [ ] better add assignee button
- [ ] make it possible to assign by hitting enter
- [ ] ...

PR #2488, opened 2022-12-10
Small bug fixes

Resaki1:
## Description

Fixed some small bugs like missing keys in lists and improved a few stylings.

## Changelog
- Added missing keys to list items
- Fixed uncontrolled/controlled input bug
- fixed Board Settings not updating automatically when component is open and other users change values
- fixed German word in board settings not hyphenating automatically in Safari
- Simplified German translations
- fixed disabled toggle to look the same as enabled and set to off toggle

### before
Bildschirmfoto 2022-12-10 um 15 22 59
Bildschirmfoto 2022-12-10 um 15 24 00

### after
Bildschirmfoto 2022-12-10 um 15 24 40
Bildschirmfoto 2022-12-10 um 15 24 00


src/components/SettingsDialog/Components/AvatarSettings.tsx
```diff
@@ -97,12 +97,11 @@ export const AvatarSettings: FC<AvatarSettingsProps> = ({id}) => {
       <div className="avatar-settings__settings-wrapper">
         <div className="avatar-settings__settings">
           {Object.entries(settingGroups).map(([label, props], groupIndex, array) => (
-            <>
+            <div key={label}>
```
CronJorian, line 100: You can also use `<Fragment key={label}>` instead of the tag. This will omit the visible `div` tag and the error will not appear.
https://reactjs.org/docs/fragments.html#keyed-fragments
Additionally it might be possible to create an HR element with CSS only and place it as ::before/::after element of :nth-child() **_OR_** maybe it's possible to use the CSS `gap` attribute and style it to let it look like an HR.

TLDR:
Solution is fine, but `<Fragment key={label}>` might be better, and I think the hardcoded `<hr>` in line 135/133 should be replaced by pure CSS. Then neither (Fragment/div) is required.

src/components/SettingsDialog/Components/AvatarSettings.tsx
```diff
@@ -115,25 +114,24 @@ export const AvatarSettings: FC<AvatarSettingsProps> = ({id}) => {
                         .some((val) => val);
 
                     return (
-                      <>
+                      <div key={element.key}>
```
CronJorian, line 117: See comment above

src/components/SettingsDialog/Participants/Participants.tsx
```diff
@@ -38,7 +38,7 @@ export const Participants = () => {
             {state.others.length > 0 && <hr className="settings-dialog__separator" />}
             {state.others.length > 0 &&
               state.others.map((participant, index) => (
-                <>
+                <div key={participant.user.id}>
```
CronJorian, line 41: See comment above about Fragments

src/components/Toggle/Toggle.scss
```diff
@@ -42,6 +42,8 @@
 }
 
 .toggle.toggle--disabled {
+  cursor: not-allowed;
+  opacity: 0.5;
```
CronJorian, line 46: Maybe add a newline after the line, but I dont care

CronJorian (changes requested):

CronJorian (approved):

PR #2533, opened 2022-12-23
Gifs and Images on Notes

Resaki1:
## Description

Notes that have an image link (including gifs) now automatically display that image. A little indicator shows when a link will be converted.

## Changelog

- added check if note text is an image link, if so display that image on note
- added indicator showing when a link will be converted
- fixed missing english translation for board access policy on board header

## Checklist

## Visual Changes

Bildschirmfoto 2022-12-23 um 15 37 02


src/components/Note/Note.tsx
```diff
@@ -105,14 +111,25 @@ export const Note = (props: NoteProps) => {
   const stackSetting: "stackOntop" | "stackBetween" | "stackBelow" = "stackBetween";
 
   return (
-    <div className={classNames("note__root")}>
+    <div className={classNames("note__root", getColorClassName(column?.color))}>
```
brandstetterm, line 114: You don't need to add `getColorClassName(column?.color)` here because the accent colors are already provided by the parent / column component. Since the column already has the specific class name for it's color, all the css properties belonging to this classname can also be used by all of the children of a column.

src/components/Note/Note.tsx
```diff
@@ -12,6 +12,9 @@ import {Actions} from "store/action";
 import {Participant} from "types/participant";
 import "./Note.scss";
 import {getEmptyImage} from "react-dnd-html5-backend";
+import {addProtocol} from "utils/images";
+import {getColorClassName} from "constants/colors";
```
brandstetterm, line 16: Can be removed because the columns accent color is already provided by the column component. Since notes are children of columns, you can use the accent colors introduced by the column.

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
```diff
@@ -26,21 +28,32 @@ export const NoteDialogNoteContent: FC<NoteDialogNoteContentProps> = ({noteId, a
...
-            e.preventDefault();
-            (e.target as HTMLTextAreaElement).blur();
-          }
-        }}
-        defaultValue={text}
-      />
+      {isImage ? (
+        <img
+          src={addProtocol(text)}
+          className="note-dialog__note-content--image"
```
brandstetterm, line 38: Why doesn't the image use the full width and height here? 
Screenshot 2023-01-09 at 11 09 04

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
```diff
@@ -26,21 +28,32 @@ export const NoteDialogNoteContent: FC<NoteDialogNoteContentProps> = ({noteId, a
...
+          alt=""
```
brandstetterm, line 39: You should add an ALT text that for example generally describes that the note displays an image

brandstetterm (dismissed):

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
Resaki1: On the dashboard, the images are fully stretched over the note because it looks better. This however usually leads to some parts of the image being cut off. In order to see the full image, they aren't stretched in stack view.
This was discussed with the team in slack. Would you prefer it if they were filling the whole note in stack view as well?

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
```diff
@@ -36,7 +36,7 @@ export const NoteDialogNoteContent: FC<NoteDialogNoteContentProps> = ({noteId, a
         <img
           src={addProtocol(text)}
           className="note-dialog__note-content--image"
-          alt=""
+          alt="user-provided media"
```
brandstetterm, line 39: I think the alt text should be translated too

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
Schwehn42: I think a larger view would look better on a lot of media, especially ones with text or images that are stretched horizontally. In focus mode, it's kinda difficult to make out what's displayed.

A: In normal view mode, this image is cut off. In focus mode, you can't read the text because it's too small.
grafik
B: This cat would look a lot cuter if it was displayed in a larger size.
grafik

In any case, the team already voted for this option, just wanted to add my two cents :)

Schwehn42:
This seems to work quite well for the image extensions as defined in the code.  
During testing with various media sources, I noticed that many sites (for instance Reddit) like to pretend the file is a GIF, when it's actually a mp4.  
This leads to the following url being displayed just as text: https://preview.redd.it/fl4puwzxt9ha1.gif?format=mp4&v=enabled&s=8a0346424b448ba342c09dcc5629e04d8e0171c2
grafik

Removing the query params in the url (https://preview.redd.it/fl4puwzxt9ha1.gif) results in this:
grafik

Now I tried to dig into into it a little and stumbled upon this article: https://cloudinary.com/blog/evolution_of_img_gif_without_the_gif

According to above article, you might be able to use something like this
```html
<video autoplay loop muted inline>
  <source src="eye-of-the-tiger-video.webm" type="video/webm">
  <source src="eye-of-the-tiger-video.mp4" type="video/mp4">
  <img src="eye-of-the-tiger-fallback.gif" />
</video>
```
in order to display videos with the option to use a fallback gif. Or maybe use `<video>` and `` for the respective file extension? In any case I think it would be cool if these type of media could be displayed.

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
Resaki1: I kept it the way it was but added a simple functionality to zoom in on the pictures in stack view. Let me know what you think about it.

Resaki1:
> This seems to work quite well for the image extensions as defined in the code. During testing with various media sources, I noticed that many sites (for instance Reddit) like to pretend the file is a GIF, when it's actually a mp4. This leads to the following url being displayed just as text: https://preview.redd.it/fl4puwzxt9ha1.gif?format=mp4&v=enabled&s=8a0346424b448ba342c09dcc5629e04d8e0171c2 grafik
> 
> Removing the query params in the url (https://preview.redd.it/fl4puwzxt9ha1.gif) results in this: grafik
> 
> Now I tried to dig into into it a little and stumbled upon this article: https://cloudinary.com/blog/evolution_of_img_gif_without_the_gif
> 
> According to above article, you might be able to use something like this
> 
> ```
> <video autoplay loop muted inline>
>   <source src="eye-of-the-tiger-video.webm" type="video/webm">
>   <source src="eye-of-the-tiger-video.mp4" type="video/mp4">
>   
> </video>
> ```
> 
> in order to display videos with the option to use a fallback gif. Or maybe use `<video>` and `` for the respective file extension? In any case I think it would be cool if these type of media could be displayed.

Thanks for the feedback! I couldn't get this to work for various reasons as of now. IMO we should add a new issue and try to solve this some time in the future, what do you think?

Schwehn42:
> IMO we should add a new issue and try to solve this some time in the future, what do you think?

@Resaki1 alright, sounds reasonable :)

PS: why can't I reply as a thread, on GitLab there's an option for it but on GitHub it needs to be a review comment I guess? 😅

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
Schwehn42: I like the idea of being able to zoom into the picture by clicking on it, but I would implement it differently.  
Zooming into the center doesn't necessarily help to see the picture more clearly.
https://user-images.githubusercontent.com/7889564/219742741-d0009878-86ec-49d8-b778-e202293d3607.mov
In above example, zooming actually makes us see less content that's required to understand the meme (or whatever picture I just picked)

My suggestion would be to use imgurs approach, that is showing the media in **full size** and not in the container itself, but above everything else. Clicking anywhere on the screen results in it disappearing, or zooming out so to speak. That way, you would be able to see every detail of the picture, optionally.

Example:
https://imgur.com/gallery/TS0IEpP
Bildschirm­foto 2023-02-17 um 19 11 48

As you can see, the bird takes full size of the screen while we want to look at it. Wdyt?

Resaki1: Thanks for pointing that out. That's kinda the way I intended it to be, but it didn't work in Firefox which I didn't test. I fixed that and am now using a React portal to achieve the same effect as imgur. This is what it looks like:
https://user-images.githubusercontent.com/35272402/219949628-1d56f638-2b5c-4b0b-aa9f-ac7768dc87ce.mov
Is that good enough?

Resaki1:
> > IMO we should add a new issue and try to solve this some time in the future, what do you think?
> 
> @Resaki1 alright, sounds reasonable :)
> 
> PS: why can't I reply as a thread, on GitLab there's an option for it but on GitHub it needs to be a review comment I guess? 😅

Yeah, you can only reply as a thread to review comments, not normal ones (which is pretty dumb imo 😅)

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
Schwehn42: Yes, I like it a lot 👍

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.scss
```diff
@@ -24,17 +36,63 @@
...
+.note-dialog__note-content--image-zoom-backdrop {
+  height: 100vh;
+  width: 100vw;
+  position: fixed;
+  top: 0;
+  left: 0;
+  cursor: zoom-out;
+  background-color: rgba(0, 0, 0, 0.8);
+  backdrop-filter: blur(10px) brightness(0.76) saturate(0);
+  background: rgba(var(--accent-color-rgb), 0.42);
```
Schwehn42, line 80: `backgrund-color` seems to be overwritten by `background` and could be removed

src/components/Note/Note.scss
```diff
@@ -170,12 +165,13 @@ $note__author-image-size: 26px;
   flex-grow: 0;
   height: 28px;
   padding: 0 $margin--default 0 0;
+  background: $color-white;
+  border-radius: 21px;
+  padding-left: 2px;
```
Schwehn42, line 170: `padding` and `padding-left` can be merged here

Schwehn42 (commented):
i commented one more small thing, other than that it looks really good 👍

src/components/Note/Note.scss
Resaki1: fixed it, thanks

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
```diff
@@ -12,9 +19,22 @@ type NoteDialogNoteContentProps = {
 };
 
 export const NoteDialogNoteContent: FC<NoteDialogNoteContentProps> = ({noteId, authorId, text, viewer}: NoteDialogNoteContentProps) => {
```
Schwehn42, line 21: random comment so I can start a review and approve 👀

Schwehn42 (approved):

PR #2545, opened 2022-12-29
improvement: Improve SettingsDialog

brandstetterm:
## Changelog

- Change black background to colored background (backlog-blue in light-mode & planning-pink in dark-mode)
- Add transition to SettingsDialog
- Use style constants introduced in `mb/leave-board-dialog` branch

### Before
Screenshot 2022-12-29 at 13 45 15
Screenshot 2022-12-29 at 13 45 30

### After

Screenshot 2022-12-29 at 13 44 53

Screenshot 2022-12-29 at 13 45 48


src/constants/style.scss
```diff
@@ -89,14 +89,33 @@ $action-bar__width: 64px;
...
+}
 
 // border-radius
 $note__border-radius: 16px;
-$border-radius--round: 50px;
+$rounded--small: 8px;
+$rounded--default: 16px;
+$rounded--medium: 24px;
+$rounded--large: 32px;
+$rounded--full: 9999px;
```
Resaki1, line 118:
```suggestion
$rounded--full: 100%;
```

Resaki1 (changes requested):
personally, I'd speed up the animation a little so the dialog can be interacted with sooner, but that's just my preference.
Also, on mobile the slide in looks kinda weird imo

Resaki1 (approved):

src/components/SettingsDialog/SettingsDialog.tsx
```diff
@@ -1,4 +1,5 @@
-import {useEffect, VFC} from "react";
+import React, {useEffect} from "react";
```
Resaki1, line 1:
```suggestion
import {useEffect, FC} from "react";
```

src/components/SettingsDialog/SettingsDialog.tsx
```diff
@@ -18,14 +19,20 @@ import {ReactComponent as ExportIcon} from "assets/icon-export.svg";
 import {ReactComponent as FeedbackIcon} from "assets/icon-feedback.svg";
 import "./SettingsDialog.scss";
 
-export const SettingsDialog: VFC = () => {
+export const SettingsDialog: React.FC = () => {
```
Resaki1, line 22:
```suggestion
export const SettingsDialog: FC = () => {
```

Resaki1 (changes requested):

brandstetterm:
> personally, I'd speed up the animation a little so the dialog can be interacted with sooner, but that's just my preference. Also, on mobile the slide in looks kinda weird imo

Personally, I don't think the transition duration is too long. 
I've added an additional transition config which will be used for screens smaller than 450px (when the dialog is fullscreen).

PR #2549, opened 2023-01-02
fix: prevent columns from changing their background color when a note is dragged

brandstetterm:
## Description

We've had a bug that the columns changed their background color everytime a note has been dragged around. That's because the `DrayLayer` of our drag & drop provider has been the first child in our Board component. Moving the `DragLayer` to the spot of the last child fixed the problem.


Resaki1 (approved):
crazy that I didn't catch that. ltgm 👍

PR #2594, opened 2023-01-17
Fix: Disable instantaneous drag on mobile

CronJorian:
This is only to be seen as a quickfix for the current dnd solution. On long term the current dnd lib will be replaced.

## Description

Add a small delay to the dnd for touch devices.

## Changelog

- Create a copy of the `HTML5toTouch` variable to add the `delayTouchStart` option.


brandstetterm (approved):
LGTM

PR #2598, opened 2023-01-18
chore: upgrade react, react-dom, react-dnd and react-redux

brandstetterm:
## Description

- Upgrade `react`, `react-dom`, `react-dnd` and `react-redux` 
- Remove `@types/react-dnd.....` since they are deprecated and `react-dnd` packages now provide their own types
- Remove all unit tests mocking dispatch, because it's the wrong way to test these things. Instead, we should add integration tests that tests if a UI has been updated correctly after a dispatch has been fired for example. 
- Add `PropsWithChildren` to function components with children (`children` prop has been removed from `FC`)
- Apply other changes introduced with new react version


brandstetterm:
Seems like there's still an ongoing issue with `react-dnd` that prevents us from upgrading. 
```
/home/runner/work/scrumlr.io/scrumlr.io/node_modules/react-dnd/dist/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){export * from './core/index.js';
                                                                                      ^^^^^^

    SyntaxError: Unexpected token 'export'

      1 | import classNames from "classnames";
      2 | import React, {useRef, useEffect} from "react";
    > 3 | import {useDrag, useDrop} from "react-dnd";
        | ^
```

But luckily, there's already a pull request fixing this issue that is waiting to get reviewed and merged. 
https://github.com/react-dnd/react-dnd/pull/3526

brandstetterm:
I think I've solved the issue. 
```  
"jest": { "transformIgnorePatterns": [ "/node_modules/(?!react-toastify|react-dnd|core-dnd|@react-dnd|dnd-core|react-dnd-html5-backend)" ] },
```

The problem is that `react-dnd` is now ESM only, and Jest doesn't support ESM modules right now (only as an experimental feature). That's why I've added `react-dnd` to the list of `transformIgnorePatterns`. 
I've also included `react-toastify` in the list because there was a problem with the imports of their styling files.

brandstetterm:
**To the reviewers:** Please insure yourself by testing the whole application locally that everything works as it should. I hope nothing broke with the upgraded dependencies.

Resaki1 (changes requested):
Seems mostly fine to me. Only thing I found which doesn't work is the column edit menu, which doesn't open for me (Chrome, Safari and Firefox on MacOS)

Edit: Fixed it by setting `capture` of the clickOutside event handlers to true, which also fixed the mobile menu. While I was at it I also fixed the Stack View header and nav menu, which changed height and width when navigating between columns.

Resaki1 (approved):

PR #2678, opened 2023-02-20
Js/2596 bug hidden notes stack view

### LINKED ISSUE #2596: Bug: Hidden Notes can be seen via StackView
EC:
- "Show notes of other users" is disabled
- StackView should only show visible notes
AC:
- StackView shows all notes in column

https://user-images.githubusercontent.com/44020029/213153769-b4f54da4-3cde-4618-9099-ae97bc9ab78a.mov

Schwehn42:
## Description

Resolves #2596: Users were able to see each others notes in stack view, even if the option was disabled in board settings. 
## Changelog

- Fixed users being to see each others notes in stack view if the option is disabled by filtering them out

## Checklist


Resaki1:
Might be a minor thing, but when the setting is changed to not show notes of other users while in Stack View and the parent note is from another user, it should navigate out of the stack view.
(e.g., user A creates a note, user B opens the note in Stack View. User A changes the setting, but User B still sees the note from user A in Stack View)

Schwehn42:
> Not sure if moderators should see other cards despite the option but i think its good that way :)

@CronJorian yeah you're right about that, tbh I forgot to consider it, but if it turns out moderators should always see all notes, we can easily change it afterwards 👌

CronJorian:
> > Not sure if moderators should see other cards despite the option but i think its good that way :)
> 
> @CronJorian yeah you're right about that, tbh I forgot to consider it, but if it turns out moderators should always see all notes, we can easily change it afterwards 👌 

That's what I thought too, so it's fine that way:)

PR #2722, opened 2023-03-02
Add login provider for Azure Active Directory

seschi98:
## Description

There are already a few login providers such as Microsoft, Apple and GitHub. This PR adds a new login provider for Azure AD, so that users can login with the account provided by their company.

I would also like to discuss the possibility to disable anonymous login (via startup parameter) so login via identity provider can be enforced (I could make a separate PR for that)

## Changelog

- Added new options for `Tenant ID`, `Client ID` and `Client Secret` (Azure AD)
- Added new table to store users that sign in via Azure AD
- Added the configuration for the new provider according to the existing ones
- Added the login button in the frontend part (I re-used the Microsoft icon because AAD is still Microsoft) 

## Checklist

## (Optional) Visual Changes

This is how the new button looks (exactly the same as the others):
image


bitionaire (approved):
LGTM :) 
You can merge it if you want and we'll release it with the next update

bitionaire:
If you want you can add an environment variable to toggle the anonymous login, add the configuration status in the `/info` endpoint https://github.com/inovex/scrumlr.io/blob/main/server/src/api/info.go and toggle the availability of the anonymous login in the client. 

What we plan to to do (probably 2-3 months from now) is to add the option to disable the anonymous login for moderators but leave it activated in order to join boards (so that we can setup boards with our internal accounts and let customers interact with us on boards). 
We can probably design it so that we'll have the option to switch between different states? (e.g. disabled / enabled for participants / enabled)

seschi98:
> What we plan to to do (probably 2-3 months from now) is to add the option to disable the anonymous login for moderators but leave it activated in order to join boards (so that we can setup boards with our internal accounts and let customers interact with us on boards).
> We can probably design it so that we'll have the option to switch between different states? (e.g. disabled / enabled for participants / enabled)

That sounds good, I will keep that in mind when I implement this feature. Thanks for the review 👍

PR #2761, opened 2023-03-13
Hide timer after 15 seconds

Benjosh95:


src/components/Timer/Timer.tsx
```diff
@@ -57,6 +61,7 @@ export const Timer = (props: TimerProps) => {
     if (!previousPlayTimesUpState && playTimesUp) {
       timesUpSoundObject.on("end", () => setPlayTimesUp(false));
       playTimesUpSound();
+      hideTimerAfterSeconds(15);
```
Schwehn42, line 64: You could extract this number into a constant on top so that we don't have random numbers flying around in the code, but since it is only used once anyways I don't really care that much 👀

Schwehn42 (approved):

PR #2783, opened 2023-03-21
feat: show all authors of a note, including stack children

### LINKED ISSUE #2628: Show all author avatars on note (including authors of stacked notes)
As a user I want to see all authors including the authors of stacked notes on my board view.
Similar to the participants

Schwehn42:
## Description
Resolves #2628 
This PR changes the notes in order to display all of its authors.  
We have the following possibilities:
1. Non-stacked note: only the author is displayed, like the way it's been before.
2. stacked note:
    1. up to 3 authors are displayed next to each other, with only the parent author's name being displayed
    2. if there's more than 3 authors, two of them are displayed with their respective avatars, the rest is condensed into a rest bubble.

## Changelog
- created a new component `NoteAuthorList`
- adjusted `Note` logic to also include authors of stack notes
- add/change css in order to display authors correctly

## Checklist

## (Optional) Visual Changes
Bildschirm­foto 2023-03-16 um 13 33 17


src/components/Note/NoteAuthorList/NoteAuthorList.scss
```diff
@@ -0,0 +1,105 @@
+@import "src/constants/style.scss";
+
+.note-author-list {
+  display: flex;
+  flex-direction: row-reverse;
```
Resaki1, line 5: why not just `row` and have the elements in the desired order inside the .tsx?

src/components/Note/NoteAuthorList/NoteAuthorList.scss
```diff
@@ -0,0 +1,105 @@
...
+  justify-content: center;
+  align-items: center;
+
+  border: 1px solid $color-white-one;
+  border-radius: $rounded--full;
+}
+
+// actually affects the first one, since the flex direction is reversed
+// also we only want to change the margin if no rest users exist, since they are displayed on the left
+.note__author:not(.note__author--with-rest):last-of-type {
```
Resaki1, line 54: would need to be updated if the above comment would be addressed

src/components/Note/NoteAuthorList/NoteAuthorList.scss
```diff
@@ -0,0 +1,105 @@
...
+  letter-spacing: $letter-spacing--small;
+  line-height: $line-height--small;
+}
+
+.note__author--self {
+  background: var(--accent-color--desaturated-light);
+}
+
+.note__author--self > .note__user-avatar #Circle-Background {
+  stroke: none;
```
Resaki1, line 71: imo it'd be nice to see if my own note is somewhere inside of a stack

src/components/Note/NoteAuthorList/NoteAuthorList.tsx
```diff
@@ -0,0 +1,51 @@
...
+              <figcaption className="note__author-name">{a.displayName}</figcaption>
+            ) : null}
+          </figure>
+        )
+      )}
+      {restUsersExist && (
+        // wrapper allows for changing background
+        <div className={classNames("note-author-rest__wrapper", {"note-author-rest__wrapper--self": stackAuthor.isSelf})}>
+          <div className="note-author-rest" title={restUsersTitle}>
+            {restAuthors.length}
```
Resaki1, line 45: I'd prefer this to be at the right side of the list so that it is consistent with our board header participants (also makes more sense in my head: "This stack has notes from User A, User B and 3 others"). Could be left here if the flex-direction would be changed to `row`

Resaki1 (dismissed):

src/components/Note/NoteAuthorList/NoteAuthorList.tsx
Schwehn42: The problem here is that the name should align with the authors avatar, otherwise things would just look odd.  
Already with only two authors there would be confusion:
{stack author avatar} {second author avatar} {stack author }

Resaki1: How about this: First the stack author avatar, then the second avatar, then the number of other authors, then stack author name and something like "and 4 others"? Not sure how that would look, just a quick idea

src/components/Note/NoteAuthorList/NoteAuthorList.scss
Schwehn42: Yeah, I was considering to always display yourself as of of the avatars instead of the rest author list, independently of the position in the stack. Would that work for you?

Resaki1: sounds good, lets try that

src/components/Note/NoteAuthorList/NoteAuthorList.scss
Schwehn42: changed 👍

src/components/Note/NoteAuthorList/NoteAuthorList.scss
Schwehn42: yourself should always be visible now: ecaa612

src/components/Note/NoteAuthorList/NoteAuthorList.tsx
Schwehn42: I did end up changing the order after all, the stack author is on the very left now 👍

src/components/Note/NoteAuthorList/NoteAuthorList.scss
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// the rest bubble should be about the same size as the avatars
+.note-author-rest {
+  display: inline-flex;
+  width: $icon--medium;
+  height: $icon--medium;
+
+  font-size: $text-size--small;
+
```
brandstetterm, line 56: Please use our new font size constants here 
```suggestion
  font-size: $text--xs;
```

src/components/Note/NoteAuthorList/NoteAuthorList.scss
```diff
@@ -0,0 +1,123 @@
...
+  border-radius: $rounded--full;
+}
+
+.note-author-rest--self {
+  border-color: $menu-icon-background-color--dark;
+}
+
+.note__author-name {
+  color: $color-dark-gray;
+  font-size: $text-size--small;
```
brandstetterm, line 76:
```suggestion
  font-size: $text--xs;
```

src/components/Note/Note.tsx
```diff
@@ -28,22 +28,47 @@ export const Note = (props: NoteProps) => {
 
   const note = useAppSelector((state) => state.notes.find((n) => n.id === props.noteId), isEqual);
   const isStack = useAppSelector((state) => state.notes.filter((n) => n.position.stack === props.noteId).length > 0);
+  const noteChildren = useAppSelector((state) => state.notes.filter((n) => n.position.stack === props.noteId));
```
brandstetterm, line 31: Since you don't need the whole notes but only the authors, you should already map the notes inside the selector. Through that we can prevent the note from unnecessary rerenderings when the text of child notes has been changed.

brandstetterm: So basically do that mapping from line 41 already inside the `useAppSelector` please

src/components/Note/NoteAuthorList/__tests__/NoteAuthorList.test.tsx
```diff
@@ -0,0 +1,57 @@
+import {ApplicationState} from "types";
+import {Provider} from "react-redux";
+import getTestStore from "utils/test/getTestStore";
+import {NoteAuthorList} from "../NoteAuthorList";
+import {Participant, ParticipantExtendedInfo} from "types/participant";
+import {render} from "testUtils";
+import getTestParticipant from "../../../../utils/test/getTestParticipant";
```
brandstetterm, line 7:
```suggestion
import getTestParticipant from "utils/test/getTestParticipant";
```

src/components/Note/Note.tsx
Schwehn42: I have changed it so all filtering and mapping of `childrenNoteAuthors` is done in the selector, is this what you meant?

brandstetterm (approved):
LGTM 🥳

PR #2887, opened 2023-04-20
feat: add new notifications

Benjosh95:
## Description

Implemented the new notifications based on this design: [https://www.figma.com/file/8m2V6OJTImuOdQoxruWurP/Scrumlr?node-id=409-14916&t=IaQ3uPMTeYoRS7JA-0](url)
 
Added a customToast component, that is rendering "singleToasts" and "multiToasts".
They differ in the size, layout, behaviour, etc. depending on the passed configurations.

**For the review:**
In particular, the files utils/Toast and customToast are important. 
The other changed files are just the appropriate adjustments and the testing.

**Optional follow-up ticket:**
Rework the content of the toasts:
- self-explanatory titles
- add messages (explaining in further detail)
- some toast-types don't fit well
- ...

## Checklist


src/utils/Toast.tsx
```diff
@@ -1,39 +1,173 @@
...
+ * @param options.message The message to display in the toast notification.
+ * @param options.hintMessage An optional hint message to display in the toast notification.
+ * @param options.hintOnClick An optional callback function to execute when the hint message/checkbox is clicked.
+ * @param options.buttons An array of button-labels to display in the toast notification. (for e.g. ["Save", "Cancel"])
+ * @param options.firstButtonOnClick An optional callback function to execute when the first button is clicked.
+ * @param options.secondButtonOnClick An optional callback function to execute when the second button is clicked.
+ * @param options.autoClose The time in milliseconds after which the toast notification should automatically close.
+ * @param options.icon An optional icon to display in the toast notification.
+ * @param options.iconName An optional icon Name. Needs to be set to a any different name than "success", "error", or "info" when an icon is passed aswell.
+ * @param options.type The type of the toast notification. Can be "success", "error", or "info".
```
Schwehn42, line 46: is the param `type` actually required for some use cases? The method names already are semantic, additionally changing the type changes the color (e.g. an error toast with a green icon) which doesn't seem that useful imo

src/components/CustomToast/CustomToast.tsx
```diff
@@ -0,0 +1,136 @@
...
+      if (buttons) {
+        if (buttons.length < 2) {
+          if (titleRef.current.offsetHeight > titleMaxHeightMobile) {
+            setIsSingleToastTitle(false);
+          }
+        } else {
+          setIsSingleToastTitle(false);
+        }
+      } else if (titleRef.current.offsetHeight > titleMaxHeightDesktop) {
+        setIsSingleToastTitle(false);
```
Schwehn42, line 61: this whole if branch tree looks like it can be simplified to be more readable, but I'm not sure

src/components/CustomToast/CustomToast.tsx
Schwehn42, line 1: I noticed some weird behaviour when adding multiple toasts:
1. Initially two toasts are displayed on top of each other; after that, they don't stack but come afterwards. Is this intentional?
2. Toasts seem to repeat themselves

I added a video to demonstrate this behaviour:

https://user-images.githubusercontent.com/7889564/233999452-bc7e54ee-23fc-422d-89cf-efd93351c36e.mov

src/components/CustomToast/CustomToast.scss
```diff
@@ -0,0 +1,466 @@
+@import "constants/style";
+
+// The namings "single" and "mutli" refer to the two types of toasts.
+// The single line toast or a toast that spans multiple lines of content
+
+// These and other darkmode collors from Figma need to be added/moved to style.css
+$color-info-darkmode: #6b9cff;
+$color-error: #ff0069;
+$color-success: #17bf71;
```
Schwehn42, line 9: As the comment suggests, might as well add these to `style.scss`? :)

src/utils/Toast.tsx
Benjosh95: You are right. I removed the type param.

src/components/CustomToast/CustomToast.tsx
Resaki1: Yes, this happens only due to the strict mode and won't appear in prod

brandstetterm: Indeed

src/utils/__tests__/toast.test.tsx
```diff
@@ -0,0 +1,27 @@
+import {render, screen} from "@testing-library/react";
+import {ToastContainer} from "react-toastify";
+import {Toast, Options} from "utils/Toast";
+
+const testOptions: Options = {
+  title: "Titeliger titel",
+  message: "1 heftige Message mit gaanz super Buchstabän drinnnen",
+  hintMessage: "wennd mogsch zeig ichs nemmer o",
```
Schwehn42, line 8: 😂

src/components/CustomToast/CustomToast.tsx
```diff
@@ -0,0 +1,136 @@
...
+import {ReactComponent as CloseIcon} from "assets/icon-close.svg";
+import {ToastTypes} from "utils/Toast";
+import classNames from "classnames";
+
+export interface CustomToastProps {
+  title: string;
+  message?: string;
+  hintMessage?: string;
+  hintOnClick?: () => void;
+  buttons?: string[];
```
Schwehn42, line 12: you are able to add as many buttons as you want, however these look weird and also don't have any functionality. Maybe changing this to `primaryButton?` and `secondaryButton?` or just cutting off the extra buttons would solve this?

src/components/CustomToast/CustomToast.scss
Schwehn42, line 1: Since this is way to complicated, I guess we'll just leave it at that and hope we'll never have to touch it again. :)  
 One thing that would be nice is to make the buttons not break on the screen size for iPhone SE. so around at 400px.

Schwehn42 (approved):

PR #2961, opened 2023-05-10
improvement: add new font size constants

brandstetterm:
## Description
With the design of the new landing- and error page, the problem of inconsistent text sizes once again occurred. Since we've only had three text size constants so far, we've decided to extend the list with sizes for all possible use cases. 

H0 - Hero Title - 62px (Sondergröße)
H1 - Big Title - 40px
H2 - Large - 32px
H3 - Medium - 24px
P1 - Large - 20px
P2 - Medium - 16px
P3 - Small - 12px

Old constants haven't been replaced with new ones yet. This should be done step by step. New designs & implementations should use the new constants.

### DIFF
--- a/src/constants/style.scss
+++ b/src/constants/style.scss
@@ -141,6 +141,15 @@ $line-height--small: 16px;
 $line-height--medium: 20px;
 $line-height--large: 32px;
 
+// new font size constants
+$text--small: 12px;
+$text--base: 16px;
+$text--medium: 20px;
+$text--large: 24px;
+$text--xl: 32px;
+$text--2xl: 40px;
+$text--3xl: 62px;
+
 // filters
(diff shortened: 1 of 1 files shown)


Schwehn42 (approved):
looks good, just two questions:
1. are there plans to gradually replace the old constants with the new ones?
2. do we need to set the root font size explicitly or is assuming the default in most browser is `16px` enough?

brandstetterm:
> looks good, just two questions:
> 
> 1. are there plans to gradually replace the old constants with the new ones?
> 2. do we need to set the root font size explicitly or is assuming the default in most browser is `16px` enough?

1. Yes, we should replace the old constants with the new ones gradually.
2. The default root font size of HTML is 16px as long as someone hasn't changed it intentionally.

PR #2922, opened 2023-04-30
add action-toast for unready users when timer is up

Benjosh95:
## Description

Added a toast when timer is up that asks whether users are ready.
The button on that toast can be clicked to mark themselves as ready

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] The light- and dark-theme are both supported and tested
- [x] The design was implemented and is responsive for all devices and screen sizes
- [x] The application was tested in the most commonly used browsers (e.g. Chrome, Firefox, Safari)

ready-toast

PR #2965, opened 2023-05-11
feat: "Mark me as done" for participants in timer/voting pill

brandstetterm:


Schwehn42:
Is this the same as #2922?

brandstetterm:
> Is this the same as #2922?

No, it's something different. This pull requests adds something to the timer & vote display in the info bar.

src/components/Timer/Timer.scss
```diff
@@ -84,48 +83,64 @@ $timer__short-actions-tooltip-height: 36px;
...
+}
+.short-action__button--ready:hover > .short-action__check-icon {
+  visibility: hidden;
+}
+.short-action__button--ready:hover > .short-action__cancel-icon {
+  visibility: visible;
+}
+.short-action__button > .short-action__check-icon {
+  height: $timer__short-actions-height;
+  width: $timer__short-actions-height;
```
Schwehn42, line 103: personally, I like nesting selectors like these in one big statement like this
```scss
.selector {
  .child {
    /* ... */
  }
  :&hover {
    /* ... */
    .child {
       /* ... */ 
     }
  }
}
```
because I get a better overview of what is affected when  
But of course, your solution is not wrong or anything, just preference

brandstetterm: I don't like nesting that much 😢

PR #3121, opened 2023-06-20
fix: proper Icon for note deletion toast

Resaki1:
## Description

Changed the toast upon note deletion from the previous error toast to a new toast with a trash icon. Closes #3095 

## Changelog

- added new deletion toast with trash icon
- changed note deletion toast from error to deletion

## Visual Changes

Bildschirmfoto 2023-06-20 um 16 22 59

PR #3141, opened 2023-06-25
fix: redirect after board is deleted

### LINKED ISSUE #3097: Automatically transition to landingpage
At the moment when the board owner deleted the board all the participants are still there. They should get transferred back zu our landingpage + toast (board has deleted).

Resaki1:
## Description

When a board is deleted, all users on the board are now redirected to the homepage including a toast. Closes #3097 

## Changelog

- added redirect to homepage
- when opening the homepage after board deletion, a toast is shown


src/store/middleware/board.tsx
```diff
@@ -39,6 +39,7 @@ export const passBoardMiddleware = (stateAPI: MiddlewareAPI<Dispatch, Applicatio
 
         if (message.type === "BOARD_DELETED") {
           store.dispatch(Actions.leaveBoard());
+          window.location.assign("/?boardDeleted=true");
```
brandstetterm, line 42: Why don't you dispatch the board deleted toast here?

Resaki1: Because the toast disappears as soon as the redirect is finished

src/routes/Homepage/Homepage.tsx
```diff
@@ -31,6 +33,16 @@ export const Homepage = withTranslation()(() => {
     dispatch(Actions.signOut());
   };
 
+  useEffect(() => {
+    const boardDeleted = new URLSearchParams(window.location.search).get("boardDeleted");
+
+    if (boardDeleted) {
+      Toast.error({
```
brandstetterm, line 40: This should be an info toast and not an error toast I guess.

Resaki1: Now that you say that, I think it should be the new deletion toast from #3121, shouldn't it?

brandstetterm: No, it's a simple information for the participants that the board they have been on has been deleted. 

The deletion toast would be for the moderator that has deleted the board.

public/locales/de/translation.json
```diff
@@ -389,6 +389,7 @@
     "serverConnection": "Der Server ist leider nicht erreichbar. Bitte versuche die Seite neu zu laden.",
     "retry": "Wiederholen",
     "joinBoard": "Es tut uns leid, aber wir haben ein Problem beim Beitreten des Boards. Bitte versuche es erneut.",
+    "boardDeleted": "Das Board wurde gelöscht! Du wirst zur Startseite weitergeleitet.",
```
brandstetterm, line 392:
```suggestion
    "boardDeleted": "Das Board wurde gelöscht! Du wurdest zur Startseite weitergeleitet.",
```

PR #3058, opened 2023-06-06
feature: note reactions

Schwehn42:
## Description

resolves #2964 

this PR adds emoji reactions to notes.
Features:
- New "Reaction"-Button in the bottom left corner of notes.
- Clicking it will show the Reaction Bar, with a list of possible emojis to choose from
- You can have **one** reaction **per** user **per** note.
- Reactions are displayed next to each other, with your own reaction always being highlighted
- If there are many reactions, they will be condensed to view all of them 
- You can easily remove your own reaction by clicking it or change it by clicking another one
- You can also long press it to show the reactions popup list, where you can easily see who reacted (indended for mobile)
- Reactions are live updated for all participants.
- Option for moderators to disable reactions
 
[//]: <> (Explain the changes you’ve made. It doesn’t need to be fancy and you don’t have to get too technical.)

## Changelog

Backend:
- Add new dto types for `Reaction` and Requests
- Add Service and endpoints for creating, fetching, updating, and deleting reactions
- Add Database queries for above things
- Add Observer for notifying events
- Add `ReactionContext`
- adjust board to send reactions
- add migration scripts
- update postman collection
- tests

Frontend:
- types
- store:
    - actions for creating, deleting, updating reactions
    - reducer
    - middleware
    - api
 - new hooks for util
 - components:
    - adjust `Note` component to display authors/votes on top, reactions on bottom
    - new component `NoteReactionList` to see all reactions and possibility to add/change/remove reaction
       - `NoteReactionBar` (to add reaction)
       - `NoteReactionChip` (display single reaction)
       - `NoteReactionChipCondensed` (display multiple reactions on limited space)
       - `NoteReactionPopup` (mobile dialog to easily view all reactions)
       - `TooltipPortal` (put styled tooltips anywhere)
  - tests

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] The light- and dark-theme are both supported and tested
- [x] The design was implemented and is responsive for all devices and screen sizes
- [x] The application was tested in the most common used browsers (e.g. Chrome, Firefox, Safari)

## (Optional) Visual Changes
<details>
<summary>visual changes</summary>

reactions-preview

notes-popup-preview
</details>

PR #3170, opened 2023-07-03
refactor: move ParticipantExtendedInfo logic

Schwehn42:
## Description

this PR moves the logic of changing the `Participant[]` to `ParticipantExtendedInfo[]` array used in `NoteAuthorList`.  
semantically nothing should change, the reason I did this is so the component can be used elsewhere, and we don't have to repeat the logic from the `Note` component, but instead just input the Participants.

**Note**: I might also change the name of the component since it will also be used in #3058 to display who made a reaction

## Changelog

* move logic from `Note` component to `NoteAuthorList` component
* adjust tests because the type is now different in `Note`

## Checklist


src/components/Note/Note.tsx
```diff
@@ -32,47 +32,19 @@ export const Note = (props: NoteProps) => {
...
-    }
-
-    // if showAuthors is disabled, we still want to see cards written by yourself if you're the stack author.
-    // the other authors are excluded as we only require the stack author
-    if (!showAuthors && props.viewer.user.id === noteAuthor!.user!.id) {
-      return [allAuthors[0]]; // stack author is always first element
-    }
-
-    return allAuthors;
+    return [noteAuthor, ...childrenNoteAuthors].filter((a): a is Participant => !!a); // remove undefined values (which shouldn't exists, only for TS type assertion)
```
Resaki1, line 47:
```suggestion
    return [noteAuthor, ...childrenNoteAuthors].filter(Boolean) as Participant[]; // remove undefined values (which shouldn't exists, only for TS type assertion)
```

This is more readable imo, what do you think?

Resaki1 (approved):
Looks good to me, good job. Have a look at the suggestion I made, but if you don't like it just merge this PR

PR #3187, opened 2023-07-10
fix: stack view author avatar  

brandstetterm:
## Description
I accidentally deleted a line in the styling which missaligned the avatar.

## Changelog
- Add back line to realign the avatar

## (Optional) Visual Changes
Before:
Screenshot 2023-07-10 at 18 55 25

After:
Screenshot 2023-07-10 at 18 55 01


Schwehn42 (approved):
this works, although we should consider just reusing the NoteAuthorList to avoid duplicate code

brandstetterm:
> this works, although we should consider just reusing the NoteAuthorList to avoid duplicate code

I agree. We could think about reusing the NoteAuthorList there too. As a next step, we could try to refactor it. For now this should be good enough.

PR #3196, opened 2023-07-11
improvement: add ready indication of user avatars to participants list and prevent animations on reload

brandstetterm:
## Description
I've noticed that the ready indicator of the user avatars hasn't been visible in the participants list, even tho it's very useful. On boards with a high number of participants, you cannot see all participants on the top right corner and therefore cannot find out which participants are ready. That's why the proposed changes are highly useful for moderators. 

I've also added some code to prevent the ready animation on reload. With the proposed changes, the animation of the ready indicator will only be played if the ready state has been changed from false to true. On reload, the ready indicator will be displayed without animation. 

## Changelog
- Show ready indicators on user avatars in the participants list
- Add `useEffect` and `useRef` to UserAvatar to prevent the ready animation on reload

## (Optional) Visual Changes

imageimage


Resaki1 (approved):
good job 👍

PR #3198, opened 2023-07-12
refactor: center the avatar in its container

brandstetterm:
## Description

I've noticed that the avatar itself isn't centered in it's container. Therefore ugly positioning using margins, paddings etc. was necessary.

## Changelog

- Center avatar inside it's container using `transform: translateY(...)`
- Remove unnecessary `top` and `margin-bottom` used for aligning the avatar

## (Optional) Visual Changes
Before:
Screenshot 2023-07-12 at 11 10 42

After:
Screenshot 2023-07-12 at 11 10 07

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@types/workbox-precaching": "^5.0.0",
     "@types/workbox-routing": "^5.0.0",
     "@typescript-eslint/eslint-plugin": "^5.62.0",
-    "@typescript-eslint/parser": "^5.61.0",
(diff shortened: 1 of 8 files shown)


Resaki1 (approved):
Looks good.
One thing that has bugged me for a while now tho is that the add-user button is slightly smaller than the avatars. Should we fix that, and if so, do you want to do it in this PR or should we open a new issue?

Resaki1:
Bildschirmfoto 2023-07-12 um 15 58 38
I just noticed that the focus border is off-centered

PR #3346, opened 2023-09-11
fix: hide board urls in analytics

Lennart01:
## Description
Hashes Board URLs if analytics are enabled.
Prevents board URLs from showing up in the analytics data.


src/index.tsx
```diff
@@ -28,8 +28,21 @@ if (ANALYTICS_DATA_DOMAIN && ANALYTICS_SRC) {
     domain: ANALYTICS_DATA_DOMAIN,
     apiHost: ANALYTICS_SRC,
   });
-
-  trackPageview();
+  const handleAnalytics = async () => {
+    if (window.location.href.includes("/board/")) {
+      const [baseUrl, boardId] = window.location.href.split("/board/");
```
Resaki1, line 33: This can lead to different boardIds for the same board, because we have URLs with more info after the boardId (e.g., for stackview, it is `{baseUrl}/board/{boardId}/note/{noteId}/stack` or for settings it is `{baseUrl}/board/{boardId}/settings`). Is this intended?

If this is not intended, one approach might be to only keep the first 36 chars after "/board/", since we are using uuids with the length of 36, and omit everything after that using slice()

Resaki1 (changes requested):

Resaki1 (approved):

Resaki1:
lgtm

Issue #3267, opened 2023-08-09
Revise color palette


SelinaBuff:
Over time, new colors have been added and changed from time to time. Let's bring everything together and make our color palette consistent for light & dark mode.

![Image](https://github.com/inovex/scrumlr.io/assets/116653219/d86bd037-f15c-446e-9dfe-541a6a16c964)

_If you have any questions or uncertainties, please contact Selina or Manu._

Resaki1:
Please also keep accessibility in mind!
For example, right now the contrast on e.g. buttons is too low in both light and dark mode:


![Image](https://github.com/inovex/scrumlr.io/assets/35272402/e8c81fe8-71bd-405e-87f9-d083100e161a)
![Image](https://github.com/inovex/scrumlr.io/assets/35272402/558767a6-91b6-497e-87aa-4ee942bd4887)


This site has helpful information and tools regarding color contrast and a11y: https://dequeuniversity.com/rules/axe/4.7/color-contrast

SelinaBuff:
The colors were proved and for our use – white medium/bold letters and icons on buttons (Ui components) they will be fine.

![Image](https://github.com/inovex/scrumlr.io/assets/116653219/66589cf2-a7cb-4e78-9f7e-497648985df3)

![Image](https://github.com/inovex/scrumlr.io/assets/116653219/c3491df0-4652-49ee-b5d2-cf398b6b3b88)

PR #3357, opened 2023-09-15
feat: board reactions

### LINKED ISSUE #177: Emojis on board to display own reactions

Schwehn42:
## Description

resolves #177
this PR adds board reactions.  
includes:
- a board reaction menu (can be openened just like the other menus or per shortcut `R`)
- select between multiple emojis (click or shortcuts `1`through `5`)
- reaction is displayed with user name for all users
- reaction has an animation to fly to the top and fade out after 5s
- setting to disable board reactions for oneself

## Changelog
frontend:
- add `BoardReaction` type
- add components
    - `BoardReactionMenu`
    - `BoardReactionContainer`
    - `BoardReaction`
- adjust `MenuBars`
- add board reaction actions in store (actions, reducer, mw)
- add `showBoardReactions` flag to `view` state
- add option in `Appearance` menu to change state
- state is saved in `LocalStorage`
- styles

backend:
- add `BoardReaction` type
- add `board-reactions` route
- add server event and notify all clients

## Checklist


public/locales/de/translation.json
```diff
@@ -266,7 +270,9 @@
     "noteMovedToStack": "Die Karte wurde in einen Stapel verschoben!",
     "noteDeleted": "Die Karte wurde gelöscht!",
     "ready": "Bist du fertig?",
-    "readyButton": "Fertig!"
+    "readyButton": "Fertig!",
+    "boardReactionsDisabled": "Board Reaktionen sind deaktiviert",
```
Resaki1, line 274:
```suggestion
    "boardReactionsDisabled": "Board-Reaktionen sind deaktiviert",
```

public/locales/en/translation.json
```diff
@@ -239,6 +240,9 @@
     "customTime": "Custom time",
     "cancelTimer": "Cancel timer"
   },
+  "BoardReactionsMenu": {
+    "react": "React {{reaction}}"
```
Resaki1, line 244:
```suggestion
    "react": "React with {{reaction}}"
```
(not sure about this)

src/api/boardReaction.ts
```diff
@@ -0,0 +1,24 @@
+import {SERVER_HTTP_URL} from "config";
+import {BoardReactionType, ReactionType} from "types/reaction";
+
+export const BoardReactionAPI = {
+  addBoardReaction: async (boardId: string, reactionType: ReactionType) => {
```
Resaki1, line 5: I don't know if this is the right location for this, and also it's just a general idea, but: Should we add some kind of limit/debounce to sending board reactions? Right now you can spam them to the point where the application starts to lag, and that's just from one person doing it.

src/components/BoardReaction/BoardReaction.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const BoardReaction = memo((props: BoardReactionProps) => {
+  const emoji = BoardReactionImageMap.get(props.reaction.reactionType);
+  const me = useAppSelector((state) => state.participants!.self);
+  const others = useAppSelector((state) => state.participants!.others);
+  const all = [me, ...others];
+  const reactionUser = all.find((p) => p.user.id === props.reaction.user)!;
+  const {name} = reactionUser.user;
+  const reactedSelf = reactionUser.user.id === me.user.id;
+
+  const leftOffset = getRandomNumberInRange(5, 85);
```
Resaki1, line 23: On mobile, the board reactions are sometimes cut off be the right side of the screen. It's not bad, but might be worth to take a look at
Bildschirmfoto 2023-09-28 um 09 40 44

Resaki1: you could do the offset purely in scss (which imo would be better anyway) and then change the value using a media query

src/components/BoardReactionMenu/BoardReactionMenu.scss
```diff
@@ -0,0 +1,100 @@
...
+.board-reactions__root {
+  position: absolute;
+  display: flex;
+  justify-content: center;
+  width: 100%;
+  bottom: $column__border-width;
+}
+
+.board-reactions__container {
+  width: 374px;
```
Resaki1, line 12:
```suggestion
```
It's not needed as far as I can tell

src/components/BoardReactionMenu/BoardReactionMenu.scss
```diff
@@ -0,0 +1,100 @@
...
+  width: 54px;
+  height: 54px;
+
+  font-size: $text--xl;
+
+  border-radius: $rounded--full;
+
+  span {
+    margin-left: 0.3em; // manually center those stubborn emojis
+  }
```
Resaki1, line 51:
```suggestion
display: flex;
justify-content: center;
align-items: center;
```

src/components/BoardReactionMenu/BoardReactionMenu.scss
```diff
@@ -0,0 +1,100 @@
...
+
+  .board-reactions__close {
+    color: $color-white;
+  }
+}
+
+@media screen and (max-width: 400px) {
+  .board-reactions__container {
+    width: 300px;
+    gap: 4px;
```
Resaki1, line 98:
```suggestion
    gap: 0;
```

src/components/MenuBars/MenuBars.tsx
```diff
@@ -170,64 +210,81 @@ export const MenuBars = ({showPreviousColumn, showNextColumn, onPreviousColumn,
...
-                <TooltipButton direction="right" label={t("MenuBars.settings")} onClick={showSettings} icon={SettingsIcon} />
+              <li
+                className={classNames("menu-bars-mobile__fab-option", "menu-bars-mobile__fab-option--horizontal", {"menu-bars-mobile__fab-option--active": showBoardReactionsMenu})}
+              >
+                <TooltipButton
+                  active={showBoardReactionsMenu}
+                  direction="left"
+                  label={t("MenuBars.openBoardReactionMenu")}
+                  icon={BoardReactionIcon}
+                  onClick={toggleBoardReactionsMenu}
```
Resaki1, line 253: for some reason, clicking this button again to close the reaction menu doesn't work for me on mobile

src/constants/style.scss
```diff
@@ -26,6 +26,9 @@ $tooltip-background-color--light: #d3daf0;
 $tooltip-background-color--dark: #485064;
 $color-progress-circle: #3c7fff;
 $color-icon-light-blue: #9ebfff;
+$color-board-reaction-name--dark: #5a647d;
+$color-board-reaction-name-self--light: #0057ff;
+$color-board-reaction-name-self--dark: #6096ff;
```
Resaki1, line 31: why not just use the colors we already have?

src/store/middleware/boardReaction.tsx
```diff
@@ -0,0 +1,10 @@
+import {Dispatch, MiddlewareAPI} from "@reduxjs/toolkit";
+import {ApplicationState} from "types";
+import {Action, ReduxAction} from "store/action";
+import {API} from "api";
+
+export const passBoardReactionMiddleware = (stateAPI: MiddlewareAPI<Dispatch, ApplicationState>, dispatch: Dispatch, action: ReduxAction) => {
+  if (action.type === Action.AddBoardReaction) {
+    API.addBoardReaction(action.context.board!, action.reactionType); // TODO handle promise & toasts
```
Resaki1, line 8: is this TODO still to do? :)

src/store/reducer/boardReaction.ts
```diff
@@ -0,0 +1,14 @@
+import {Action, ReduxAction} from "store/action";
+import {BoardReactionState} from "types/reaction";
+
+// eslint-disable-next-line @typescript-eslint/default-param-last
+export const boardReactionReducer = (state: BoardReactionState = [], action: ReduxAction): BoardReactionState => {
```
Resaki1, line 5: why do you need the default value for state here? Can it, and the eslint-disable, be removed?

src/utils/boardReaction.ts
```diff
@@ -0,0 +1,21 @@
+import {BoardReactionType} from "types/reaction";
+
+// these are custom events that are can be emitted and subscribed to.
```
Resaki1, line 3:
```suggestion
// these are custom events that can be emitted and subscribed to.
```

src/utils/random.ts
```diff
@@ -7,3 +7,13 @@ export const generateRandomString = (length = 8) => {
...
+
+/**
+ * returns a random whole number in the given range
+ * @param min lower bound
+ * @param max upper bound
+ */
+export const getRandomNumberInRange = (min: number, max: number): number => {
+  const range = max - min + 1;
+  return Math.floor(Math.random() * range) + min;
+};
```
Resaki1, line 19: If you choose to do it purely in scss as I have suggested above, this can be removed

src/components/BoardReaction/BoardReaction.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <div className="board-reaction__root" style={{left: `${leftOffset}vw`}}>
+      <div className="board-reaction__emoji-container">{emoji}</div>
+      <div className={classNames("board-reaction__name-container", {"board-reaction__name-container--self": reactedSelf})}>{name}</div>
```
Resaki1, line 28: If the reaction is by myself, it shouldn't display my name here imo, but just "me" (just like the name on notes)

Resaki1 (changes requested):
Good stuff! Just a few improvements that can be made, but other than that looks good!

src/components/BoardReactionMenu/BoardReactionMenu.tsx
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  // hotkey is converted to a number which is then used as the index for the reaction type
+  useHotkeys(["1", "2", "3", "4", "5"], (e, k) => onClickReaction(e, boardReactions[+k.keys![0] - 1][0]));
+
+  return (
+    <div className="board-reactions__root">
+      <div className="board-reactions__container">
+        {boardReactions.map(([reactionType, emoji]) => (
+          <button
```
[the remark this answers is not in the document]
Schwehn42: i solved it provisionally using a title, but it can be changed later to be a proper tooltip. :)

src/components/BoardReactionMenu/BoardReactionMenu.scss
```diff
@@ -0,0 +1,100 @@
...
+  border-radius: $rounded--full;
+
+  span {
+    margin-left: 0.3em; // manually center those stubborn emojis
+  }
+
+  &:hover {
+    background-color: $hover-background-light;
+  }
+
```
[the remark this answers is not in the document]
Schwehn42: sure, what should it look like? Do note that the the button remains focused after clicking it though

src/components/BoardReaction/BoardReaction.scss
```diff
@@ -0,0 +1,56 @@
...
+  font-size: $text--xl;
+}
+
+.board-reaction__name-container {
+  padding: 10px;
+  font-size: $text--base;
+  font-weight: 600;
+  background-color: $color-white-one;
+  border-radius: $rounded--large;
+
```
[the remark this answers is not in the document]
brandstetterm: Wouldn't it look better with less border-radius? Looks a bit too much for me

src/components/BoardReaction/BoardReaction.tsx
Schwehn42: > you could do the offset purely in scss (which imo would be better anyway) and then change the value using a media query

what should the behaviour be in this case? The overflow is visible right now, leading to a scrollbar which is ugly. Changing the overflow to hidden just cuts it off, which is a little better but for long names still annoying.  
Since the offset of the board reactions is choosen at random, maybe we could take the name length into the equasion  to thwart the overflow?  
 
> you could do the offset purely in scss (which imo would be better anyway) and then change the value using a media query

there's the random function in scss, but isn't the number fixed by the time of compilation? so the reactions would still always  appear at the same place

Schwehn42: see [this entry](https://blog.codepen.io/2013/08/26/random-function-in-sass/#random-at-runtime)

src/store/reducer/boardReaction.ts
Schwehn42: yes, the default value seems to be required, but my IDE didn't like order of the params; state should be last in its opinion. I disabled the inspection now though.

src/constants/style.scss
Schwehn42: i just followed the design here 🤷 maybe @SelinaBuff could take a look?

src/store/middleware/boardReaction.tsx
Schwehn42: depends on whether we want to do something in case something goes wrong. do we want to display a toast or is that too much for something as small as a board reaction?

src/api/boardReaction.ts
Schwehn42: I tried to implement a hook that can debounce multiple things so we can keep track of each reaction and maybe even have a small visual timer (like the background of the emoji changing progress circle), but something didn't work so I'll have to take another look ^^

src/components/BoardReactionMenu/BoardReactionMenu.scss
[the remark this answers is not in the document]
Resaki1: I'd say a simple border/outline would be fine. You can use `:focus-visible` instead of just `:focus` so that the focus state is not visible after clicking the button: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible

src/api/boardReaction.ts
Resaki1: I feel like a little debounce hook (we could just use a library for that) on the reactions in general would be enough, so that one user can only send a limited amount of reactions per second would be enough, without any visual timer. I think this is how Google Meet does it.
Imo, for everything else we could create a new issue. What do you think?

src/components/BoardReaction/BoardReaction.tsx
Resaki1: Oh yeah, you're right about the random numbers in scss, my bad.
About the overflow: Is it possible to just decrease the leftOffset for mobile users so that reactions don't appear so far to the right of the screen? I think taking the name into account would be a bit overkill here

src/components/BoardReactionMenu/BoardReactionMenu.scss
Resaki1: weird, it works for me:
Bildschirmfoto 2023-10-06 um 09 48 32

What browser are you using?

This is the code:

```
.board-reactions__reaction {
  all: unset; // reset button
  cursor: pointer;

  width: $reaction-size;
  height: $reaction-size;

  font-size: $text--xl;

  border-radius: $rounded--full;

  display: flex;
  justify-content: center;
  align-items: center;

  &:hover {
    background-color: $hover-background-light;
  }

  &--disabled {
    filter: grayscale(100%);
    cursor: not-allowed;
  }
}
```

src/store/middleware/boardReaction.tsx
Resaki1: I'd say it's too much

src/api/boardReaction.ts
Schwehn42: 👍 agreed

src/components/MenuBars/MenuBars.tsx
Schwehn42: fixed

src/components/BoardReaction/BoardReaction.tsx
Schwehn42: it should now take the size of the container into account, so it's never cut off 👍

src/api/boardReaction.ts
Schwehn42: debounce is now implemented

src/constants/style.scss
Schwehn42: the colors are correct, I'll keep them as is, they just have to be renamed when #3267 is done.

Resaki1 (changes requested):
Works beautifully, well done!
Just one more tiny detail: The button to close the reaction menu is not focusable by tabbing. Could you fix that as well? Then this can be merged 👏

src/components/BoardReactionMenu/BoardReactionMenu.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const onClickReaction = (e: React.MouseEvent<HTMLButtonElement> | KeyboardEvent, reaction: ReactionType) => {
+    e.stopPropagation();
+    if (!showBoardReactions) {
+      Toast.info({
+        title: t("Toast.boardReactionsDisabled"),
+        buttons: [t("Toast.enable")],
+        firstButtonOnClick: () => dispatch(Actions.setShowBoardReactions(true)),
+      });
+    } else if (debounce) {
+      // not ready yet
```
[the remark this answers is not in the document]
Schwehn42: huh, I think when I wrote this originally it was 
```typescript
else if (debounce) {
    return; // not ready yet
}
```

but the return was removed by the linter on commit automatically, because it's unnecessary syntactically 🙃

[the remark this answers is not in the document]
Schwehn42: i tried to add it without hooks but then the linter will fail 🤷

Resaki1 (approved):

PR #3374, opened 2023-09-22
feat: record metrics of deleted boards

BenedictHomuth:
## Description
When a board is deleted, all data is removed as well. This PR introduces a before-delete trigger, which records some metrics about the deleted board. Used for analytic purposes.

## Changelog
Added an up and down SQL file containing a `deleted_boards` table schema, the before delete trigger and its corresponding function.

## Checklist


brandstetterm:
Why do we need this? Has there been an issue for that?

server/src/database/migrations/sql/8_add_analytics_board_deletion_trigger.up.sql
```diff
@@ -0,0 +1,62 @@
...
+    id uuid not null primary key,
+    access_policy access_policy,
+    total_columns int, 
+    hidden_columns int,
+    total_notes int,
+    hidden_notes int,
+    first_note_created timestamptz,
+    last_note_created timestamptz,
+    avg_chars_per_note int,
+    total_users int,
```
bitionaire, line 11: `total_participants`

server/src/database/migrations/sql/8_add_analytics_board_deletion_trigger.up.sql
```diff
@@ -0,0 +1,62 @@
...
+    total_admins int,
```
bitionaire, line 12: `total_moderators`

server/src/database/migrations/sql/8_add_analytics_board_deletion_trigger.up.sql
```diff
@@ -0,0 +1,62 @@
...
+        total_votings int;
+
+        BEGIN 
+            -- Retrieve all needed stats
+            SELECT count(id) INTO total_columns FROM columns WHERE board = OLD.id;
+            SELECT count(id) INTO hidden_columns FROM columns c WHERE board = OLD.id AND c."visible" = false;
+
+            SELECT count(id) INTO total_notes FROM notes WHERE board = OLD.id;
+            SELECT count(n.id) INTO hidden_notes FROM notes n INNER JOIN columns c ON n."column" = c.id WHERE c."visible" = false AND n."board" = OLD.id;
+            SELECT created_at INTO first_note_created FROM notes WHERE board = OLD.id ORDER BY created_at LIMIT 1;
```
bitionaire, line 41: Maybe select min/max instead? Dunno whether that's faster than order by

server/src/database/migrations/sql/8_add_analytics_board_deletion_trigger.up.sql
```diff
@@ -0,0 +1,62 @@
...
+            SELECT created_at INTO last_note_created FROM notes WHERE board = OLD.id ORDER BY created_at DESC LIMIT 1;
+            SELECT AVG(LENGTH(text)) INTO avg_chars_per_note FROM notes WHERE board = OLD.id;
+
+            SELECT count(user) INTO total_users FROM board_sessions WHERE board = OLD.id;
```
bitionaire, line 45: Add `WHERE` clause here for role is `PARTICIPANT`

bitionaire (changes requested):

server/src/database/migrations/sql/8_add_analytics_board_deletion_trigger.up.sql
BenedictHomuth: TL;DR: min/max is simpler to read. Performance is not noticeably different in local setup.

I played around with `explain` & `analyze` in order to get a deeper look into the query. 
The output, as I understand it, is that the min/max creates less overhead. In terms of execution time, I could not spot a noticeable difference (both bouncing in the same time range). This might differ from the actual prod-system.

Here is the output of the analyzed queries:

**Order By:**
orderby png


**Min/Max:**
min

bitionaire (changes requested):
You have to bump the migration to no. 9 since we merged the reactions by now 😅

bitionaire (approved):

PR #3380, opened 2023-09-26
improvement: show toast when note dialog contains an empty string

BenedictHomuth:
## Description

When you want to update a note text field into an empty string, a request is sent to the backend. This fails and responds with a 500 internal server error since the database does not allow an empty string for a note text.

This PR contains a client-side check if the note dialog is empty. If this is the case, a toast is displayed, which informs the user that he needs to enter text. Furthermore, there is no request sent to the backend if the string is empty.

## Changelog

- added a check for an empty string on note text change
- added toast messages in i18n

## Checklist

## (Optional) Visual Changes
Before: 
Bildschirm­foto 2023-09-26 um 17 12 55

After:
Bildschirm­foto 2023-09-27 um 11 36 50


Resaki1:
I am sorry if this is a little late, and I don't know if this has been discussed before, but I feel like it would be better for UX if, instead of showing an error toast, we just reset the text to the last non-empty value. What do you think?

BenedictHomuth:
> I am sorry if this is a little late, and I don't know if this has been discussed before, but I feel like it would be better for UX if, instead of showing an error toast, we just reset the text to the last non-empty value. What do you think?

That is fine with me. I just wanted to get rid of the "unnecessary" backend call.

Resaki1 (approved):

Issue #3263, opened 2023-08-09
Stretch dashed background in the invisible column


SelinaBuff:
**User Stories:**

> 1. As a user, I want the dashed background to extend to the end of the invisible column, to ensure a seamless visual transition and avoid a choppy look.

![Image](https://github.com/inovex/scrumlr.io/assets/116653219/647bec15-a0e7-43a7-9111-f2c9e9e40e82)

![Image](https://github.com/inovex/scrumlr.io/assets/116653219/ab2ab119-b86a-4b4a-8845-daba9c5f446e)

Issue #3048, opened 2023-06-02
Toasts are displaying an incorrect border-radius on the Firefox browser


brandstetterm:
### The problem

Our toasts are displaying an incorrect border-radius when using the Firefox browser. Safari & Chrome are displaying the correct border-radius. 

### Browser

Firefox 113.0.2 (64-bit)

### Screenshots

<img width="508" alt="Screenshot 2023-06-02 at 21 52 57" src="https://github.com/inovex/scrumlr.io/assets/36969812/076d73ec-b48c-449a-a225-b7b86b201326">

brandstetterm:
The cause for this issues is that Firefox doesn't support the CSS `:has()` selector yet. Firefox plans to fully support the selector with version 116.0 (current version is 114.0), they just have to fix a few more bugs. Therefore, I don't think it's worth working on a complex workaround and just leave it as it is since it will fix itself (hopefully) soon. The toast without the rounded borders thankfully isn't too ugly.

brandstetterm:
This issue has been resolved since Firefox supports the CSS `:has()` selector since version 121

PR #3403, opened 2023-10-03
feat: seamless stripe transition between columns and stripes on board spacers

blxckOxymoron:
## Description

- create custom hook to offset and align the stripes on every column. 
  - This removes the "break" in the stripes when two adjacend columns are hidden
- add stripes to the spacers on the side of the boards to implement #3263

## Changelog

- create the `useStripeOffset` hook
  - calculates the gradient offset so that every gradient using this "connects" 
  - updates the gradient offset whenever the element size changes using a ResizeObserver
- add the `useStripeOffset` to the column as well as the two spacers
- add a useEffect to update the stripe offset whenever a column is added or removed
- move the gradient declaration to the `constants/style.scss` file as `$column-stripes--light` and `$column-stripes--dark`
- add css rules to check for hidden columns next to the spacers and apply the same stripe-gradient background
  - ⚠️ this uses an `:has()` rule, which isn't supported by firefox yet but will be (#3048)
- add a mock-`ResizeObserver` to the Board and Column tests
- update the Column snapshot to include the `--stripe-offset` style

## Checklist

## (Optional) Visual Changes

grafik
stripes not aligned, no stripes on spacer

grafik
stripes aligned, added stripes on spacer

grafik
full page view


Resaki1 (approved):
you're a wizard

Issue #3376, opened 2023-09-25
Endpoint for participated boards


SelinaBuff:
As a user i want to retrieve all boards i have participated from the Backend. 

![Image](https://github.com/inovex/scrumlr.io/assets/116653219/4129e2e8-e968-43ba-88d4-eee49bf8f83a)

PR #3472, opened 2023-10-16
feat: adds endpoint to receive all boards that the target user has been participating in

MarvinRoth:
## Description
Adds a endpoint on /boards/user that returns all the boards the provided user is participating in

## Changelog
-  Several formatting changes on changed files to satisfy linting conditions
- adding the getUserBoards funciont to server/src/api/boards.go
- adding the /boards/user route to server/src/api/router.go
- adding the database call to retreive the boards for the specific user in server/src/database/boards.go
- adding test for the added database function in server/src/database/boards_test.go
- adding the function to call the added server/src/database/boards.go function in server/src/services/boards/boards.go
- adding the GetUserBoards function to the server/src/services/services.go Board type

## Checklist

- ✅ I have performed a self-review of my own code
- ✅ I have commented my code, particularly in hard-to-understand areas
- ❌ The light- and dark-theme are both supported and tested
- ❌ The design was implemented and is responsive for all devices and screen sizes
- ❌ The application was tested in the most commonly used browsers (e.g. Chrome, Firefox, Safari)


Schwehn42 (dismissed):
could you also update the `postman_collection` to include the new endpoint? :)

server/src/database/boards.go
```diff
@@ -167,3 +168,17 @@ func (d *Database) GetBoard(id uuid.UUID) (Board, error) {
...
 }
+
+func (d *Database) GetUserBoards(userID uuid.UUID) ([]Board, error) {
+	var boards []Board
+	err := d.db.NewSelect().
+		TableExpr("boards AS b").
+		ColumnExpr("b.*").
+		Join("INNER JOIN board_sessions AS s ON s.board = b.id").
+		Where("s.user = ?", userID).
+		Scan(context.Background(), &boards)
```
Schwehn42, line 179: I think it would be nice if fetching the user sessions would already include all the required data in #3376, like the amount of columns and participants to avoid multiple BE calls, what do you think?

MarvinRoth: Sounds like a good idea, take a closer look on how to implement it.

MarvinRoth: > I think it would be nice if fetching the user sessions would already include all the required data in #3376, like the amount of columns and participants to avoid multiple BE calls, what do you think?

Would you rather have the column and participant data already counted or the full individual objects and count them in the frontend?

Schwehn42: By the look of the current design, we need the names of the columns as well, for participants I'm not sure. Maybe @Kraft16 knows more?

server/src/api/boards.go
```diff
@@ -60,6 +60,46 @@ func (s *Server) deleteBoard(w http.ResponseWriter, r *http.Request) {
...
+
+	boards, err := s.boards.GetUserBoards(r.Context(), user)
+	if err != nil {
+		log.Errorw("unable to get boards for this user", "err", err)
+		common.Throw(w, r, common.InternalServerError)
+		return
+	}
+	boardsWithDetails := make([]*dto.BoardWithDetails, len(boards))
+	for i, board := range boards {
+		board, _, sessions, cols, _, _, _, _, _, err := s.boards.FullBoard(r.Context(), board.ID)
```
brandstetterm, line 75: I have concerns about the performance here. The method fetches all data associated with a board (including all notes, votes, reactions, etc.), even if they are not needed. Perhaps a different, new method should be used here?

server/src/api/boards.go
```diff
@@ -60,6 +60,46 @@ func (s *Server) deleteBoard(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		for _, session := range sessions {
+			if session.User.ID == user {
+				sessionCreated := session.CreatedAt
+				boardsWithDetails[i] = &dto.BoardWithDetails{
+					Board:        board,
+					Participants: participantNum,
+					CreatedAt:    sessionCreated,
+					Columns:      cols,
```
brandstetterm, line 94: Do we need the details of the columns or is the number sufficient here as with the participants?

server/src/api/boards.go
```diff
@@ -60,6 +60,46 @@ func (s *Server) deleteBoard(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			log.Errorw("unable to get full board details", "err", err)
+			common.Throw(w, r, common.InternalServerError)
+			return
+		}
+		if err != nil {
+			log.Errorw("unable to get full board details", "err", err)
+			common.Throw(w, r, common.InternalServerError)
+			return
+		}
```
brandstetterm, line 86: I think one `if err != nil {...}` is sufficient ^^

server/src/api/boards.go
```diff
@@ -60,6 +60,46 @@ func (s *Server) deleteBoard(w http.ResponseWriter, r *http.Request) {
...
+		log.Errorw("unable to get boards for this user", "err", err)
+		common.Throw(w, r, common.InternalServerError)
+		return
+	}
+	boardsWithDetails := make([]*dto.BoardWithDetails, len(boards))
+	for i, board := range boards {
+		board, _, sessions, cols, _, _, _, _, _, err := s.boards.FullBoard(r.Context(), board.ID)
+		participantNum := len(sessions)
+		if err != nil {
+			log.Errorw("unable to get full board details", "err", err)
```
brandstetterm, line 78: This error message also seems a bit odd, because the full board details aren't requested (only some details for overview), so why should this error occur

server/src/api/router.go
```diff
@@ -171,7 +172,11 @@ func (s *Server) protectedRoutes(r chi.Router) {
 			s.initVotingResources(r)
 			s.initVoteResources(r)
 			s.initAssignmentResources(r)
-			s.initBoardReactionResources(r)
+      s.initBoardReactionResources(r)
+		})
+
+		r.Route("/boards/user", func(r chi.Router) {
+			r.Get("/", s.getUserBoards)
 		})
```
brandstetterm, line 180: Wouldn't `GET /boards` be a more suitable endpoint?

server/src/common/dto/boards.go
```diff
@@ -118,3 +119,10 @@ type BoardUpdateRequest struct {
 
 	ID uuid.UUID `json:"-"`
 }
+
+type BoardWithDetails struct {
```
brandstetterm, line 123: How about this? 
```suggestion
type BoardOverview struct {
```

server/src/database/boards.go
```diff
@@ -172,3 +173,17 @@ func (d *Database) GetBoard(id uuid.UUID) (Board, error) {
 	err := d.db.NewSelect().Model(&board).Where("id = ?", id).Scan(context.Background())
 	return board, err
 }
+
+func (d *Database) GetUserBoards(userID uuid.UUID) ([]Board, error) {
```
brandstetterm, line 177: Imo, the name should only be `GetBoards`, because with the user id as a parameter it is clear that these are the boards of a user

server/src/services/services.go
```diff
@@ -40,7 +40,9 @@ type Boards interface {
 	GetColumn(ctx context.Context, boardID, columnID uuid.UUID) (*dto.Column, error)
 	ListColumns(ctx context.Context, boardID uuid.UUID) ([]*dto.Column, error)
 
-	FullBoard(ctx context.Context, boardID uuid.UUID) (*dto.Board, []*dto.BoardSessionRequest, []*dto.BoardSession, []*dto.Column, []*dto.Note, []*dto.Reaction, []*dto.Voting, []*dto.Vote, error)
+  FullBoard(ctx context.Context, boardID uuid.UUID) (*dto.Board, []*dto.BoardSessionRequest, []*dto.BoardSession, []*dto.Column, []*dto.Note, []*dto.Reaction, []*dto.Voting, []*dto.Vote, error)
+	BoardOverview(ctx context.Context, boardID uuid.UUID) (*dto.Board, []*dto.BoardSession, []*dto.Column, error)
+	GetBoards(ctx context.Context, userID uuid.UUID) ([]*dto.Board, error)
```
brandstetterm, line 45: Consider returning only a list of board IDs with GetBoards instead of fetching the boards twice - once to get all boards of a user and a second time when you fetch the boards for the overview. This approach is common in API design and would improve the efficiency of the process.

server/src/database/board_overview.go
brandstetterm, line 1: Why was a new file created for this? Was there no more space in `database/boards.go`?

server/src/database/board_overview.go
```diff
@@ -0,0 +1,28 @@
...
+
+	sessions, err = d.GetBoardSessions(id)
+	if err != nil {
+		return Board{}, nil, nil, err
+	}
+
+	columns, err = d.GetColumns(id)
+	if err != nil {
+		return Board{}, nil, nil, err
+	}
```
brandstetterm, line 24: Why do you return `Board{}, nil, nil, err` in the error cases here? Shouldn't that be `nil, nil, nil, err`?

server/src/common/dto/board_sessions.go
```diff
@@ -48,7 +51,9 @@ func (b *BoardSession) From(session database.BoardSession) *BoardSession {
 	b.RaisedHand = session.RaisedHand
 	b.ShowHiddenColumns = session.ShowHiddenColumns
 	b.Role = session.Role
+	b.CreatedAt = session.CreatedAt
 	b.Banned = session.Banned
+	b.CreatedAt = session.CreatedAt
```
brandstetterm, line 56:
```suggestion
	b.CreatedAt = session.CreatedAt
	b.Banned = session.Banned
```

server/src/common/dto/board_sessions.go
```diff
@@ -32,6 +33,8 @@ type BoardSession struct {
 	// start voting sessions etc.
 	Role types.SessionRole `json:"role"`
 
+  //Reference for when board_session has been created
+	CreatedAt time.Time `json:"CreatedAt"`
```
brandstetterm, line 37:
```suggestion
	CreatedAt time.Time `json:"createdAt"`
```

server/src/database/board_overview.go
```diff
@@ -0,0 +1,28 @@
+package database
+
+import "github.com/google/uuid"
+
+func (d *Database) GetBoardOverview(id uuid.UUID) (Board, []BoardSession, []Column, error) {
```
brandstetterm, line 5: Why don't you return `BoardOverview` from the `common/dto` package here instead of every ingredient individually?

server/src/services/boards/boards.go
```diff
@@ -92,6 +105,16 @@ func (s *BoardService) FullBoard(ctx context.Context, boardID uuid.UUID) (*dto.B
 	return new(dto.Board).From(board), dto.BoardSessionRequests(requests), dto.BoardSessions(sessions), dto.Columns(columns), dto.Notes(notes), dto.Reactions(reactions), dto.Votings(votings, votes), personalVotes, err
 }
 
+func (s *BoardService) BoardOverview(_ context.Context, boardID uuid.UUID) (*dto.Board, []*dto.BoardSession, []*dto.Column, error) {
+	board, sessions, columns, err := s.database.GetBoardOverview(boardID)
+	sessions, err = s.database.GetBoardSessions(boardID)
```
brandstetterm, line 110: Why are you getting `session` twice here?

server/src/services/boards/boards.go
```diff
@@ -92,6 +105,16 @@ func (s *BoardService) FullBoard(ctx context.Context, boardID uuid.UUID) (*dto.B
 	return new(dto.Board).From(board), dto.BoardSessionRequests(requests), dto.BoardSessions(sessions), dto.Columns(columns), dto.Notes(notes), dto.Reactions(reactions), dto.Votings(votings, votes), personalVotes, err
 }
 
+func (s *BoardService) BoardOverview(_ context.Context, boardID uuid.UUID) (*dto.Board, []*dto.BoardSession, []*dto.Column, error) {
+	board, sessions, columns, err := s.database.GetBoardOverview(boardID)
```
brandstetterm, line 109: This `err` is unchecked

server/src/database/board_overview.go
MarvinRoth: That could theoretically be done, but we would break with the model that was implemented otherwise. We usually receive all the database objects, transform them into a data transfer objects inside the service package and then provide them through the api package. 
I see though, that the BoardOverview should probably be created inside the service package and not inside the api

brandstetterm (approved):

Issue #3478, opened 2023-10-16
Language Dropdown Broken


Resaki1:
### The problem

When clicking on the Language Dropdown inside the appearance settings, the dropdown is not visible. You have to scroll first to get to the dropdown.
This could probably be fixed by wrapping the dropdown in [react-doms `createPortal`](https://react.dev/reference/react-dom/createPortal#createportal)

### Browser

Firefox, Safari, and Arc (Chromium) for MacOS

### Steps to reproduce the behavior

- open appearance settings
- click on language dropdown

### Screenshots

https://github.com/inovex/scrumlr.io/assets/35272402/6b3b94c7-6978-4244-80e3-f7010bf38cc1

blxckOxymoron:
what is your opinion on using a builtin `<select>` element? 
this would handle overflow and accessibility for us
the drawback is that you can't really style the options but you can of course style the button

blxckOxymoron:
but I also found a way to get it back to work as it was before
I thought I had checked this but apparently I forgot to add the full height back in when adding the scrollbar

PR #3500, opened 2023-10-19
fix: extend height of appearance container

blxckOxymoron:
## Description

- extend the height of the apppearance settings container so that the language selector doesn't require scrolling when space is available (#3478)
- Improve UX by rotating the arrow when the dropdown is open

## Changelog

- add `flex-grow: 1;` to the appearance-container
- add `transform` and `transition` to the dropdown arrow
- scroll to options when the dropdown gets opened

## Checklist

## (Optional) Visual Changes

grafik
updated: no scrolling necessary when space is available


Schwehn42 (commented):
While this does make the selection option available, on some screen sizes (e.g. 420x705) it's still not apparent that the dropdown is visible (other than looking at the scrollbar).  
I would suggest maybe turning the little V arrow by 180deg to make it clear to the user something happened 😄

blxckOxymoron:
I agree, I'll also add a `.scrollIntoView()` call whenever the menu is opened.

Resaki1 (approved):

Issue #2909, opened 2023-04-26
Inline emojis in cards using :emoji_id: string


Schwehn42:
Inspired by Slack and others, there could be the option to add emojis in the note text using the :emoji_id: string template, e.g. `:rocket:` => 🚀  
This allows web users to add emojis anywhere.

Resaki1:
👐💎🚀

Schwehn42:
implementation: #3508

PR #3508, opened 2023-10-22
feat: emoji picker

blxckOxymoron:
> [!WARNING]  
> 🚧 WIP 
> I just wanted to make sure everyone knows this is being worked on so we don't work on this twice

## Description

- based on #2909
- add emojis with :emoji_name:
- get autocompletion for emoji names
  - keyboard & mouse support
- skin tone support

## Changelog

- created new `EmojiSuggestions` component & styles
- added a list of emoji names from [unicode-emoji-json](https://github.com/muan/unicode-emoji-json) which is based on [the unicode spec](http://www.unicode.org/Public/emoji/1.0/emoji-data.txt)
- WIP

## Checklist

## (Optional) Visual Changes

grafik
emoji picker


blxckOxymoron:
ready for review & feedback👍

src/constants/emojis.json
Schwehn42, line 1: maybe the json could be extended to also include [alternative keywords](https://unicode-org.github.io/emoji/emoji/format.html#col-annotations), additionally to the CLDR short name? that way, you wouldn't have to write the exact emoji name.  
For example, in Slack, I can write the following:
Bildschirm­foto 2023-11-20 um 14 03 07

The actual input is highlighted, but related emojis are also displayed.  
The keywords seem to vary by language though, so the question is whether supporting just the English version would be enough

blxckOxymoron: yes I'll look into that

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
```diff
@@ -81,19 +86,30 @@ export const NoteDialogNoteContent: FC<NoteDialogNoteContentProps> = ({noteId, a
...
+                e.preventDefault();
+                e.currentTarget.blur();
+              }
+            }}
+          />
+
+          {!isStackedNote && (
+            <div className="note-dialog__note-content--emoji-suggestions">
+              <EmojiSuggestions {...emoji.suggestionsProps} />
+            </div>
```
Resaki1, line 110: Why did you choose to make the emoji dropdown only available for the top note?

src/utils/hooks/useEmojiAutocomplete.ts
```diff
@@ -0,0 +1,215 @@
...
+ */
+export const useEmojiAutocomplete = <ContainerElement extends HTMLElement>(
+  {
+    maxInputLength = MAX_NOTE_LENGTH,
+    initialValue = "",
+    suggestionsHidden = false,
+  }: {
+    maxInputLength?: number;
+    initialValue?: string;
+    suggestionsHidden?: boolean;
```
Resaki1, line 41:
```suggestion
    maxInputLength: number;
    initialValue: string;
    suggestionsHidden: boolean;
```

Since you provide default values above, theses props are always provided. Or am I missing something?

src/utils/hooks/useEmojiAutocomplete.ts
```diff
@@ -0,0 +1,215 @@
...
+    setEmojiName(newEmojiName);
+  }, [value, cursor, emojiData, acceptSuggestion, skinToneComponent]);
+
+  // handle: update input, update suggestions
+  // FormEventHandler for TextareaAutosize
+  const handleChange: FormEventHandler<InputElement> & ChangeEventHandler<InputElement> = useCallback(
+    (e) => {
+      const newVal = e.currentTarget.value;
+      // prevent exceeding max input length by slicing the input
+      if (maxInputLength !== undefined && newVal.length > maxInputLength) setValue(newVal.slice(0, maxInputLength));
```
Resaki1, line 145:
```suggestion
      if (newVal.length > maxInputLength) setValue(newVal.slice(0, maxInputLength));
```

The check for undefined can be removed if you make the prop non-optional above

Resaki1 (changes requested):
Looks pretty nice, well done!

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
blxckOxymoron: It looking somewhat weird having transparent suggestions overlap when you scroll

grafik

src/utils/hooks/useEmojiAutocomplete.ts
blxckOxymoron: If I'd leave out the `?` it would require the properties to allways be specified/present in the object when calling the function. I added them so you can specify only what you need

```ts
// currently possible, but impossible with "?" removed
useEmojiAutocomplete({
  initialValue: "lorem ipsum"
})
```

src/utils/hooks/useEmojiAutocomplete.ts
blxckOxymoron: the props are allready infered to allways be set, I wanted to make maxInputLength able to be `undefined` so that you can disable the input limit. It looks like I forgot to add `undefined` to the type

Resaki1 (approved):

PR #3529, opened 2023-10-31
fix: reaction db migration

Schwehn42:
## Description

For older boards, the migration wouldn't work because setting the default value for the column `show_note_reactions` did not work as expected. This is fixed now.

## Changelog

- properly set default value of column `show_note_reactions` to `true`


Lennart01 (approved):
lgtm :shipit:

PR #3584, opened 2023-11-16
Fix: Two hotfixes

### LINKED ISSUE #3583: App crashes when opening stack view as participant (not owner/moderator)
Going into stack view as a participant (not as owner or moderator) leads to a complete crash. This is the error message:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'user')
    at NoteAuthorList.tsx:20:56
    at Array.map (<anonymous>)
    at NoteAuthorList.tsx:18:8
    at FL (NoteAuthorList.tsx:45:30)
    at _i (react-dom.production.min.js:167:137)
    at _u (react-dom.production.min.js:290:337)
    at bs (react-dom.production.min.js:280:389)
    at gs (react-dom.production.min.js:280:320)
    at Cs (react-dom.production.min.js:280:180)
    at is (react-dom.production.min.js:271:88)
```

The issue comes from the non-null assertion of the variable `a` in line 20 of `NoteAuthorList.tsx`:
`const displayName = isSelf ? t("Note.me") : a!.user.name;`

Resaki1:
Fixes two bugs:
- When attempting to view a note in stack view as a participant (not owner/moderator) while the option to display the note authors is disabled, the app crashes (as described in #3583). This has been fixed by not attempting to render the note author list anymore when this board setting is disabled
- When changing a board setting that affects the notes while no note is present on the board, a runtime error is produced. This has been fixed by passing `null` to the redux dispatch when no notes exist

Closes #3583

### DIFF
--- a/src/components/Column/Column.tsx
+++ b/src/components/Column/Column.tsx
@@ -36,9 +36,11 @@ export const Column = ({id, name, color, visible, index}: ColumnProps) => {
   const notes = useAppSelector(
     (state) =>
       state.notes
-        .filter((note) => !note.position.stack)
-        .filter((note) => (state.board.data?.showNotesOfOtherUsers || state.auth.user!.id === note.author) && note.position.column === id)
-        .map((note) => note.id),
+        ? state.notes
+            .filter((note) => !note.position.stack)
+            .filter((note) => (state.board.data?.showNotesOfOtherUsers || state.auth.user!.id === note.author) && note.position.column === id)
+            .map((note) => note.id)
+        : [],
     _.isEqual
   );
   const moderating = useAppSelector((state) => state.view.moderating, _.isEqual);
--- a/src/components/Note/Note.tsx
+++ b/src/components/Note/Note.tsx
@@ -101,9 +101,7 @@ export const Note = (props: NoteProps) => {
     >
       <button className={`note note--${stackSetting}`} onClick={handleClick} onKeyDown={handleKeyPress} ref={noteRef}>
         <header className="note__header">
-          <div className="note__author-container">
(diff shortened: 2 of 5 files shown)


Schwehn42 (commented):
Thanks for fixing the two issues so fast :)
Just one complaint: it's _still_ supposed to show your own avatar on your own notes, even with the option disabled!  
See the following code from `NoteAuthorList`:

https://github.com/inovex/scrumlr.io/blob/3049c8b349c39702df69e22662249b3d23623383/src/components/Note/NoteAuthorList/NoteAuthorList.tsx?plain=1#L37-L42

If this code didn't exist, `props.showAuthors` would be obsolete.

If you want, I can still approve if an issue is created for this :eyes:

Resaki1:
> Thanks for fixing the two issues so fast :) Just one complaint: it's _still_ supposed to show your own avatar on your own notes, even with the option disabled! See the following code from `NoteAuthorList`:
> 
> https://github.com/inovex/scrumlr.io/blob/3049c8b349c39702df69e22662249b3d23623383/src/components/Note/NoteAuthorList/NoteAuthorList.tsx?plain=1#L37-L42
> 
> If this code didn't exist, `props.showAuthors` would be obsolete.
> 
> If you want, I can still approve if an issue is created for this 👀

Should be fixed now, thanks for pointing that out.
I also fixed a bug where, when showAuthors is set to false, every note would show the current user as the note author even if that was wrong (again, only for participants and not owners/mods)

Schwehn42 (approved):
> I also fixed a bug where, when showAuthors is set to false, every note would show the current user as the note author even if that was wrong (again, only for participants and not owners/mods)

Interesting, so you basically got three birds with one stone with this PR ^^

PR #3625, opened 2023-11-30
fix: Hotkeys

### LINKED ISSUE #2762: Fix hotkey cheat sheet
**User Stories:**

> 1. As a user, when I press "SHIFT + /" on my English keyboard layout, the hotkey toggle should be triggered to open the hotkey manual, as specified in the cheatsheet. This functionality needs to be fixed.
> 
> 2. As a user, when I press "CMD + U," I expect to see the notes of other users. Currently, only "CTRL + U" is working, which is not the intended behavior.
> 
> 3. As a user, when I press "CMD + J," I expect to be able to show hidden columns. Currently, only "CTRL + J" is working, which is not the intended behavior.
> 
> 4. As a user, when I trigger the settings dialog with the "S" key, it should also close the dialog if it's already open.
> 
> 5. As a user, when I trigger the voting dialog with the "V" key, it should also close the dialog if it's already open.
> 
> 6. As a user, when I trigger the timer dialog with the "T" key, it should also close the dialog if it's already open.
> 
> Additionally, after issue #3357 is merged, the following functionality should be added:
> 
> 7. As a user, when I press the "R" key, it should trigger a board reaction.

Resaki1:
## Description
Fixed some shortcuts, especially for english keyboard layouts. Changed hotkey behavior for dialogs to toggle. Also updated the hotkey cheat sheet pdf. Closes #2762.

## Changelog
- changed some hotkeys to also allow usage on english keyboards
- changed some hotkeys to use the `meta` key instead of OS-specific keys
- changed dialog hotkeys to toggle the dialog instead of only opening it
- updated the hotkey to show the toggle option and the new board reactions


Resaki1:
> is the pdf correct with `cmd ⌘` = `Win ⊞`? I thought the equivalent is `Ctrl`.

It used to be wrong, but now it is correct. I changed the `cmd` key in the hotkeys to `meta`, which corresponds to the `cmd` key on MacOS and the `Win` key on Windows

brandstetterm (approved):
LGTM

PR #3740, opened 2024-01-12
feature: increment timer by one minute

brandstetterm:
The timer pill should soon have a button to increment the ongoing timer by one minute. Therefore, I've implemented the backend changes required for that.


MarvinRoth (approved):
should work as intended

PR #3745, opened 2024-01-12
feature: Option to lock board

BenedictHomuth:
## Description

Added the possibility to make a board (notes, reactions, votes) uneditable. This might be useful when you want to persist a current stage of a board.

When the option is active only moderators are able to edit the board further. Participants can only "view" the state.

## Changelog

- Added a new context for backend API calls, which filters out, if edits are currently allowed
   - When you are a moderator / owner
   - You are a participant and the board is currently open for edits
- Updated frontend checks for when a user is able to edit

## Checklist


server/src/api/router.go
```diff
@@ -184,10 +184,11 @@ func (s *Server) protectedRoutes(r chi.Router) {
 
 func (s *Server) initVoteResources(r chi.Router) {
 	r.Route("/votes", func(r chi.Router) {
-		r.Use(s.BoardParticipantContext)
+		r.Use(s.BoardEditableContext)
+
 		r.Post("/", s.addVote)
 		r.Delete("/", s.removeVote)
-		r.Get("/", s.getVotes)
+		r.With(s.BoardParticipantContext).Get("/", s.getVotes)
```
brandstetterm, line 191: This solution isn't really readable in my opinion. The BoardParticipantContext basically tells me that for a specific endpoint, the requestor has to be a participant of the board. Now with this change I'm confused because it is not immediately recognisable to me that all endpoints require the participation of the board since only getVotes uses the BoardParticipantContext.

Maybe something like that would work:
```go
func (s *Server) initVoteResources(r chi.Router) {
	r.Route("/votes", func(r chi.Router) {
    r.Use(s.BoardParticipantContext)

		r.Get("/", s.getVotes)

    r.Group(func(r chi.Router) {
      r.Use(s.BoardEditableContext)

      r.Post("/", s.addVote)
      r.Delete("/", s.removeVote)
    })
	})
}

```

server/src/api/context.go
```diff
@@ -1,131 +1,183 @@
...
+		if err != nil {
+			log.Errorw("unable to check board session", "err", err)
+			common.Throw(w, r, common.InternalServerError)
+			return
+		}
+
+		if !exists {
+			common.Throw(w, r, common.ForbiddenError(errors.New("user board session not found")))
+			return
+		}
```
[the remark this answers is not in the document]
BenedictHomuth: I changed it. Now the board and user is read from the then already present context.

server/src/api/router.go
BenedictHomuth: You are correct! Changed it to your example.

brandstetterm (approved):
lgtm

Issue #3720, opened 2024-01-10
If the first or last column is hidden and the presenter mode is active, the hatching colour is that of the presenter mode.


brandstetterm:
If the first or last column is hidden, the columns and the edge of the board is hatched. If the presentation mode is also active, the hatching at the edge of the board has the color of presenter mode indicator. Instead, the hatching at the edge of the board should have the same color as the hidden column.

![Image](https://github.com/inovex/scrumlr.io/assets/36969812/cec8fffd-d871-40bd-88e6-f8370ddb0a93)

PR #3755, opened 2024-01-15
fix: only apply lime color to the spacer border in presenter mode

blxckOxymoron:
## Description

- #3720
- The presenter mode no longer affects the stripe background of the spacer

## Changelog

- instead of changing the whole theme color of the spacer, only update the border color
- added a new styles for `.board__spacer--moderation-isActive` to change the border color to green
- removed changing the accent color to green, instead apply the `.board__spacer--moderation-isActive` class conditionally

## Checklist

## (Optional) Visual Changes

grafik
updated, correct colors

### DIFF
--- a/src/components/Board/Board.scss
+++ b/src/components/Board/Board.scss
@@ -39,9 +39,17 @@
 
 .board__spacer-left {
   @include inset-border($top: true, $left: true, $bottom: true);
+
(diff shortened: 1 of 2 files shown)


brandstetterm (approved):
LGTM! Thank you for your contribution 🥳

PR #3770, opened 2024-01-19
feature: ban participants – backend

BenedictHomuth:
## Description

This PR enhances the UpdateBoardSession with a field that indicates if a participant is banned. When that is the case, the participant is not able to join the board.

## Changelog

Created a new context, since we need to return an error while on the joinBoard route, so the frontend can catch it and do not allow for a banned participant to enter.

Created a new error for when a participant is banned (403 – Gone). I choose this one since 403 - Forbidden is already in use, when the participant enters the wrong passphrase.

Enhanced the updateBoardSession Route, with the new banned field.

Added test to verify that the updated banned field works.

## Checklist


server/src/api/context.go
```diff
@@ -53,7 +52,35 @@ func (s *Server) BoardParticipantContext(next http.Handler) http.Handler {
...
 		user := r.Context().Value("User").(uuid.UUID)
+		banned, err := s.sessions.ParticipantBanned(r.Context(), board, user)
+		if err != nil {
+			log.Errorw("unable to check if participant is banned", "err", err)
+			common.Throw(w, r, common.InternalServerError)
+			return
+		}
+
+		if banned {
+			common.Throw(w, r, common.GoneError(errors.New("participant is currently banned from this session")))
```
MarvinRoth, line 63: Im not sure if GoneError is the most suitable for a banning error. Maybe it should be something like access denied.
Gone would indicate that a ressource is gone for good and not just that the access to the ressource is not granted

server/src/common/error.go
```diff
@@ -40,6 +41,15 @@ func ForbiddenError(err error) *APIError {
 	}
 }
 
+func GoneError(err error) *APIError {
```
MarvinRoth, line 44: just as mentioned in server/src/api/context.go, maybe not the right error for the job

MarvinRoth (commented):
Maybe reconsider the error type. The business logic seems to work as intended

server/src/api/context.go
brandstetterm: But the 403 Forbidden is better than 401 Unauthorized. 
401 Unauthorized is specifically for the case that the client is unauthenticated, which is not the case here.

brandstetterm (approved):
lgtm 👍🏼

server/src/api/context.go
BenedictHomuth: 403 won in the end! :)

PR #3779, opened 2024-01-23
chore: remove assignments code 

brandstetterm:
## Description
Since this feature isn't planned anymore, I've removed the already existing code from our front- and backend to declutter the codebase.

## Changelog
- Remove front- and backend code related to assignments


server/src/server
BenedictHomuth, line 1: This should be removed, since it seems like a locally generated debug binary. The rest looks good to me.

brandstetterm: I've removed the binary @BenedictHomuth

BenedictHomuth (approved):

PR #3865, opened 2024-02-14
fix: prevent gif overflow

### LINKED ISSUE #3846: GIFs overflow the note component in the stack view
As a user, I want embedded GIFs to respect the boundaries of a note in the stack view and not overflow it.
Image

CronJorian:
## Description

Prevent gif overflow and add rounded corners for the board view.
Closes #3846.

## Changelog

`min-height` and `min-width` properties are set to 100%.
This may be set to 0 instead, but was set to 100% to ensure the content grid area always uses the whole space.
More info can be found here: https://stackoverflow.com/questions/43311943/prevent-content-from-expanding-grid-items

## Checklist

## (Optional) Visual Changes

image
image
Focus is still working as intended:
image

Meow meow

Instead of this:
image
It looks like this:
image

### DIFF
--- a/src/components/Note/Note.scss
+++ b/src/components/Note/Note.scss
@@ -147,6 +147,7 @@ $note__footer-gap: 42px;
   position: absolute;
   top: 0;
   left: 0;
+  border-radius: $note__border-radius;
 }
 
 .note__stack-icon {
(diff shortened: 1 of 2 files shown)


brandstetterm (approved):
lgtm

PR #3875, opened 2024-02-16
fix: increase google avatar_url

Lennart01:
## Description
Increases the maximum length of the Google avatar URL, as it was causing login issues when the avatar URL was too long.


server/src/database/migrations/sql/13_increase_gauth_avatar_url.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE IF EXISTS google_users ALTER COLUMN avatar_url TYPE varchar(2048);
```
CronJorian, line 1: Can you add a newline at the end? :)

CronJorian (approved):
It'd be reasonable to check whether the other authentications also require a longer url.
Otherwise lgtm 🚀

server/src/database/migrations/sql/13_increase_gauth_avatar_url.down.sql
brandstetterm, line 1: I actually don't know if that works. What happens if there are urls with a length greater than 256?

Lennart01: I guess that Postgres would refuse the operation.
Technically we would need to remove the avatar URL before migrating back or not migrating back at all.

brandstetterm (approved):

PR #3901, opened 2024-02-26
fix: board header menu overflow

brandstetterm:
## Description
I've noticed that the edit button of the board name in the header menu has been pushed outside the container. The problem is that flexbox set the `min-width` of the input element to auto and therefore it couldn't be smaller than it's content, pushing the edit button out of the container. 

## Changelog
- Set the `min-width` of the input element to 0 

## (Optional) Visual Changes
Before:
image

After:
image


src/components/BoardHeader/HeaderMenu/BoardSettings/BoardSettings.scss
```diff
@@ -50,6 +51,7 @@ $header-menu-button-width: 24px;
   padding: 0;
   transition: border-color 0.1s linear;
   caret-color: $color-white;
+  min-width: 0;
```
CronJorian, line 54: Omitting this doesn't seem to change anything, even with an empty board name.
Is it required?

CronJorian (approved):
Just a question about the `min-width`.
Otherwise good.

Additionally, I found that the input of `.board-settings__board-name` has no `maxlength` limit.
Should I add an issue for it or is it already planned for the board menu reimplementation?

src/components/BoardHeader/HeaderMenu/BoardSettings/BoardSettings.scss
brandstetterm: For me it's only Firefox that's making problems here

PR #3949, opened 2024-03-07
fix: change http response of passphrase request

### LINKED ISSUE #3920: Better HTTP status codes for board join requests
Currently, the backend returns the 403 code for multiple occasions:
- User tries to connect to a protected board the first time (without a password)
- User tries to connect to a protected board with the wrong password
- User is banned from a session

We should have different status codes for these cases, so that we can handle them better in the frontend. Right now, to check if a user is banned, we need to match the whole error string, which is not ideal imo.

mateo-ivc:
## Description

Change of http status codes when sending a join request with an incorrect passphrase to a protected board for better differentiation during error evaluation.

Closes #3920 
 
## Changelog

In 'api/boards.go': add check for empty passphrase and change of status code from ForbiddenRequest to BadRequest 
In 'api/participant.ts': Added status codes to the return value when a 400 request is received when trying to join a board

## Checklist

### DIFF
--- a/server/src/api/boards.go
+++ b/server/src/api/boards.go
@@ -159,11 +159,11 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
 		var body JoinBoardRequest
 		err := render.Decode(r, &body)
 		if err != nil {
-			http.Error(w, "unable to parse request body", http.StatusBadRequest)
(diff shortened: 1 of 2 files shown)


brandstetterm (approved):
LGTM

PR #3955, opened 2024-03-08
feat: lock board

### LINKED ISSUE #3538: Add option to lock board
### Discussed in https://github.com/inovex/scrumlr.io/discussions/3030

<sup>Originally posted by **brandstetterm** May 30, 2023</sup>
I think it could be cool if a moderator could "lock" a board so that no more notes can be added / notes can't be edited anymore. </div>

**User Stories:**

> 1. As a moderator, I want the ability to lock a board so that no more notes can be added, in order to prevent further contributions and maintain the current state of the board.

> 2. As a moderator, I expect a simple and intuitive interface to lock and unlock the screen. (I can imagine this being included as an icon (lock) on the right side of the moderator menu.)

> 3. As a user, I want to receive visual feedback when the screen is locked. It should be clear that no further changes can be made.

**You can find the Design on Figma. Go on the left menu on "🧩  Components" and then to the section "Lock board":**

brandstetterm:
## Description
Frontend changes related to the locking of a board. If a board is locked, participants (without moderation permissions) are no longer able to create, update and delete notes, dispatch board reactions, add or remove reactions to a note, add or remove votes to a note. Therefore, there have been some changes in the frontend (e.g. gray out buttons, hide buttons, etc.). A detailed list of the changes can be found below.

Closes #3538 

## Changelog
- NoteInput: Disable textarea, add lock icon, gray out plus icon
- StackView: Disable textarea to edit note content, hide delete and unstack buttons
- Action Bars: Disable button to dispatch board reactions, automatically close board reactions menu if board gets locked
- Notes: Hide button to open reactions dropdown, disable button to remove note reactions, hide button to add a vote and disable button to remove a vote
- Update several tests and snapshots

## Future tasks
- Implement notifications manager and dispatch notification that board is locked on join

## Visual changes
Screenshot 2024-03-18 at 17 31 49
Screenshot 2024-03-18 at 17 32 05


src/components/NoteInput/NoteInput.scss
```diff
@@ -14,19 +16,25 @@ $note-input__input-right: 40px;
...
+  &:not(:has(.note-input__input:disabled)):hover,
+  &:not(:has(.note-input__input:disabled)):focus-within {
     border-color: rgba(var(--accent-color-rgb), 0.5);
     box-shadow: 0 6px 9px 0 rgba(var(--accent-color-rgb), 0.16);
   }
 }
 
+.note-input__lock-icon {
+  width: 32px;
+  height: 32px;
```
Schwehn42, line 30: hard-coded val

src/components/MenuBars/MenuBars.tsx
Schwehn42, line 1: the board reaction menu can still be accessed using the hotkey, and sending a reaction will result in an error

src/components/Note/NoteReactionList/NoteReactionList.tsx
```diff
@@ -45,6 +45,9 @@ export const NoteReactionList = (props: NoteReactionListProps) => {
   const others = useAppSelector((state) => state.participants?.others) ?? [];
   const participants = [me, ...others];
 
+  const isModerator = useAppSelector((state) => ["OWNER", "MODERATOR"].some((role) => state.participants!.self.role === role));
+  const boardLocked = useAppSelector((state) => !state.board.data!.allowEditing);
```
Schwehn42, line 49: I'm not sure about negating the state for every use – would it not make more sense to just rename the `allowEditing` property to `boardLocked` or `isBoardLocked`?

Schwehn42:
Added a blocked label, because #4104 should be merged before this

brandstetterm:
Thanks for your review @Schwehn42 
Since I'm limited in time right now, you can commit the suggested changes yourself if you want 😄

src/components/MenuBars/MenuBars.tsx
Schwehn42: I think it should be discussed whether locking the board should disallow board reactions, since these don't change the state of the board.

src/components/Note/NoteReactionList/NoteReactionList.tsx
Schwehn42: This would of course require a backend change as well, so this should be talked about beforehand.

src/components/MenuBars/MenuBars.tsx
Schwehn42: changed with #4144

src/components/Note/NoteReactionList/NoteReactionList.tsx
Schwehn42: see #4145

Schwehn42:
added feats: #4144 and #4145, ready to review now

BenedictHomuth:
Should the feature of locking a board also be an option, when creating a board? 
Meaning that it should be possible to set the "isLocked" field in the API request when creating a board?

If so I would add it. I do not remember if this was requested or not when we formulated this feature.

Schwehn42:
@BenedictHomuth 
> Should the feature of locking a board also be an option, when creating a board?

This isn't an AC in #3538, so I would hold on implementing it, since the create board screen will be overhauled soon anyways.

laila-rin (approved):
looks good to me, works well.

Would it make sense to have a little lock icon for the moderator also, so they know if the board is locked or not? Atm it's just the non-moderator participants that have a lock icon to indicate that, right?

Schwehn42:
@laila-rin Good point, I actually like the initial idea from the issue (#3538) where instead of toggle in board settings, it would be an button on the right where the rest of the moderator tools (start vote, start timer ...) are. That way, moderators could always see whether it's active, and also toggle it with one click instead of two.

What do you think about this? Also, does it make more sense to just use this version and make these changes in a new PR, if @SelinaBuff approves of the idea?

laila-rin (approved):

PR #3984, opened 2024-03-18
fix: asking for passphrase when joining a public board

brandstetterm:
## Description
There has been a bug where the server returned `403 Forbidden` with the error `user board session not found` when a user tried to join a public board the first time. 

## Changelog
- Use the `BoardParticipantContext` for every board route individually instead of the whole group


mateo-ivc (approved):
lgtm

PR #4001, opened 2024-03-25
refactor: dto – NoteDeleteRequest

BenedictHomuth:
## Description

This PR extends the dto – NoteDeleteRequest.

It is now closer aligned in use with other dto´s serving similar purposes.

## Changelog

added fields to dto NoteDeleteRequest
created NoteDelete struct in DB package for easier handling
tests for database and observer adjusted for the change

## Checklist

### DIFF
--- a/server/src/common/dto/notes.go
+++ b/server/src/common/dto/notes.go
@@ -76,7 +76,6 @@ type NoteCreateRequest struct {
 
 // NoteUpdateRequest represents the request to update a note.
 type NoteUpdateRequest struct {
-
 	// The text of the note.
 	Text *string `json:"text"`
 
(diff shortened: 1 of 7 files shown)


MarvinRoth (approved):
good start towards getting the backend into a more uniform state.
we need to go through the whole codebase to make it more uniform across the board though

server/src/common/dto/notes.go
```diff
@@ -89,7 +88,9 @@ type NoteUpdateRequest struct {
 
 // NoteDeleteRequest represents the request to delete a note.
 type NoteDeleteRequest struct {
+	ID uuid.UUID `json:"-"`
```
brandstetterm, line 91: Since the NoteDeleteRequest struct is a representation of the request coming from the frontend, isn't it missleading that you've included the note- & board-id, even tho they are extracted from the URL and are not contained within the request body?

PR #4005, opened 2024-03-26
fix: Check user permission when board is locked

mateo-ivc:
## Description

Changed router.go that check if the user has sufficient rights for this request.
Change to notes_test.go so that the boardID is not sent via the context
## Changelog

In 'context.go' removed unnecessary check when boardID is sent via context, removed checks which are already covered by AuthContext
In 'notes_test.go' sending boardID now as url parameter
In 'router.go' added middelware calls to check user premissions

## Checklist

### DIFF
--- a/server/src/api/context.go
+++ b/server/src/api/context.go
@@ -49,13 +49,6 @@ func (s *Server) BoardParticipantContext(next http.Handler) http.Handler {
 			return
 		}
 
-		userValue := r.Context().Value("User")
-		if userValue == nil {
-			log.Error("User value not found in the context")
-			common.Throw(w, r, common.BadRequestError(errors.New("user not found in the context")))
-			return
-		}
-
 		user := r.Context().Value("User").(uuid.UUID)
 		exists, err := s.sessions.SessionExists(r.Context(), board, user)
 		if err != nil {
--- a/server/src/api/context.go
+++ b/server/src/api/context.go
@@ -43,15 +43,34 @@ func (s *Server) BoardCandidateContext(next http.Handler) http.Handler {
 func (s *Server) BoardParticipantContext(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		log := logger.FromRequest(r)
-
 		boardParam := chi.URLParam(r, "id")
+
 		board, err := uuid.Parse(boardParam)
 		if err != nil {
 			common.Throw(w, r, common.BadRequestError(errors.New("invalid board id")))
 			return
 		}
 
-		user := r.Context().Value("User").(uuid.UUID)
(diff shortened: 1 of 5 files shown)


server/src/api/notes_test.go
```diff
@@ -151,5 +247,83 @@ func (suite *NotesTestSuite) TestGetNote() {
...
+					AllowEditing: tt.allowEditing,
+				}, tt.err)
+				noteMock.On("Delete", mock.Anything, mock.Anything).Return(tt.err)
+			}
+
+			req := NewTestRequestBuilder("DELETE", fmt.Sprintf("/notes/%s", noteID.String()), strings.NewReader(`{"deleteStack": false}`))
+			rctx := chi.NewRouteContext()
+			rctx.URLParams.Add("id", boardID.String())
+			req.AddToContext(chi.RouteCtxKey, rctx)
+			req.AddToContext("User", userID)
```
mateo-ivc, line 318: Is there maybe a better way to create the HTTP Request?

server/src/api/context.go
```diff
@@ -43,15 +43,34 @@ func (s *Server) BoardCandidateContext(next http.Handler) http.Handler {
...
 			return
 		}
 
-		user := r.Context().Value("User").(uuid.UUID)
+		userValue := r.Context().Value("User")
+		if userValue == nil {
+			log.Error("User value not found in the context")
+			common.Throw(w, r, common.BadRequestError(errors.New("user not found in the context")))
+			return
+		}
```
brandstetterm, line 59: I don't think it's possible for the userValue to be nil since the preceding AuthContext is checking it

server/src/api/context.go
```diff
@@ -43,15 +43,34 @@ func (s *Server) BoardCandidateContext(next http.Handler) http.Handler {
...
+		if !ok {
+			log.Error("Invalid user type in the context")
+			common.Throw(w, r, common.BadRequestError(errors.New("invalid user type in the context")))
+			return
+		}
+		if s.sessions == nil {
+			log.Error("s.sessions is nil")
+			common.Throw(w, r, common.InternalServerError)
+			return
+		}
```
brandstetterm, line 71: Since AuthContext is doing `uuid.Parse(userID)` and throws an error if something goes wrong, I'd expect it to be a uuid and therefore this should never throw an error.

mateo-ivc: Since BoardParticipantContext is only called in protected routes and AuthContext is also used for all protected routes,  this check is obsolete so we can remove it

server/src/api/context.go
mateo-ivc: True, AuthContext is actually the methode that writes the user id into the context

brandstetterm (approved):

PR #4051, opened 2024-04-08
chore: add golangci-lint to github workflow

mateo-ivc:
## Description

Setting up a github action for golanci-lint.
Also refactoring the code according to the linter
## Changelog

'golangci-lint.yaml': Basic configuration for the github action.
'/identifiers/identifiers.go': Since the docs: [context.WithValue](https://pkg.go.dev/context#WithValue) do not recommed to use built-in types as keys. Added structs which are used to extract and map values stored in context.Context 
Also changed removed string keys in context.Value, context.WithValue, ... and replaced them with the structs.
'api/board_session_requests_listen_on_request.go': Substitute infinity loop for for-loop that waits for a channel input
'feedback.go': returns now an error which is beeing handled
 

brandstetterm:
> '/identifiers/identifiers.go': added structs which are used to extract and map values stored in context.Context

Please explain further. I don't understand completely why does where added.

mateo-ivc:
As you can see in the documents of [context.WithValue](https://pkg.go.dev/context#WithValue). The provided key should not be of type String or any other built-in type to avoid collisions between packages. For this reason, we need to implement our own types for the keys.

server/src/api/board_reaction.go
```diff
@@ -6,12 +6,13 @@ import (
...
 	"scrumlr.io/server/common"
 	"scrumlr.io/server/common/dto"
+	"scrumlr.io/server/identifiers"
 )
 
 // createBoardReaction creates a new board reaction
 func (s *Server) createBoardReaction(w http.ResponseWriter, r *http.Request) {
-	board := r.Context().Value("Board").(uuid.UUID)
-	user := r.Context().Value("User").(uuid.UUID)
+	board := r.Context().Value(identifiers.KeyBoardIdentifier).(uuid.UUID)
```
brandstetterm, line 14: I'd go with sth like `BoardIdentifier` or `BoardKey` (KeyBoard always confuses me with keyboard 😄 )

brandstetterm (approved):

PR #4086, opened 2024-04-17
feat: documentation using github pages

Lennart01:
## Checklist


brandstetterm:
Tbh I would prefer this to be in a separate GitHub repository  :D

Schwehn42:
> Tbh I would prefer this to be in a separate GitHub repository :D

why?

Schwehn42:
I prefer it in here, if it's in another repo I can already see information not being up to date because it was forgotten. The main repository is the most accessible not only for us devs, but also the first place to look for anyone who actually needs them. Furthermore once this is integrated I'm assuming the information won't change that frequently so I don't believe it's going to get in our way.

brandstetterm:
> > Tbh I would prefer this to be in a separate GitHub repository :D
> 
> why?

Because the codebase & commit history of the main application should be kept clean in my opinion. For the same reason, the landing page is a separate project.

Schwehn42:
> Also, as far as I've seen, the GitHub action to deploy the documentation to GitHub pages is planned to run on every commit on main. With 99% of all commits probably not being related to the documentation, that's unnecessary.

Since a commit to main equals a merged PR, I think it would be interesting to know if the whole documentation changes even though something completely unrelated was commited or only the relevant parts are affected. Because in the first case, I completely agree with as this would clutter the history, but if it's a smart tool I think it would be nice to have the most up to date documentation within the same repo.

Schwehn42:
That being said I don't insist on this, so if you really say this is better for the reasons described then I don't mind as well

Lennart01:
> > > Tbh I would prefer this to be in a separate GitHub repository :D
> > 
> > 
> > why?
> 
> Because the codebase & commit history of the main application should be kept clean in my opinion. For the same reason, the landing page is a separate project.
> 
> Also, as far as I've seen, the GitHub action to deploy the documentation to GitHub pages is planned to run on every commit on main. With 99% of all commits probably not being related to the documentation, that's unnecessary.

Which is something we can tackle.
We can only run the docs pipeline if we made changes to the folder like (this)[https://www.cbui.dev/how-to-only-run-github-actions-steps-if-files-change/#:~:text=If%20you%20want%20to%20conditionally,only%20operates%20on%20entire%20workflows.]

Schwehn42 (approved):

PR #4088, opened 2024-04-18
fix: wait for dev cluster to be healthy before octomind

Lennart01:
## Checklist


.github/workflows/continuous-integration.yml
```diff
@@ -84,6 +84,35 @@ jobs:
     needs: deploy_to_dev_cluster
     runs-on: ubuntu-latest
     steps:
+      - name: Chance DNS
```
brandstetterm, line 87: I think I would prefer a better name for the step. I don't think I could imagine what is meant by "Chance DNS" when I read it, for example if it's failing. How about "Await DNS"?

Lennart01: well thats what it does. change the dns server, Maybe change dns server?

brandstetterm: Sounds fine :)

brandstetterm (approved):

PR #4104, opened 2024-04-27
improvement: locked board notification

### LINKED ISSUE #4081: Notify participants on join if a board is locked
## Description

With #3538, it has been implemented that moderators could locked a board for participants to prevent further changes. As a next step, the participants who join a board should be notified that a board is currently locked and no changes are possible.

Schwehn42:
## Description

Notifies joining users if the board is currently locked.
Closes #4081. 

## Changelog

- get locked prop from state
- send toast if true
- add localization

## Checklist

## TBD
is there a more sensible location for the code? i just put it somewhere because we don't really have a defined way to handle this

### DIFF
--- a/src/i18n/en/translation.json
+++ b/src/i18n/en/translation.json
@@ -287,7 +287,7 @@
     "moderatorResetReadyStatesButton": "Reset!",
     "bannedParticipant": "Participant '{{user}}' was removed",
     "unbannedParticipant": "The ban for participant '{{user}}' was revoked",
-    "lockedBoard": "The board is currently locked; you can't make changes to it.",
+    "lockedBoard": "The board is currently locked; only moderators can make changes to it.",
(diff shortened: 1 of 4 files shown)


src/i18n/en/translation.json
```diff
@@ -287,6 +287,7 @@
     "moderatorResetReadyStatesButton": "Reset!",
     "bannedParticipant": "Participant '{{user}}' was removed",
     "unbannedParticipant": "The ban for participant '{{user}}' was revoked",
+    "lockedBoard": "The board is currently locked; you can't make changes to it.",
```
brandstetterm, line 290: Maybe there should be a different toast message for participants and moderators since moderators can still make changes

Schwehn42: I changed it to something more general so only one text is required

brandstetterm (approved):

PR #4113, opened 2024-04-30
improvement: Split the messaging channel between user and moderator/owner

mateo-ivc:
## Description

Splitted up the board subscription into two channels. Prior every user subscribed to `boards.<boardID>` now based on the role each user connects to `boards.<boardID>.participant` or `boards.<boardID>.moderator`.
Why? To avoid sending events to participants they don't have access to.

## Changelog

`api/boards_listen_on_board.go`
- clients and subscriptions are now stored in a map based on their roles
- running each channel in it's own goroutine
- added method to change channels when participants role gets updated.
- added mutex to BoardSubscription struct to avoid concurrent access on the websockets.

`event_filter.go`
- change the method signature to b *Boardsubscription to be uniform
- added if clause for realtime.BoardEventParticipantUpdated to cache updated user

`board_sessions_observer.go`
- columns and notes included in the parameters of UpdateSession -> must be passed on in order to be able to send Column_Update and Notes_Sync events.
- added db calls to get notes and columns from board

`database/columns.go`
- storing the column to be deleted in context to check later if this column_delete event should be send to all participant or only mods/owner

`database/columns_observer.go`
- added the column to be deleted to the params of DeleteColumn to later check if this column is hidden or not

`database/notes.go`
- storing the note to be deleted in the context to later check if this note is in a hidden column or not

`database/notes_observer.go`
- added note and columns as parameters to DeletedNote call

`realtime/boards.go`
- Added SessionChannel type for the two channels
- Edited BroadcastToBoard() to be able to broadcast to multiple channels
-  BroadcastToBoard() and GetBoardChannel() now now subscribe/listen to `boards.<boardID>.moderator/participant` 

`services/...`
- Fixed all services by adding channels when calling realtime.BroadcastToBoard
- in some cases the event should only be sent to the moderators -> DeletedColumns, DeletedNote
in order to send the event to the correct channel, it's necessary to check if the column is hidden or not.

`services/boards/sessions.go`
- UpdatedSession -> added two more broadcasts that are needed when a participants role gets updated to display hidden columns and their notes.

## Checklist


brandstetterm:
> Splitted up the board subscription into two channels. Prior every user subscribed to `boards.<boardID>` now based on the role each user connects to `boards.<boardID>.participant` or `boards.<boardID>.moderator`.
> Why? To avoid sending events to participants they don't have access to.

I think I've noted it before, but I think it would be better if every user continued to have `boards.<boardID>` subscribed. Moderators additionally have `boards.<boardID>.moderator` subscribed. This prevents us from sending more or less the same information in two different channels.

server/src/api/event_filter.go
```diff
@@ -82,6 +81,20 @@ func parseVotingUpdated(data interface{}) (*VotingUpdated, error) {
 	return ret, nil
 }
 
+func parseParticipent(data interface{}) (*dto.BoardSession, error) {
```
brandstetterm, line 84: Typo: Participent &rarr; Participant

server/src/api/boards_listen_on_board.go
```diff
@@ -1,153 +1,221 @@
...
+      if err != nil {
+        logger.Get().Warnw("failed to send message", "message", filteredMsg, "err", err)
+      }
+    }
+
+  }
+}
+
+func (b *BoardSubscription) changeChannel(channel realtime.SessionChannel, oldSession *dto.BoardSession) {
+  if oldSession.Role == types.SessionRoleModerator && channel == realtime.SessionChannelModerator || oldSession.Role == types.SessionRoleParticipant && channel == realtime.SessionChannelParticipant || oldSession.Role == types.SessionRoleOwner && channel == realtime.SessionChannelModerator {
```
brandstetterm, line 187: `oldSession.Role == types.SessionRoleOwner` 
As of know, Owners will always be Owners and they can't change their role

mateo-ivc:
> > Splitted up the board subscription into two channels. Prior every user subscribed to `boards.<boardID>` now based on the role each user connects to `boards.<boardID>.participant` or `boards.<boardID>.moderator`.
> > Why? To avoid sending events to participants they don't have access to.
> 
> I think I've noted it before, but I think it would be better if every user continued to have `boards.<boardID>` subscribed. Moderators additionally have `boards.<boardID>.moderator` subscribed. This prevents us from sending more or less the same information in two different channels.

If we were to have the moderators subscribe to a second channel just for moderator information, we would have to write a new filter that removes the public data, or we would be sending the same information twice.
So I thought it would be a better approach to use the already existing event_filter and just split the channels.
Issue #3413 also suggests splitting the channel, if I understand correctly

brandstetterm:
> > > Splitted up the board subscription into two channels. Prior every user subscribed to `boards.<boardID>` now based on the role each user connects to `boards.<boardID>.participant` or `boards.<boardID>.moderator`.
> > > Why? To avoid sending events to participants they don't have access to.
> > 
> > 
> > I think I've noted it before, but I think it would be better if every user continued to have `boards.<boardID>` subscribed. Moderators additionally have `boards.<boardID>.moderator` subscribed. This prevents us from sending more or less the same information in two different channels.
> 
> If we were to have the moderators subscribe to a second channel just for moderator information, we would have to write a new filter that removes the public data, or we would be sending the same information twice. So I thought it would be a better approach to use the already existing event_filter and just split the channels. Issue #3413 also suggests splitting the channel, if I understand correctly

Yeah, but the intention of the issue has been that the new moderator channel will be exclusively used for the information that normal participants shouldn't see afaik

PR #4122, opened 2024-05-03
fix: close go-routines after session request has been answered

mateo-ivc:
## Description

If the access policy for the board is “manual approval”, each user opens a request and thus a go routine.
Since there is a for loop within the routine that never returns, this routine runs forever.  
To fix this, simply remove the for loop and use a select statement instead.
## Changelog

`api/board_session_requests_listen_on_request.go`
- replaced the for loop with a select statement
## Checklist


server/src/api/board_session_requests_listen_on_request.go
```diff
@@ -63,7 +63,8 @@ func (s *Server) listenOnBoardSessionRequest(boardID, userID uuid.UUID, conn *we
 }
 
 func (b *BoardSessionRequestSubscription) startListeningOnBoardSessionRequest(userId uuid.UUID) {
-	for msg := range b.subscriptions[userId] {
+	select {
+	case msg := <-b.subscriptions[userId]:
```
brandstetterm: Why not like this?
```suggestion
	msg := <-b.subscriptions[userId]:
```

PR #4148, opened 2024-05-16
refactor: summerize all migration files

mateo-ivc:
## Description

Summarized all migrations into the initial sql file
## Changelog

- Removed all migration files
- Added the migrations into initial_schema.up.sql
- Updated initial_schema.down.sql

## Checklist


Schwehn42 (commented):
is there a particular reason the migration files aren't needed anymore? isn't this nice to have in the case one needs to revert to the previous/different state of the db, e.g. in case of another branch?

brandstetterm:
> > is there a particular reason the migration files aren't needed anymore? isn't this nice to have in the case one needs to revert to the previous/different state of the db, e.g. in case of another branch?
> 
> Maybe because it were so many that now was a nice point to summarize them before it gets to chaotic. Mayber @brandstetterm can say more about it.

I thought it's quite hard finding out the current blueprint of our database with so many migration script files. But if there's a good reason why you shouldn't combine them into one (maybe because the project is still ongoing?), we can of course keep them.

Schwehn42:
@brandstetterm @mateo-ivc well personally, I don't mind since I usually flatten the DB anyways before switching to a new feature branch 😂 

But concerning maintainability I would like to hear @Lennart01's opinion on this

PR #4150, opened 2024-05-17
chore: Removal of note observer

BenedictHomuth:
## Description

This PR removes only the observers for notes. It is meant for a proof of concept before removing the remaining observers.

In order to test, I needed to mock the real-time broker (which includes its client aka. the con field).
Before I could not create a Mock-Broker, since the "con" field was not exported. With the changes at hand, I now export the "Con" field, which makes it possible to mock out the broker ultimately leading to running tests 🙂.

## Changelog

- Removed the note observers
- Moved the observer logic inside the service layer
- Made the general code work with those changes
- Exported the Con field of real-time broker in order to be able to mock the real-time package
- Adapted tests to run without observers
## Checklist


mateo-ivc (approved):
lgtm

PR #4160, opened 2024-05-22
chore: removal of board service observer

BenedictHomuth:
## Description

This PR removes the observers for the board service. 
## Changelog

- moved observer logic into service layer
- added logging to before unlogged errors

## Checklist


server/src/services/boards/boards.go
```diff
@@ -133,7 +133,11 @@ func (s *BoardService) BoardOverview(_ context.Context, boardIDs []uuid.UUID, us
 }
 
 func (s *BoardService) Delete(_ context.Context, id uuid.UUID) error {
-	return s.database.DeleteBoard(id)
+	err := s.database.DeleteBoard(id)
+	if err != nil {
+		logger.Get().Errorw("unable to delete board", "err", err)
+	}
```
mateo-ivc, line 139: Wouldn't it be better to log the error in the api layer?
Like this we would check twice if the error is not null

server/src/services/boards/columns.go
```diff
@@ -8,23 +8,37 @@ import (
...
 	"scrumlr.io/server/common/dto"
+	"scrumlr.io/server/common/filter"
 	"scrumlr.io/server/realtime"
 
 	"scrumlr.io/server/database"
 	"scrumlr.io/server/logger"
 )
 
-func (s *BoardService) CreateColumn(_ context.Context, body dto.ColumnRequest) (*dto.Column, error) {
+func (s BoardService) CreateColumn(_ context.Context, body dto.ColumnRequest) (*dto.Column, error) {
```
mateo-ivc, line 18: I don't think its relevant if the receiver is a pointer or value but we should keep the consistency by having all receivers be a pointer or value

server/src/services/boards/columns.go
```diff
@@ -8,23 +8,37 @@ import (
...
 	column, err := s.database.CreateColumn(database.ColumnInsert{Board: body.Board, Name: body.Name, Color: body.Color, Visible: body.Visible, Index: body.Index})
+	if err != nil {
+		logger.Get().Errorw("unable to create column", "err", err)
+	}
```
mateo-ivc, line 22: Shouldn't we return here?
Since the db called failed there are no updates to send

server/src/services/boards/columns.go
```diff
@@ -8,23 +8,37 @@ import (
...
+	s.UpdatedColumns(body.Board)
 	return new(dto.Column).From(column), err
 }
 
 func (s *BoardService) DeleteColumn(_ context.Context, board, column, user uuid.UUID) error {
-	return s.database.DeleteColumn(board, column, user)
+	err := s.database.DeleteColumn(board, column, user)
+	if err != nil {
+		logger.Get().Errorw("unable to delete column", "err", err)
+	}
```
mateo-ivc, line 31: As well return here?

server/src/services/boards/columns.go
```diff
@@ -8,23 +8,37 @@ import (
...
+	s.DeletedColumn(user, board, column)
+	return err
 }
 
 func (s *BoardService) UpdateColumn(_ context.Context, body dto.ColumnUpdateRequest) (*dto.Column, error) {
 	column, err := s.database.UpdateColumn(database.ColumnUpdate{ID: body.ID, Board: body.Board, Name: body.Name, Color: body.Color, Visible: body.Visible, Index: body.Index})
+	if err != nil {
+		logger.Get().Errorw("unable to update column", "err", err)
+	}
```
mateo-ivc, line 40: As well return here?

server/src/services/boards/columns.go
```diff
@@ -51,10 +65,14 @@ func (s *BoardService) ListColumns(ctx context.Context, boardID uuid.UUID) ([]*d
 	return dto.Columns(columns), err
 }
 
-func (s *BoardService) UpdatedColumns(board uuid.UUID, columns []database.Column) {
-	err := s.realtime.BroadcastToBoard(board, realtime.BoardEvent{
+func (s *BoardService) UpdatedColumns(board uuid.UUID) {
+	dbColumns, err := s.database.GetColumns(board)
+	if err != nil {
+		logger.Get().Errorw("unable to retrieve columns in updated notes", "err", err)
+	}
```
mateo-ivc, line 72: Shouldn't we also return here, since dbColumns should be nil

server/src/services/boards/columns.go
```diff
@@ -104,8 +122,12 @@ func (s *BoardService) DeletedColumn(user, board, column uuid.UUID, notes []data
 		logger.Get().Errorw("unable to broadcast updated columns", "err", err)
 	}
 
-	eventNotes := make([]dto.Note, len(notes))
-	for index, note := range notes {
+	dbNotes, err := s.database.GetNotes(board)
+	if err != nil {
+		logger.Get().Errorw("unable to retrieve notes in deleted column", "err", err)
+	}
```
mateo-ivc, line 128: Here as well

server/src/services/boards/columns.go
```diff
@@ -116,8 +138,12 @@ func (s *BoardService) DeletedColumn(user, board, column uuid.UUID, notes []data
 		logger.Get().Errorw("unable to broadcast updated notes", "err", err)
 	}
 
+	boardVotes, err := s.database.GetVotes(filter.VoteFilter{Board: board})
+	if err != nil {
+		logger.Get().Errorw("unable to retrieve votes in deleted column", "err", err)
+	}
```
mateo-ivc, line 144: and here

server/src/services/boards/columns.go
BenedictHomuth: True, completely missed the return calls in this file, my bad.

server/src/services/boards/columns.go
BenedictHomuth: yes

server/src/services/boards/boards.go
BenedictHomuth: In my opinion we should log the error as close to the source as possible.

mateo-ivc (approved):
Looks good so far

PR #4179, opened 2024-05-29
feat: middleware allow actions from authenticated accounts only

mateo-ivc:
## Description

Implemented middleware that allows actions from authenticated accounts only
## Changelog

`api/context.go`
Middleware that checks if the user's account-type is `Anonymous`
`dto/users.go`
added account-type field to the dto struct
## Checklist


Schwehn42 (approved):
lookin' good

might consider adding a test file for all the context functions at some point

Issue #3760, opened 2024-01-16
Can't disable anonymous login


ImDevinC:
### The problem

I have google authentication enabled for my self-hosted scrumlr instance but I can't find a way to disable anonymous login.

### Browser

Chrome

### Steps to reproduce the behavior

Set `SCRUMLR_AUTH_GOOGLE_CLIENT_ID`, `SCRUMLR_AUTH_GOOGLE_CLIENT_SECRET` and `SCRUMLR_AUTH_CALLBACK_HOST`

ImDevinC:
Are there any plans to address that? Without disabling anonymous login, that means scrumlr cannot be hosted on a public address without concern of malicious actors

Schwehn42:
I mean, the fact that no registration is required is one of the main selling points of scrumlr, but I agree that offering the option to disable it sounds sensible. I'll pass the idea on to the team.

Lennart01:
> Are there any plans to address that? Without disabling anonymous login, that means scrumlr cannot be hosted on a public address without concern of malicious actors

For now block `/api/login/anonymous`.
This will prevent a user from creating an anon account.
Depending on your deployment just configure your Webserver/ingress controller to reply with a 403.
Google Callbacks run on a different URL (`/api/login/google/callback`).

PR #4188, opened 2024-06-03
feat: option to disable anonymous accounts

### LINKED ISSUE #4169: Option to disable anonymous login
### Discussed in https://github.com/inovex/scrumlr.io/discussions/3987

<sup>Originally posted by **Schwehn42** March 10, 2024</sup>
Inspired by #3760, there should be an option to disable anonymous login for those who decide to host scrumlr themselves and want to force users to use third party auth.

This would entail the following:

Backend:
- Add CLI Flag and Config Var for disabling anonymous login (default=`false`)
- Blocking the relevant routes for anonmous login if the flag is set (see https://github.com/inovex/scrumlr.io/issues/3760#issuecomment-1930434546)

Frontend:
- Removing or disabing (i.e. gray out) the anonymous login form</div>

Schwehn42:
## Description
This PR adds the option to disable anonymous accounts.  
When disabled (either by using the CLI flag `disable-anonymous-login` or in the `config.toml`), new instances of anonymous accounts cannot be created. Instead, only active providers (Google, Apple, etc.) can be used.
In the frontend, the button is also grayed out (subject to change?)
Closes #4169.

*Note:* Already logged in acounts can be still used

### backend
- `main`
  - Add CLI flag `enable-anonymous-login` and config entry
- `router`
  - add flag to Server struct
  - use context `AnonymousLoginEnabledContext` for route `/login/anonymous`
- `AnonymousLoginEnabledContext`
  - check flag `enable-anonymous-login`, and throw error if false
- `info`
  - add to info struct

### frontend
- store
  - adjust info state to include `enableAnonymousLogin`
- `LoginBoard`
  - conditionally gray out anon login form

## Checklist


server/src/api/router.go
```diff
@@ -130,7 +137,7 @@ func (s *Server) publicRoutes(r chi.Router) chi.Router {
 		r.Post("/feedback", s.createFeedback)
 		r.Route("/login", func(r chi.Router) {
 			r.Delete("/", s.logout)
-			r.Post("/anonymous", s.signInAnonymously)
+			r.With(s.AnonymousLoginEnabledContext).Post("/anonymous", s.signInAnonymously)
```
mateo-ivc, line 140: If I understand this correctly, you are not able to login with a anonymous account. But you would still be able to join a board if you were already logged in.

Schwehn42: Yes exactly. While this seems sloppy on my part it was actually intended; since usually this is an option set at the very start and likely only once, not changed frequently. Since this is built with #3760 in mind, I think it's more likely to be set at the very beginning and never changed.  
The option also titled "enable anonymous login", so imo it isn't unclear that already logged in users still may be used. 

However, I am open to discuss renaming the option to something like "enableAnonymousAccountCreation" or disabling the route for anon users alltogether.

server/src/api/context.go
```diff
@@ -172,6 +172,20 @@ func (s *Server) BoardAuthenticatedContext(next http.Handler) http.Handler {
 	})
 }
 
+func (s *Server) AnonymousLoginEnabledContext(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		log := logger.FromRequest(r)
+
+		if !s.anonymousLoginEnabled {
+			log.Errorw("not allowed to login anonymously")
+			common.Throw(w, r, common.ForbiddenError(errors.New("not authorized to login anonymously")))
```
brandstetterm, line 181: 404 seems more fitting

server/src/main.go
```diff
@@ -102,6 +102,13 @@ func main() {
 				Required: false,
 				Value:    "/",
 			}),
+			altsrc.NewBoolFlag(&cli.BoolFlag{
+				Name:     "enable-anonymous-login",
+				EnvVars:  []string{"SCRUMLR_ENABLE_ANONYMOUS_LOGIN"},
+				Usage:    "enables/disables the login of anonymous clients",
+				Required: false,
+				Value:    true,
+			}),
```
brandstetterm, line 111: since anonymous login being enable should probably be the default, wouldn't it be more logical to have a flag to disable it? -> 'disable-anonymous-login'

src/routes/LoginBoard/LoginBoard.tsx
brandstetterm, line 1: Wouldn't it be better to hide the anonymous login entirely?

server/src/api/context.go
Schwehn42: how so? the route still exists but musn't be used.

server/src/main.go
Schwehn42: it's a matter of semantics but yeah I agree

src/routes/LoginBoard/LoginBoard.tsx
Schwehn42: Since there's no design I decided to do it this way. I think it's more intuitive to explicitely state to use the providers, since anonymous login is something that we advertise so people might be confused as to why it doesn't work. Also, if no providers are enabled there would be a blank screen.  
I could also add a fail-safe that triggers a warning if there is neither anonymous login nor providers available.

Schwehn42: update: added a message if no providers are available at all

mateo-ivc (approved):
lgtm :shipit:

PR #4191, opened 2024-06-03
fix: saveable share qr code

### LINKED ISSUE #4170: Make QR code exportable as image
### Discussed in https://github.com/inovex/scrumlr.io/discussions/3870

<sup>Originally posted by **Resaki1** February 15, 2024</sup>
Currently, the QR code to share a session is rendered as an SVG. Right-clicking it does not give the option to save the QR code as an image (at least in Chromium). Of course you could make a screenshot, but rendering it as an image would make it easier imo</div>

Schwehn42:
## Description

Makes the QR code available in the share view saveable as a file. Note this works for FF and Chrome, but not Safari for me, seems to be a limitation on their part so I won't be bothered 🫣   
Closes #4170
#### `ShareSession`
- change QR code render type from `svg` to `canvas` (makes it saveable)
- increase render size 10-fold (otherwise it's blurry)
## Checklist


brandstetterm:
The QR code doesn't look that nice in dark mode anymore :/ 

Before:
image
After:
image

Schwehn42:
@brandstetterm seems like `qrcode.react` doesn't support setting colors directly using CSS. The workaround used until this point was changing the fill of the svg, however, this isn't possible anymore since we're using a canvas now. Different libraries like https://github.com/rosskhanas/react-qr-code have the same issue. There are two ways to go on about this:
1. set `fgColor` and `bgColor` of element manually using the retreived attribute `theme` (JSX)
2. invert color of whole class using the [filter function](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/invert) if `[theme='dark']` (CSS)

I opted for using 1.,but am open for the other option as well.

PR #4203, opened 2024-06-07
chore: removal of session and session-request observer

BenedictHomuth:
## Description

Removed the session and session request observer and moved the corresponding logic into the service layer.

Two more changes:
1. Created one new db function, since before this logic was written plainly in the observer.

2. Since user updates need to be published to all board members (e.g. on name or appearance change) I added the realtime package to the UserService. Before it was implicitly handled in one of the observers.

## Changelog

- removed session and session observer
- moved logic into corresponding service layers
- Created a new db function which fetches all currently connected sessions of a given user
- Added the realtime package to the UserService
## Checklist


server/src/services/users/users.go
```diff
@@ -74,5 +78,28 @@ func (s *UserService) Update(_ context.Context, body dto.UserUpdateRequest) (*dt
...
+		}
+		err = s.realtime.BroadcastToBoard(session.Board, realtime.BoardEvent{
+			Type: realtime.BoardEventParticipantUpdated,
+			Data: new(dto.BoardSession).From(session),
+		})
+
+		if err != nil {
+			logger.Get().Errorw("unable to broadcast updated user", "err", err)
+		}
+	}
```
mateo-ivc, line 104: If I understand correctly, every time a user's role is updated, a ParticipantUpdated event is sent to all boards the user is connected to?

That is, if a user is currently connected to multiple boards and their role is updated on one board, the event will still be sent to all boards.

server/src/services/boards/sessions.go
```diff
@@ -244,12 +271,23 @@ func (s *BoardSessionService) CreatedSession(board uuid.UUID, session database.B
...
+			logger.Get().Errorw("unable to get board session of user", "board", session.Board, "user", session.User, "err", err)
+			return
+		}
+		err = s.realtime.BroadcastToBoard(session.Board, realtime.BoardEvent{
+			Type: realtime.BoardEventParticipantUpdated,
+			Data: new(dto.BoardSession).From(userSession),
+		})
+		if err != nil {
+			logger.Get().Errorw("unable to broadcast updated board session", "err", err)
+		}
```
mateo-ivc, line 290: I don't think it is necessary to update all active board sessions of user when the status changes on one board

server/src/services/users/users.go
BenedictHomuth: ParticipantUpdated is not only called, when a user is promoted or demoted but also when he changes his appearance (avatar), name etc.

This is why we need to get all currently connected boards of a user and update the changes on all of them.

When we loop over the connectedBoards, each session is retrieved from the database one by one. This is why a name change would be published to all currently connected boards, but a ban would only take effect on a specific board.

server/src/services/boards/sessions.go
BenedictHomuth: See above 😄 

Maybe we should investigate splitting the participant part (board session) and user part (account) going forward for clearer communication and less overhead per change.

server/src/services/users/users.go
BenedictHomuth: It returns the same information as before which therefore should be okay. 😄

server/src/services/boards/sessions.go
brandstetterm: I think splitting would be a good idea

BenedictHomuth: Sounds good to me. 
I will open an issue for it since doing the splitting inside this PR would lead to more confusion than anything else.
To ensure this issue is resolved, I will work on it next before starting anything else.

brandstetterm: Thank you @BenedictHomuth

BenedictHomuth: Details here: #4217

mateo-ivc (approved):
lgtm 👍🏽

Issue #4201, opened 2024-06-06
Basic Framework (Templatepage)


SelinaBuff:
**User Story 1: Set Background**
As a user, I want to see a background that matches the header area of our landing page so that the design is cohesive and works in both light and dark modes.

**User Story 2: Create Template and Recommendations Sections**
As a user, I want clearly distinguishable sections for templates and recommendations. There will be a distinction between registered and anonymous users. Anonymous users will not be able to create their own templates; they can only do so if they are registered. Each section should have a heading that clearly states the purpose of the section.

_Acceptance Criteria:_

- Anonymous Users: There will be only one section (Our Recommendations).

- Registered Users: There will be two separate, equally sized sections in the basic framework: one for templates and one for recommendations.

**User Story 3: Insert Stan in the Background**
Stan should be visible on the left side of the background of the sections. The placement of Stan should be consistent and visually appealing, without overlapping other content.

_The design can be found in Figma under (Template/Session Page) -> (No. 2)_

![Image](https://github.com/inovex/scrumlr.io/assets/116653219/8c502626-48d4-4947-8e4a-b54d6207f45f)

PR #4229, opened 2024-06-17
refactor: add isAnonymous property to auth state

Schwehn42:
## Description

This PR adds a new property `isAnonymous` to the `Auth` state. This is required by #4201 to differentiate between users that registered anonymously or using one of the providers (Google, Apple, etc).  
The reason I didn't just add `accountType` to the `Auth` interface is that I think it's easier to handle a boolean than a string, anticipating that this might be used at more than one place.

## Changelog

- `types/auth.ts`: 
  - Add `isAnonymous` prop to `Auth` interface
  - Add new interface `AuthDto` resembling the actual json retrieved from backend, which was omitted until now.
- `api/auth.ts`
  - `getCurrentUser()`: Cast as `AuthDto`, as this includes the `accountType` property
- `middleware/auth.tsx`
  - set `isAnonymous` to whether the account is anonymous
- Add `isAnonymous` prop to other files depending on `Auth` interface, like `getTestApplicationState`

## Checklist


src/types/auth.ts
```diff
@@ -1,8 +1,20 @@
...
+  avatar?: AvataaarProps;
+}
+
+// as used in frontend
 export interface Auth {
   id: string;
   name: string;
+  isAnonymous: boolean;
   avatar?: AvataaarProps;
 }
```
mateo-ivc, line 19: Imo we could add the accountType to Auth instead of isAnonymouse and remove the AuthDto. To check if an account is anonymous we could perform this check: `accountType != ACCOUNT_TYPE_ANONYMOUS`. But if you think it's better to handle it this way, it works for me too 😄.

mateo-ivc (approved):

src/types/auth.ts
Schwehn42: Where exactly would you perform that check? I think you need the second type because it gets sent like that from the backend; so either perform the cast like I did it the frontend or you change the backend to send `isAnonymous` instead of `accountType`. Just changing the property name in `Auth` would likely result in a type error somewhere.

Schwehn42: I do agree though it looks kinda odd so maybe a backend change is the way to go

mateo-ivc: Everywhere where you need the isAnonymous variable you could perform the `accountType != ACCOUNT_TYPE_ANONYMOUS` check.
But I guess it's not the cleanest solution either.
I don't think changing the backend is a better solution. Just like in the frontend you  would have another user struct wich has the isAnonymous variable instead of AccountType.

Schwehn42: Yeah, that's reason I opted for using another variable in the first place, because the frontend just needs to differentiate between anonymous and non-anonymous users. Whether the authentification used is Google, Apple or whatever should be irrelevant and handling a boolean is easier than a string 😄   
That being said I'll leave it like this for now, changing it in the future is easy enough.

PR #4234, opened 2024-06-18
fix: hidden columns not showing after promotion to moderator

BenedictHomuth:
## Description

Whenever someone gets promoted to moderator, he/she should be able to see the hidden columns. This is now possible since we sync columns (`ColumnsUpdated`) on a `ParticipantUpdated` event. 

Before, this call was not present resulting in a situation where the newly named moderator needed to refresh the page to finally see the hidden columns.

Since #4179 has broadened the scope of a BoardSession DTO – specifically the AccountType – I needed to extend our DB calls, since when I parse the event in the EventFilter, it errors, since AccountType can not be an empty string (at least for now).

Therefore we should check and decide, if we want to publish the AccountType information or not. If not, I can also just annotate the AccountType in the DTO with "omitempty".

## Changelog

- added the accounttype to our db calls 
- added caching to board participants in the event filter
- added the column refresh for when a participant got updated.
## Checklist

### DIFF
--- a/server/src/services/boards/sessions.go
+++ b/server/src/services/boards/sessions.go
@@ -289,6 +289,20 @@ func (s *BoardSessionService) UpdatedSession(board uuid.UUID, session database.B
 			logger.Get().Errorw("unable to broadcast updated board session", "err", err)
 		}
 	}
+
+	columns, err := s.database.GetColumns(board)
+	if err != nil {
+		logger.Get().Errorw("unable to get columns on a updatedsession", "err", err)
+	}
+
+	err = s.realtime.BroadcastToBoard(board, realtime.BoardEvent{
+		Type: realtime.BoardEventColumnsUpdated,
+		Data: dto.Columns(columns),
+	})
+	if err != nil {
+		logger.Get().Errorw("unable to broadcast update columns following a updatedsession", "err", err)
+	}
+
 }
 
 func (s *BoardSessionService) UpdatedSessions(board uuid.UUID, sessions []database.BoardSession) {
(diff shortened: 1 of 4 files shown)


Schwehn42 (commented):
> Therefore we should check and decide, if we want to publish the AccountType information or not. If not, I can also just annotate the AccountType in the DTO with "omitempty".

`AccountType` is required for #4229, does this have an impact on it?

BenedictHomuth:
> > Therefore we should check and decide, if we want to publish the AccountType information or not. If not, I can also just annotate the AccountType in the DTO with "omitempty".
> 
> `AccountType` is required for #4229, does this have an impact on it?

If I see it correctly, the information of `AccountType` is planned to be available in the frontend. Therefore I would suggest keeping this PR as is, since with the changes present in this PR the `AccountType` is now published – at least from a backend perspective.

server/src/services/boards/sessions.go
```diff
@@ -289,6 +289,20 @@ func (s *BoardSessionService) UpdatedSession(board uuid.UUID, session database.B
...
+	}
+
+	err = s.realtime.BroadcastToBoard(board, realtime.BoardEvent{
+		Type: realtime.BoardEventColumnsUpdated,
+		Data: dto.Columns(columns),
+	})
+	if err != nil {
+		logger.Get().Errorw("unable to broadcast update columns following a updatedsession", "err", err)
+	}
+
```
mateo-ivc, line 305: Have you checked if the notes in the hidden column are displayed when a user is promoted to moderator?
As far as I know, BoardEventColumnsUpdated only sends information about the columns, but not about the notes in them

BenedictHomuth: Good catch! 🤝 

Seems that it was to late yesterday 😅 
Fixed it!

mateo-ivc (approved):
lgtm

PR #4246, opened 2024-06-20
refactor: SessionUpdate (Backend)

BenedictHomuth:
## Description

Reduced the amount of information sent when a participant either changes his appearance or status on a board.

Before a `PARTICIPANT_UPDATED` event was published containing all information (account and board session). This is now split into two:
- Account (ID, name, avatar, account type) – for now remained `PARTICIPANT_UPDATED`
- Board Session (connected, banned, etc.) – named `PARTICIPANT_SESSION_UPDATED`

## Changelog

- Updated the BoardSession dto – removing the whole account data and just leaving the id
## Checklist


Schwehn42:
@BenedictHomuth I'll do the frontend part as soon as all/most of the other active PRs are merged into main – I'm worried that with such a change, there are just too many conflicts bound to happen 😅

BenedictHomuth:
> @BenedictHomuth I'll do the frontend part as soon as all/most of the other active PRs are merged into main – I'm worried that with such a change, there are just too many conflicts bound to happen 😅

Definitely worth its own release 😅
Hopefully, TS will make your life a little easier there. Let me know if I can help in any way.

Schwehn42:
Update: I'm going to take a stab at this after the new store (related to https://github.com/inovex/scrumlr.io/pull/4423) is implemented.

Schwehn42 (commented):
I'm a little confused by the new workflow: 

- The "INIT" event gives us only user IDs
- Both the endpoints for `/participants/` and `/participants/{id}` gives us only user IDs

So we receive the session data but not the user data, and there is no endpoint to manually request it.
The user data is sent over the old "PARTICIPANT_UPDATED" event but that isn't activated when joining a board.

We need some initial form for the participant user names etc, after that we can receive session and user updates separately.

SimonRocker:
What is the status here? Is this still relevant? Seems like there were some problems regarding the initial setup with the method only using user ids. There are many conflicts though, as this PR is almost 1.5 years old.

mateo-ivc:
outdated

PR #4255, opened 2024-06-21
chore: add TS Config for Cypress

Schwehn42:
## Description

Adds a TS Config for Cypress, since IDEs would throw multiple errors (like JSON module import not being allowed).
The configuration consists of the folllowing:
- Recommended Config: https://docs.cypress.io/guides/tooling/typescript-support#Configure-tsconfigjson
- [resolveJsonModule](https://www.typescriptlang.org/it/tsconfig/#resolveJsonModule) and [esModuleInterop](https://www.typescriptlang.org/it/tsconfig/#esModuleInterop), both of which are required to import JSON files 

## Changelog

- adds TS Config

### DIFF
--- a/cypress/tsconfig.json
+++ b/cypress/tsconfig.json
@@ -1,0 +1,10 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "lib": ["es5", "dom"],
+    "types": ["cypress", "node"],
+    "resolveJsonModule": true,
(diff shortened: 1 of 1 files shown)


BenedictHomuth:
Not an expert, but LGTM 👍

BenedictHomuth (approved):

PR #4256, opened 2024-06-22
revert: unnecessary changes

Schwehn42:
## Description
Reverts two changes introduced in #4205 which aren't really necessary.

## Changelog
- readds `name` in `docker-compose.yml`
- reverts changes from `tsconfig.json`, this is handled in #4255


Schwehn42:
ok seems like this doesn't work as easily as I thought it would, I'll come back to this some other time 👀

PR #4270, opened 2024-06-26
refactor: avatar settings

Schwehn42:
## Description
Updates types to work with the latest TypeScript version, 5.5.2.

## Changelog

- moves types and constants to separate files
- updates types to make compiler happy 
- some general refactors, like extracting functions and moving them around


src/types/auth.ts
```diff
@@ -1,4 +1,4 @@
-import {AvataaarProps} from "components/Avatar";
```
liza-kl, line 1: Why is it "Avat_aaa_rProps" instead of "AvatarProps" 👀?

src/constants/avatar.ts
```diff
@@ -99,3 +101,32 @@ export type AvatarEyebrowType = (typeof AVATAR_EYEBROW_TYPES)[number];
...
+  ],
+  facialFeatures: [
+    {values: AVATAR_SKIN_COLORS, key: "skinColor"},
+    {values: AVATAR_EYEBROW_TYPES, key: "eyebrowType"},
+    {values: AVATAR_EYE_TYPES, key: "eyeType"},
+    {values: AVATAR_MOUTH_TYPES, key: "mouthType"},
+  ],
+  clothing: [
+    {values: AVATAR_ACCESSORIES_TYPES, key: "accessoriesType", disabledOn: {topType: ["Eyepatch"]}},
+    {values: AVATAR_CLOTHE_TYPES, key: "clotheType"},
```
liza-kl, line 124: If this PR is intended as a general refactoring – maybe rename the constant to `AVATAR_CLOTHING_TYPES` / `AVATAR_CLOTHES_TYPES`– because everything else is also a noun and not a verb.

src/types/avatar.ts
```diff
@@ -0,0 +1,42 @@
...
+  AvatarSkinColor,
+  AvatarTopType,
+} from "constants/avatar";
+
+export interface AvataaarProps {
+  // possibly look into separating accentColor from the other properties at some point
+  // otherwise don't forget to omit when needed
+  accentColorClass: string;
+  skinColor: AvatarSkinColor;
+  topType: AvatarTopType;
```
liza-kl, line 21: Maybe to make it more descriptive, call it "hairStyle"?

src/types/auth.ts
Schwehn42: because the library used under the hood is called [Avataaars](https://www.npmjs.com/package/avataaars), too :)

src/constants/avatar.ts
Schwehn42: Although you're technically correct, the property names were taken from the underlying library. While changing the property keys is no problem is the frontend, the backend would be tricky as the json is stored like that in the backend, which in turn would require some sort of migration to change all existing instances of keys in the json `clotheColor` to `clothingColor`. 

So for a small refactor like this which was just intended to make TS happy I'm actually not to keen on implementing new logic, what do you think?

src/types/avatar.ts
Schwehn42: Same as in https://github.com/inovex/scrumlr.io/pull/4270#discussion_r1705414372

liza-kl (approved):

PR #4180, opened 2024-05-29
feat: stack note complete content

Schwehn42:
## Description

Closes #2627  
Reworks the stack view so that more content ist shown. If exceeding the height limit, the notes are scrollable. On Desktop, the parent will be displayed on the left hand side, while the child notes are on the right hand side. On Mobile, notes will remain on top of each other.

## Changelog

- reorder notes on desktop
- make note content height dynamic
- add media query for mobile
- add a styled scrollbar

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] The light- and dark-theme are both supported and tested
- [x] The design was implemented and is responsive for all devices and screen sizes
- [x] The application was tested in the most commonly used browsers (e.g. Chrome, Firefox, Safari)

## (Optional) Visual Changes

https://github.com/inovex/scrumlr.io/assets/7889564/288f4a22-b78a-4d92-b38e-81d99dcb2b02

Bildschirmfoto 2024-05-29 um 16 26 34
Bildschirmfoto 2024-05-29 um 16 26 01

PR #4277, opened 2024-07-01
feat: marker for edited notes

mateo-ivc:
## Description

Adding a marker that indicates if a note has been edited or not
## Changelog

`dto/notes.go`
- added field if a note has been edited

`database/migrations`
- added field to notes table, that stores if a note has been edited

`database/notes.go`
- updates the edited field in the database to true, since text has been changed

`components/Note`
- added marker beneath text
- added css where needed
## Checklist

## (Optional) Visual Changes

image

image


src/components/Note/Note.scss
```diff
@@ -140,6 +134,20 @@ $note__footer-gap: 42px;
...
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+  text-overflow: ellipsis;
+
+  display: -webkit-box;
+}
+
+.note__marker-edited {
+  color: rgb(163, 166, 170);
+  font-size: 13px;
```
Schwehn42, line 148: try using the constant colors and spacing found in `constans/style.scss` wherever possible (in this case `$color-middle-gray` and `$spacing--sm`)

src/components/Note/Note.scss
```diff
@@ -175,6 +183,9 @@ $note__footer-gap: 42px;
 .note__author-container {
   flex-grow: 1;
   overflow: hidden;
+  display: flex;
+  gap: 5px;
```
Schwehn42, line 187: same here, refer to constant name

src/utils/test/getTestApplicationState.ts
Schwehn42, line 1: try adding a test where you check if an edited note has the class `note__marker-edited`:)

src/components/Note/Note.tsx
```diff
@@ -120,10 +119,14 @@ export const Note = (props: NoteProps) => {
             />
           </div>
         ) : (
-          <main className={classNames("note__text", {"note__text--extended": !showNoteReactions})}>
-            <NoteTextContent text={note.text} truncate />
+          <main className={classNames("note__container")}>
+            <div className={classNames("note__text", {"note__text--extended": !showNoteReactions})}>
+              <NoteTextContent text={note.text} truncate />
+            </div>
+            {note.edited && <div className={classNames("note__marker-edited")}>({t("Note.edited")})</div>}
```
Schwehn42, line 126: don't need to use the `classNames` package here if only one class is provided. At least for my taste, it isn't wrong ofc

Schwehn42 (changes requested):
looking good all in all! 👍 I do have to backtrack about my opinion that having the edited class below doesn't matter though; having it inline would really look better since it doesn't impact the height. Only problem is that I have no idea how to achieve that 😅

src/components/Note/Note.scss
Schwehn42: Also use a lighter gray for dark mode (try looking which color we used for other components?)

Resaki1:
> looking good all in all! 👍 I do have to backtrack about my opinion that having the edited class below doesn't matter though; having it inline would really look better since it doesn't impact the height. Only problem is that I have no idea how to achieve that 😅

just my two cents, but I strongly believe this current solution to be better:
1. This way it is directly visible that the `(edited)` is not part of the note's text.
2. This is how other platforms solve it. Jakob's Law (hehe) states that users prefer our site to work like all other sites, and if you look at Whatsapp, Twitter etc., they solve it just like @mateo-ivc has solved it here.
3. I personally think increasing the height of the note slightly is less bad then cutting the text off earlier than necessary.

What do you think @Schwehn42?

mateo-ivc:
>  do have to backtrack about my opinion that having the edited class below doesn't matter though; having it inline would really look better since it doesn't impact the height.

I talked with Selina about that. She said that it's okay this way. But if we find a solution I am happy to change it so that it is inline

Schwehn42:
@Resaki1 @mateo-ivc good points, let's do it that way then 👍

Schwehn42 (dismissed):

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.scss
```diff
@@ -41,14 +41,24 @@
...
     &:hover,
     &:focus-visible {
       border-color: var(--accent-color--desaturated-dark);
     }
   }
 }
+.note-dialog__note-content-container {
+  width: 100%;
+  height: 100%;
+}
```
Schwehn42, line 54: don't think this actually makes a difference. if you wanna make any changes to the note please make sure that #4180 isn't impacted by it

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.scss
```diff
@@ -41,14 +41,24 @@
...
+
+.note-dialog__marker-edited {
+  color: $color-middle-gray;
+  font-size: $text--base;
+  border: 2px transparent;
```
Schwehn42, line 59: what does the transparent border do?

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.scss
```diff
@@ -41,14 +41,24 @@
...
+  padding: 0 4px;
```
Schwehn42, line 60: use spacing constant

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
```diff
@@ -87,28 +90,30 @@ export const NoteDialogNoteContent: FC<NoteDialogNoteContentProps> = ({noteId, a
...
+                  e.currentTarget.blur();
+                }
 
+                if (e.key === "Escape") {
+                  e.currentTarget.blur();
+                  e.stopPropagation();
+                }
+              }}
+            />
+            <div className="note-dialog__marker-edited">({t("Note.edited")})</div>
```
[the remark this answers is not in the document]
mateo-ivc: ups 🫢

src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.scss
mateo-ivc: That's old css i used when the marker was displayed in the note header. Forgot to remove it 😅

Schwehn42 (approved):
lookin' good 🙌

PR #4290, opened 2024-07-04
fix: adding skin tone to emojis

mateo-ivc:
## Description

In the boa rdsettings a user is able to select their prefered skin tone. But till now all emojis would use the default color.

## Changelog

`types/boardReaction`
- changing map from <ReactionType, string> to <ReactionType, {string, boolean}> since not all emojis support skin tones

`types/reaction`
- changing map from <ReactionType, string> to <ReactionType, {string, boolean}> since not all emojis support skin tones

`components/BoardReaction`
- adding variable that includes then skin tone if its possible for a given emoji 

`components/BoardReactionMenu`
- adding variable that includes then skin tone if its possible for a given emoji

`components/NoteReactionBar`
- adding variable that includes then skin tone if its possible for a given emoji

`components/NoteReactionChip`
- adding variable that includes then skin tone if its possible for a given emoji

`components/NoteReactionChipCondensed`
- displaying emoji with skin tone if possible

## Checklist


src/components/BoardReaction/BoardReaction.tsx
```diff
@@ -50,9 +51,11 @@ export const BoardReaction = memo((props: BoardReactionProps) => {
     }
   }, [containerRef, containerDimensions, displayOffset]);
 
+  const emojiWithSkinTone = emoji!.skinToneSupported ? emoji!.emoji + skinTone.component : emoji!.emoji;
```
Schwehn42, line 54: for this a util func should be used since it's repeated at multiple places

src/components/Note/NoteReactionList/NoteReactionBar/NoteReactionBar.tsx
```diff
@@ -4,6 +4,7 @@ import classNames from "classnames";
 import {REACTION_EMOJI_MAP, ReactionType} from "types/reaction";
 import {ReactionModeled} from "../NoteReactionList";
 import "./NoteReactionBar.scss";
+import {useAppSelector} from "../../../../store";
```
Schwehn42, line 7: just using "store" as import path is enough. also works for "types", "constants", "components" etc.  
Also always put the stylesheet as last item

src/components/Note/NoteReactionList/NoteReactionBar/NoteReactionBar.test.tsx
```diff
@@ -29,8 +29,10 @@ describe("NoteReactionBar", () => {
   it("focuses the first emoji button after TAB is pressed on the last emoji button", async () => {
     render(<NoteReactionBar closeReactionBar={() => {}} reactions={[]} handleClickReaction={() => {}} />);
 
-    const firstEmojiButton = screen.getByText([...REACTION_EMOJI_MAP.values()].at(0)!);
-    const lastEmojiButton = screen.getByText([...REACTION_EMOJI_MAP.values()].at(-1)!);
+    // @ts-ignore
+    const firstEmojiButton = screen.getByText([...REACTION_EMOJI_MAP.values()].at(0).emoji!);
```
Schwehn42, line 33: add a `!` after `.at(0)` to assert the value is defined, then you can remove the ts-ignore

src/types/boardReaction.ts
```diff
@@ -9,12 +9,12 @@ export type BoardReactionType = {
...
-  ["applause", "👏"],
-  ["heart", "💖"],
-  ["like", "👍"],
-  ["dislike", "👎"],
+export const BOARD_REACTION_EMOJI_MAP = new Map<ReactionType, {emoji: string; skinToneSupported: boolean}>([
+  ["tada", {emoji: "🎉", skinToneSupported: false}],
+  ["applause", {emoji: "👏", skinToneSupported: true}],
+  ["heart", {emoji: "💖", skinToneSupported: false}],
+  ["like", {emoji: "👍", skinToneSupported: true}],
+  ["dislike", {emoji: "👎", skinToneSupported: true}],
```
Schwehn42, line 17: works for now, but in the future this might be customizable. fortunately `constants/emojis.json` exists with all the necessary information

Schwehn42 (commented):
Why is the skintone just stored locally btw? Would make more sense for it to be another db column in users imo, and then be retrieved from `store.user`. That way, other people can also see the respective skin colors, which makes more sense if you ask me.

src/components/Note/NoteReactionList/NoteReactionBar/NoteReactionBar.test.tsx
mateo-ivc: > Why is the skintone just stored locally btw? Would make more sense for it to be another db column in users imo, and then be retrieved from store.user. That way, other people can also see the respective skin colors, which makes more sense if you ask me.

I thought the same about this.
Would definitely implement this

mateo-ivc:
> Why is the skintone just stored locally btw? Would make more sense for it to be another db column in users imo, and then be retrieved from store.user. That way, other people can also see the respective skin colors, which makes more sense if you ask me.

I thought the same about this.
Would definitely implement this

src/components/Note/NoteReactionList/NoteReactionBar/NoteReactionBar.test.tsx
mateo-ivc: the problem is when i remove the ts-ignore it gets displayed as an error and the tests still fail

Schwehn42: that's a different issue though, seems like a missing `Provider` is the problem, you can see how it's solved in other tests

Schwehn42:
> I thought the same about this.
Would definitely implement this

@mateo-ivc you could also consider changing the logic so the emoji is sent with the skin tone included, then it wouldn't need to be manually added afterwards

src/components/Note/NoteReactionList/NoteReactionBar/NoteReactionBar.test.tsx
```diff
@@ -1,36 +1,45 @@
...
 
   it("calls handleClickBar when a reaction button is clicked", () => {
-    const handleClickReaction = jest.fn();
-    render(<NoteReactionBar closeReactionBar={() => {}} reactions={[]} handleClickReaction={handleClickReaction} />);
+    const handleClickReactionFunction = jest.fn();
+    renderNoteReactionBar(undefined, handleClickReactionFunction);
     fireEvent.click(screen.getByText("👍"));
-    expect(handleClickReaction).toHaveBeenCalled();
+    expect(handleClickReactionFunction).toHaveBeenCalled();
   });
```
Schwehn42, line 30: could you add a test where you overwrite the `skinTone` state and see if the 👍 colored correctly?

src/components/Note/NoteReactionList/NoteReactionBar/NoteReactionBar.test.tsx
```diff
@@ -1,40 +1,57 @@
...
     lastEmojiButton.focus();
     expect(lastEmojiButton).toHaveFocus();
     await userEvent.tab();
     expect(firstEmojiButton).toHaveFocus();
   });
+
+  it("render emoji in different skin tone", () => {
+    renderNoteReactionBar(undefined, undefined, {skinTone: {name: "dark", component: "🏿"}});
+    REACTION_EMOJI_MAP.forEach((emoji) => {
+      expect(screen.getByText(getEmojiWithSkinTone(emoji, {name: "dark", component: "🏿"}))).toBeInTheDocument();
```
mateo-ivc, line 54: Is there a way to get the skin tone from the state?

Schwehn42: `getTestStore()` in the `<Provider/>` returns a store where you could get the state. However, I believe it would be not the best practice to test the store itself. Instead, we should focus about unit testing what the impact of the changed state is (i.e. changed components etc)

src/components/Note/NoteReactionList/NoteReactionChipCondensed/NoteReactionChipCondensed.tsx
```diff
@@ -5,6 +5,8 @@ import {uniqueId} from "underscore";
 import {TooltipPortal} from "components/TooltipPortal/TooltipPortal";
 import {ReactionModeled} from "../NoteReactionList";
 import "./NoteReactionChipCondensed.scss";
+import {useAppSelector} from "../../../../store";
+import {getEmojiWithSkinTone} from "../../../../utils/reactions";
```
Schwehn42, line 9: one last nitpick 😄 (don't forget to put the `scss` as last import)

Schwehn42 (changes requested):

src/components/Note/NoteReactionList/NoteReactionChipCondensed/NoteReactionChipCondensed.tsx
mateo-ivc: 😄

Schwehn42: the stylesheet still isn't the last import 😭

Schwehn42 (approved):
lgtm

PR #4300, opened 2024-07-11
fix: color

Schwehn42:
## Description
adjusts note and column input color for dark mode, as there were visibility problems


BenedictHomuth (approved):
:shipit:

PR #4283, opened 2024-07-02
refactor: color palette

Schwehn42:
## Description
Replaces old color palette with new one. This unifies the general theme of the frontend. Also, some adjustments to the theme colors have been made.

### New Colors
`$blue--800`: `#002366`
`$blue--700`: `#003499`
`$blue--600`: `#0046cc`
`$blue--500`: `#0057ff` primary light
`$blue--400`: `#3379ff` primary dark
`$blue--300`: `#669aff`
`$blue--200`: `#99bcff`
`$blue--100`: `#ccddff`

`$pink--800`: `#66002a`
`$pink--700`: `#99003f`
`$pink--600`: `#cc0054`
`$pink--500`: `#ff0069` primary light
`$pink--400`: `#ff3387` primary dark
`$pink--300`: `#ff66a5`
`$pink--200`: `#ff99c3`
`$pink--100`: `#ffcce1`

`$purple--800`: `#500e66`
`$purple--700`: `#791599`
`$purple--600`: `#a11ccc`
`$purple--500`: `#c923ff` primary light
`$purple--400`: `#d44fff` primary dark
`$purple--300`: `#df7bff`
`$purple--200`: `#e9a7ff`
`$purple--100`: `#f4d3ff`

`$violet--800`: `#250062`
`$violet--700`: `#380094`
`$violet--600`: `#4a00c5`
`$violet--500`: `#5d00f6` primary light
`$violet--400`: `#7d33f8` primary dark
`$violet--300`: `#9e66fa`
`$violet--200`: `#be99fb`
`$violet--100`: `#dfccfd`

`$green--800`: `#0b5532`
`$green--700`: `#107f4c`
`$green--600`: `#16aa65`
`$green--500`: `#1bd47e` primary light
`$green--400`: `#49dd98` primary dark
`$green--300`: `#76e5b2`
`$green--200`: `#a4eecb`
`$green--100`: `#d1f6e5`

`$yellow--800`: `#666200`
`$yellow--700`: `#999300`
`$yellow--600`: `#ccc400`
`$yellow--500`: `#fff500` primary light
`$yellow--400`: `#fff733` primary dark
`$yellow--300`: `#fff966`
`$yellow--200`: `#fffb99`
`$yellow--100`: `#fffdcc`

`$orange--800`: `#664300`
`$orange--700`: `#996500`
`$orange--600`: `#cc8600`
`$orange--500`: `#ffa800` primary light
`$orange--400`: `#ffb933` primary dark
[...]

PR #4304, opened 2024-07-12
refactor: accent colors

Schwehn42:
## Description

This PR refactors the accent colors. It's basically the second part to #4283, which changed all the color definitions.  

### Before
Previously, the two primary colors (light/dark) for each color was used to define the accent colors. For instance, using the classname `accent-color__backlog-blue` would yield the accent color `--accent-color: $blue--500` (and `accent-color--dark: $blue--400` respectively). On top of that, further CSS variables like `--accent-color-desaturated-hover-dark` were defined. The definitions were completely arbitrary and were used at own discretion instead of the actual design colors. 

Furthermore, some designs use accent colors beyond 500 and 400, e.g. the new primary/secondary buttons. This actually motivated me to refactor this.

### After
Now, not only the primary colors are used as accent colors, but all colors from the new schema are available. (e.g. `--accent-color--300`)
The primary colors are still easily accessible using shortcuts (`--accent-color--light`, `--accent-color-dark`).
Also, all shades are available as rgb, which means alpha channel can be applied.
"Excess" variables like the above mentioned desatured.. have been replaced with the new ones.

## Changelog

- move colors to separate file
- add colors to map
- new mixins `set-accent-variables` and `generate-accent-classes`
  - generate accent variables for each color class (like `backlog-blue`)
  - shades from `--accent-color--100` to `--accent-color-800`
  - shortcuts `--accent-color--light` and `--accent-color--dark`
  - split rgb version for each shade
  - remove excess accent variables
- use new accent colors in every file that requires them 
- some manual adjustements (e.g. concerning note reactions) to fix some colors

## Checklist


src/components/ScrumlrLogo/ScrumlrLogo.scss
```diff
@@ -39,7 +39,7 @@
   }
 
   .scrumlr-logo__stop {
-    stop-color: var(--accent-color);
+    stop-color: var(--accent-color--light);
```
BenedictHomuth, line 42: The logo seems broken.
image

Schwehn42: Oops I meant #4310 of course!

Schwehn42: Should be fixed with #4312 now

BenedictHomuth (approved):
lgtm

src/components/ScrumlrLogo/ScrumlrLogo.scss
BenedictHomuth: nice

PR #4310, opened 2024-07-15
refactor: theme state

Schwehn42:
## Description
Changes the logic how setting of the theme is handled.

### Before
Previously, the logic was split weirdly by using the local storage directly to set/get the theme.  
Also, the Html component would hold the logic to determine what the system theme preference is.

### After
Now, `theme` ist also part of the application store and can be set/retrieved as such with Actions etc.  
The theme is is also still saved to the local storage and retrieved initially, but the local storage key name has been streamlined with the scrumlr prefix to fit with the others.

Additionally, a hook is now used to determine the theme which is set as a global attribute.

## Changelog
- Add `theme` to the application store (as part of view), including Actions, Reducer, and Middleware.
- The middleware will now handle setting the local storage implicitly (new key: `scrumlr/theme`).
- Added a hook `useAutoTheme` which will return the currently used theme (`"light" | "dark"`), even if the theme is set to `"auto"`.
- Adjust `Html.tsx` and `ThemeSettings.tsx` to use the new app state / hook.

## Checklist

### DIFF
--- a/src/components/Html/Html.tsx
+++ b/src/components/Html/Html.tsx
@@ -1,4 +1,4 @@
-import React, {FC} from "react";
+import React, {FC, useEffect} from "react";
 import {Helmet} from "react-helmet";
 import {useAppSelector} from "store";
 import {useAutoTheme} from "../../utils/hooks/useAutoTheme";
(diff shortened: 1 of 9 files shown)


BenedictHomuth (approved):
LGTM and works in all common browsers

PR #4296, opened 2024-07-09
build: bump postgres to 16.3

BenedictHomuth:
## Description

Since our deployment will switch to postgres 16.3 shortly, this is also now reflected in our dev env.
## Changelog

- Added the new version to our docker-compose files
- Added the new version to our test container in Go tests

## Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] The light- and dark-theme are both supported and tested
- [ ] The design was implemented and is responsive for all devices and screen sizes
- [ ] The application was tested in the most commonly used browsers (e.g. Chrome, Firefox, Safari)

PR #4311, opened 2024-07-15
chore: update PostgreSQL image to version 16.3

Lennart01:
Bumps dev cluster Postgres to 16.3


Schwehn42 (approved):
I'm assuming this + #4296 should do the trick?

Issue #4280, opened 2024-07-02
Add subtitle field to column titles


SelinaBuff:
_### Discussed in https://github.com/inovex/scrumlr.io/discussions/4276_

### Description:** 
Currently, our users can only enter a title for each column. To improve user experience, we should extend the functionality to include an additional field (below the title) for a short description/subtitle.

**Acceptance Criteria:**
- Users can successfully enter, save, and display a subtitle for each column.
- Changes to the subtitle can be made at any time (procedure: same as for the title).
- The layout remains clear and user-friendly.
- If necessary, a maximum character limit must be set, or the text should be truncated with (...) -> The full text will then be displayed using a tooltip.

**Requirements:**

**Design Requirements:**
- Integrate a field for the subtitle below the existing title field.
- Both fields (title and subtitle) must be clearly labeled to avoid confusion.
- Responsive design.
- Support for both dark and light modes.

**Database Adjustments:**
- Modify the database schema to enable storing subtitles.
- Ensure that existing data migration is performed without data loss.

BenedictHomuth:
The backend part is implemented in #4331!

SelinaBuff:
Design is ready (Figma: 🧩 Components).

<img width="823" alt="Image" src="https://github.com/user-attachments/assets/a7074686-de17-411d-ac12-afde3ed1e367" />

Schwehn42:
Since the template editor has a similar title / description edit component, it's maybe possible to utilize or adjust the code for this purpose, see 080a94fecaaabc9430c09a38a9066126bce27b6c

Schwehn42:
labeled as blocked until the above mentioned code is merged, because it really makes sense to use the components that were introduced there, like `TextArea` (part of create/edit templates, https://github.com/inovex/scrumlr.io/tree/js/create-template-view).

PR #4331, opened 2024-07-21
feat: column description – backend

BenedictHomuth:
## Description

This PR includes the backend part for #4280.

For now I coined the term to be column "Description". When it should be column "Subtitle", we should change this before merging.

## Changelog

- added SQL migration file
- added needed logic to backend code
- updated the go tests
- updated the postman collection


server/src/database/migrations/sql/19_add_column_subtitles.down.sql
Schwehn42, line 1: I like `description` for the column name 👍  but the file should be named accordingly

BenedictHomuth: true, resolved

mateo-ivc (approved):
lgtm :shipit:

PR #4335, opened 2024-07-22
test: service test for UpdateVoting 

liza-kl:
## Description

Add a test for the UpdatingVoting functionality in service layer 
## Changelog

Changed `services/votings_test.go`
## Checklist


BenedictHomuth:
Looks good! 
Can we also add a test for when we update a voting with the status open? This would test whether we correctly throw an error for this condition.

BenedictHomuth (approved):
:shipit:

PR #4336, opened 2024-07-22
fix: check for error when updating user

mateo-ivc:
## Description

If the user update fails and we try to parse the value into a dto, the code will panic
## Checklist


server/src/services/users/users.go
```diff
@@ -79,6 +79,9 @@ func (s *UserService) Update(_ context.Context, body dto.UserUpdateRequest) (*dt
 		Avatar: body.Avatar,
 	})
 	s.UpdatedUser(user)
```
BenedictHomuth, line 81: Maybe we should check it even before we send an update (with`s.UpdatedUser`) to the users.
Additionally, it would be nice to create a log message before returning the error in case we need to debug it in the future.

BenedictHomuth (changes requested):

server/src/services/users/users.go
BenedictHomuth: Hard to tell where to log 😅 
We use it differently for different files. But when it is logged in the API layer, it should be sufficient for now :D

BenedictHomuth (approved):

PR #4338, opened 2024-07-22
refactor: split board template and column template 

BenedictHomuth:
## Description

Refactored our template feature. Before, it was too cluttered, meaning that the board and column templates were always combined in one request.

Now, each component is on its own to provide more fine-grain control.

Therefore we also expose more endpoints

Before: 
- only `/templates`

Now: 
- `/templates`
- `/templates/columns`

This way one can also manipulate the columns without touching the general settings of a board template.
## Changelog

- refactor and splitting of board template and column template
- update of postman collection


mateo-ivc (approved):
:shipit:

PR #4346, opened 2024-07-23
fix: colors

Schwehn42:
## Description
Quick fix:
1. Templates _Positive / Negative_ and _Plus / Delta_ had the wrong color for the second column.
2. The color selection order for columns was wrong
## Changelog

- Adjust template colors for the two above mentioned templates
- Adjust Column color order


BenedictHomuth (approved):
lgtm

PR #4347, opened 2024-07-23
fix: note reaction autofocus

Schwehn42:
### Before
When opening the note reaction bar using mouse, the first note reaction was automatically focused despite that not being the intended behaviour

### After
That no longer happens

## Changelog
`NoteReactionBar`: Remove `returnFocus`, add `autoFocus={false}` to `ReactFocusLock`


BenedictHomuth (approved):
lgtm

PR #4352, opened 2024-07-24
test: added boardDescription to go tests and fixture file

BenedictHomuth:
## Description

Additional tests for the board description feature
## Changelog

- extended `fixture.yml` file
- added go tests in `/database/boards_test.go`


mateo-ivc (approved):
lgtm 
 I just noticed that we still have some observer data inside the fixture.yml that are not being used.
 Are they still necessary or can we remove them?

BenedictHomuth:
Good catch!
No they are not needed, since the observer tests are gone as well.
I opened a PR for the fixture removal here: #4354 4354

PR #4339, opened 2024-07-22
chore: arm64 images on release & action dependency upgrades

Lennart01:
Bumps a couple dependencies.
Introduces ARM images on release.

PR #4361, opened 2024-07-25
Allow flexible architecture on containerbuilds depending on the build platform.

tibeer:
Based on <https://docs.docker.com/build/guide/multi-platform/#platform-build-arguments>
you can use variables like `TARGETARCH` or `TARGETOS` to determine the architecture used in the build command
inside the Dockerfile. This would allow multiarch binary builds for the server.

relates #4339
relates #3994

Signed-off-by: Tim Beermann <tibeer@outlook.de>

## Description

The current multi-arch builds only build the container image for the desired architecture, the go binary inside is still build with the wrong architecture. This PR tries to automatically adopt the build arguments to the related architecture given in the build command.

## Changelog

Changed the server Dockerfile to allow platform agnostic build arguments. E.g. arm64 when docker buildx is running for platform linux/arm64.

## Checklist

### DIFF
--- a/server/src/Dockerfile
+++ b/server/src/Dockerfile
@@ -1,8 +1,12 @@
 # Build application
 FROM golang:1.22
+
+ARG TARGETARCH
+ARG TARGETOS
+
 WORKDIR /go/src/github.com/inovex/scrumlr.io/
 COPY ./ .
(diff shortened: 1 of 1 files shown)


Lennart01:
Awesome this should do the trick.
I will merge this into a temp branch and create a new PR so that all our checks can run and we can test it once. 👍.

PR #4372, opened 2024-07-29
ci: changed all uses of local branches to main

BenedictHomuth:
## Description

With this PR all workflows / actions now run the CI files from the main branch. This seems to be the best practice to keep all CI activities in sync with main. It also ensures that the main branch is stable, since the checks are run against the main version – meaning that if those pass, main should not be crashed when merging PR branches.

This usage of CI running against main is already present in the steps of  `package`,  `deploy-to-dev-cluster`, `deploy-docs` etc.

## Changelog

- changed `ci.yml`


brandstetterm:
What happens if they're changed in a PR? We wouldn't know if they work since GitHub will be using the actions from the main branch, right?

BenedictHomuth:
> What happens if they're changed in a PR? We wouldn't know if they work since GitHub will be using the actions from the main branch, right?

By default, yes.
The idea here is to purposefully change the reference from `main` to `dev_branch` when modifying the CI.
Then, you can check if your CI changes work as expected in your `dev_branch`. 

Before merging into main the reference must be reset.

brandstetterm:
> > What happens if they're changed in a PR? We wouldn't know if they work since GitHub will be using the actions from the main branch, right?
> 
> By default, yes. The idea here is to purposefully change the reference from `main` to `dev_branch` when modifying the CI. Then, you can check if your CI changes work as expected in your `dev_branch`.
> 
> Before merging into main the reference must be reset.

But isn't that prone to errors?

BenedictHomuth:
> > > What happens if they're changed in a PR? We wouldn't know if they work since GitHub will be using the actions from the main branch, right?
> > 
> > 
> > By default, yes. The idea here is to purposefully change the reference from `main` to `dev_branch` when modifying the CI. Then, you can check if your CI changes work as expected in your `dev_branch`.
> > Before merging into main the reference must be reset.
> 
> But isn't that prone to errors?

Could be.

Nonetheless, we should standardize our methodology. Either by always testing against `main` or always using the `local` branch. Not a mix of both.

We can talk in the daily about that and accept a final conclusion.

BenedictHomuth:
> > > > What happens if they're changed in a PR? We wouldn't know if they work since GitHub will be using the actions from the main branch, right?
> > > 
> > > 
> > > By default, yes. The idea here is to purposefully change the reference from `main` to `dev_branch` when modifying the CI. Then, you can check if your CI changes work as expected in your `dev_branch`.
> > > Before merging into main the reference must be reset.
> > 
> > 
> > But isn't that prone to errors?
> 
> Could be.
> 
> Nonetheless, we should standardize our methodology. Either by always testing against `main` or always using the `local` branch. Not a mix of both.
> 
> We can talk in the daily about that and accept a final conclusion.

Talked it through in our daily. For now, we concluded to use the `local` version. Therefore the PR is updated.

brandstetterm (approved):

PR #4376, opened 2024-07-30
ci: updated actions; adapted to new output syntax

BenedictHomuth:
## Description

Bumps some deps for our CI. Additionally, since GitHub will deprecate its former `set-output` syntax, the new syntax is used now.

## Changelog

- bump deps in `package.yml` and `deploy_to_dev_cluster.yml`
- changed to new output syntax in `ci.yml`


brandstetterm (approved):
Thank you!

Issue #4269, opened 2024-06-26
Enhanced visual feedback for "Mark me as done"


SelinaBuff:
### Ticket: Enhanced visual feedback for "Mark me as done"

**Description:**

Some users are still unclear about when they should mark themselves as "done." Therefore, we need to add visual feedback for two use cases (Timer and Voting) in addition to the wobbling pill. This enhancement aims to provide clearer guidance and reduce user confusion about when to mark themselves as "done." Consistency in visual feedback across different functionalities (Timer and Voting) is crucial for a seamless user experience.

**Acceptance Criteria:**

**Timer Ends:**
- When the timer expires, emits a sound, and starts wobbling, the "Mark me as done" option should automatically open in the left user menu with a 2-second delay.
- This option should remain open until the user marks themselves as "done" either through the left user menu or within the timer pill (see screenshot).
- Additionally, the timer should automatically close 30 seconds after expiring (including the "Mark me as done" option). All users who have not marked themselves as "done" by then will be automatically marked as "done."

**Voting Session:**
- Once a user has submitted all their votes and the pill starts wobbling, the "Mark me as done" option should automatically open in the left user menu with a 2-second delay.
- This option should remain open until the user marks themselves as "done" either through the left user menu or within the timer pill (see screenshot).

**Technical Requirements:**
1. Implement a function that triggers the opening of the "Mark me as done" option with a 2-second delay after the timer expires or the voting session ends.
2. Ensure the "Mark me as done" option remains open until the user marks themselves as "done."
3. Implement a feature that automatically closes the timer and the "Mark me as done" option 30 seconds after the timer expires, marking all users who haven't done so as "done."
4. Ensure the visual feedback and interaction are consistent with the design shown in the provided screenshot.

![Image](https://github.com/inovex/scrumlr.io/assets/116653219/fe1909d2-65ea-483a-a8cb-f64d119a4b3c)
![Image](https://github.com/inovex/scrumlr.io/assets/116653219/d0766f57-f496-4cac-a6f0-cf038feecd08)

PR #4383, opened 2024-08-01
feat(menu-bar): Enhanced visual feedback for "Mark me as done"

liza-kl:
## Description
- Implemented the enhanced feedback behaviour for the "Mark me as Done" tooltip, as stated in [#4269](https://github.com/inovex/scrumlr.io/issues/4269).

- The cases are handled as follows (or at least this was the intention):
• **Voting is active / No timer** → Showing the open tooltip after 2 seconds and leaving it open until the user clicks. Clicking logs the toggle interaction. If a user decides to unmark them, the feedback shall not be shown. 
• **Voting is active / timer is active / votes are used up **before** the timer ends** → Showing the tooltip after 2 seconds and leaving it open a) until the user clicks or b) until 30 seconds after the timer expires and the user is set to ready automatically 

• **Voting not active / Timer is active** → After timer expires and the user is not ready, showing the tooltip after 2 seconds until 30 seconds after expiration and then setting the user automatically to done. 

• **Voting is active / timer has expired / votes are not used up**: same behaviour as _Voting is not active, but timer is active_ 
• Voting not active, timer active, user toggles their state twice within 30 seconds after expiration -> don't show the open tooltip for the second time.  

## Changelog

- Add hooks that implement logic in `MenuBars.tsx`
- Add tests for implemented logic 
## Checklist

### DIFF
--- a/src/components/MenuBars/__tests__/MenuBars.test.tsx
+++ b/src/components/MenuBars/__tests__/MenuBars.test.tsx
@@ -1,9 +1,12 @@
-import {render} from "testUtils";
 import {MenuBars} from "components/MenuBars";
 import {Provider} from "react-redux";
-import getTestStore from "utils/test/getTestStore";
 import {MockStoreEnhanced} from "redux-mock-store";
+import {render} from "testUtils";
+import getTestStore from "utils/test/getTestStore";
+import {useTimer} from "../../../utils/hooks/useTimerLeft";
 import getTestParticipant from "../../../utils/test/getTestParticipant";
+
+jest.mock("../../../utils/hooks/useTimerLeft");
 
 const createMenuBars = (store: MockStoreEnhanced) => (
   <Provider store={store}>
@@ -12,7 +15,13 @@ const createMenuBars = (store: MockStoreEnhanced) => (
 );
 
 describe("MenuBars", () => {
+  beforeEach(() => {
+    (useTimer as jest.Mock).mockReturnValue({timerExpired: false});
+  });
+
   test("should match snapshot", () => {
+    (useTimer as jest.Mock).mockReturnValue({timerExpired: false});
+
     const store = getTestStore({
       participants: {
         self: getTestParticipant({role: "MODERATOR"}),
@@ -24,6 +33,8 @@ describe("MenuBars", () => {
   });
 
   test("should render both user- and admin-menu for moderators", () => {
+    (useTimer as jest.Mock).mockReturnValue({timerExpired: false});
+
     const store = getTestStore({
       participants: {
         self: getTestParticipant({role: "MODERATOR"}),
@@ -36,6 +47,8 @@ describe("MenuBars", () => {
   });
 
   test("should only render user-menu for participants", () => {
+    (useTimer as jest.Mock).mockReturnValue({timerExpired: false});
+
     const store = getTestStore({
       participants: {
         self: getTestParticipant({role: "PARTICIPANT"}),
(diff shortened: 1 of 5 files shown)


Schwehn42 (commented):
I'm admittedly kinda confused by the behaviour:

- when a timer is active, all participants get set to ready after 30secs. The 30s don't start after the timer has expired but after the timer was started, which can't be right, can it?
- that is also the case when a voting is active, and it doesn't matter whether one actually votes
- the timer itself doesn't seem to disappear after 30s after expiration

brandstetterm:
> b) until 30 seconds after the timer expires and the user is set to ready automatically

Nice work! Two questions: Why do participants get set to ready 30s after the timer has ended automatically? What makes you think that the competitor is in fact ready?

src/utils/hooks/useTimerLeft.ts
```diff
@@ -0,0 +1,32 @@
+import {useEffect, useState} from "react";
+import {Timer as TimerUtils} from "../timer";
+
+/**
+ * A hook which calculates the time left based on a provided end Date.
+ * Intended for: https://github.com/inovex/scrumlr.io/issues/4217
+ * @param timerEnd T
```
brandstetterm, line 7: The explanation of the parameter seems to be missing

src/utils/hooks/useTimerLeft.ts
```diff
@@ -0,0 +1,32 @@
...
+      } else {
+        timer = setTimeout(updateTimer, 250);
+        setTimerExpired(false);
+      }
+    };
+    updateTimer();
+    return () => clearTimeout(timer);
+  }, [timerEnd]);
+
+  return {timerExpired};
```
brandstetterm, line 31: Is it possible to only return the boolean value of `timerExpired` or does it have to be wrapped within an object?

src/components/TooltipButton/TooltipButton.tsx
```diff
@@ -22,6 +23,7 @@ export const TooltipButton = (props: TooltipButtonProps) => {
       className={classNames("tooltip-button", `tooltip-button--${props.direction ?? "right"}`, {"tooltip-button--active": props.active}, props.className)}
       onClick={() => props.onClick()}
       aria-label={props.label}
+      data-testid={props.dataTestid}
```
brandstetterm, line 26: Using data test ids should be avoided if possible since it doesn't resemble how our application is used

src/components/MenuBars/__tests__/MenuBars.test.tsx
```diff
@@ -47,3 +62,54 @@ describe("MenuBars", () => {
...
+    jest.useFakeTimers();
+    (useTimer as jest.Mock).mockReturnValue({timerExpired: true});
+    const store = getTestStore({
+      participants: {
+        self: getTestParticipant({ready: true}),
+        others: [],
+      },
+    });
+    const {getByTestId} = render(createMenuBars(store));
+    const button = getByTestId("mark-as-done-button");
```
brandstetterm, line 106: Using test ids should be avoided if possible, since it doesn't resemble the real world usage of our application. It always better to test from a user perspective => Using text or labels to get elements.

liza-kl:
> > b) until 30 seconds after the timer expires and the user is set to ready automatically
> 
> Nice work! Two questions: Why do participants get set to ready 30s after the timer has ended automatically? What makes you confident that the participants are in fact ready?

This was a technical requirement (No. 3) stated in #4269. However, I see that I haven't implemented the "closes the timer automatically". Shall we remove the 30s rule then in general?

brandstetterm:
> This was a technical requirement (No. 3) stated in #4269. However, I see that I haven't implemented the "closes the timer automatically". Shall we remove the 30s rule then, in general?

Thanks for the clarification! To be honest, I'm not really keen on automatically marking participants as ready, as we can never be sure that they are actually ready. However, I also don't want to simply change requirements for features that @SelinaBuff has defined. I think I'm going to start a discussion in our Slack channel to find out what the others think.

src/components/MenuBars/__tests__/MenuBars.test.tsx
liza-kl: Replaced by using text now :)

src/components/TooltipButton/TooltipButton.tsx
liza-kl: Removed it

liza-kl:
> > This was a technical requirement (No. 3) stated in #4269. However, I see that I haven't implemented the "closes the timer automatically". Shall we remove the 30s rule then, in general?
> 
> Thanks for the clarification! To be honest, I'm not really keen on automatically marking participants as ready, as we can never be sure that they are actually ready. However, I also don't want to simply change requirements for features that @SelinaBuff has defined. I think I'm going to start a discussion in our Slack channel to find out what the others think.

That's a good idea. Maybe we should also discuss the edge case @Schwehn42 pointed out :)

Schwehn42 (commented):
Still some issues with it, especially when adding a new timer or resetting it after it ran out. Didn't test with the votes yet, just the timer.
Hard reset means concluding the current timer before starting a new one, while soft reset means just initializing a timer, even though the old one is still active (despite being at 0s).

| Use Case         | Shows Banner |
|------------------|--------------|
| initial timer     | ✅            |
| mid soft reset   | ✅            |
| mid add +1       | ✅            |
| post soft reset  | ⛔            |
| post add +1      | ⛔            |
| post hard reset  | ⛔            |

liza-kl:
- New "simplified" logic (was discussed and approved)
  - Everytime a condition for a ready state is fulfilled (e.g. timer expired, or votes are used up) show the extended tooltip after 2 seconds for another 28 seconds.

laila-rin (approved):
Looks good!

Issue #4362, opened 2024-07-25
inline emoji menu is shown behind notes


BenedictHomuth:
### The problem

When there are notes on a board, and I want to type in an emoji for a new note, the menu is displayed behind the already present notes.

Similar issue when I want to edit a note, that was already present

### Browser

Chrome

### Screenshots

On board:
![image](https://github.com/user-attachments/assets/91c56dbe-0fe4-402d-8e44-98aad0265d90)

In Note view:
![image](https://github.com/user-attachments/assets/d4a34fd7-57e3-4915-9c9e-b5c6f9e2ab23)

liza-kl:
Fix proposed in #4391

PR #4391, opened 2024-08-05
fix(emoji-suggestion): fix emoji suggestion overlay

liza-kl:
## Description

Bugfix for Issue: #4362

## Checklist


src/components/Column/Column.scss
```diff
@@ -49,6 +49,7 @@
 }
 
 .column__header {
+  z-index: 1;
```
Schwehn42, line 52: lgtm, just one small thing, could you use one of the variables provided in `style.scss` instead of a hardcoded value? Like `$base-z-index`.

Schwehn42 (approved):
lgtm 🙏

PR #4392, opened 2024-08-05
feat: import exported boards

mateo-ivc:
## Description

Since we have an export feature it would make sense to be able to import those boards again.
Therefore, when creating a new board, there is now an option to import a JSON file that reflects a board.

### Backend
`api/router.go`
- Added _import_ route

`api/boards.go`
- Implementation of import board route
- Parses the json string into the components that will be inserted into the database

`dto/...`
- Added structs to parse the JSON data into usable structs.

`database/notes.go`
- Added a query to import notes

`services/notes.go`
- Added import method for notes -> calls importNote in the DB layer. 

### Frontend
`api/boards.ts`
- Implementation of importBoard route -> sends json string
 
`Note/NoteAuthorList`
- Added authorID of note to be able to create a avatar prop based on the authorID

`PassphraseModal`
- Popup that asks for a password when importing a previously password-protected board.

`constants/name.ts`
- Since imported Notes with the same author should have the same avatar and name, i implemented a function that returns a random name with seed

`NewBoard`
- Added a new button to import a board by drag and drop or by selecting a file.
- Asks for password if board was prior password protected
## Checklist


Schwehn42:
@mateo-ivc is this ready?

mateo-ivc:
> @mateo-ivc is this ready?

I think the implementation is done, but I should add some tests

src/api/board.ts
```diff
@@ -35,6 +35,27 @@ export const BoardAPI = {
...
+        credentials: "include",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: boardJson,
+      });
+
+      if (response.status === 201) {
+        const body = await response.json();
+        return body.id;
```
Schwehn42, line 51: can you cast the returned body to its respective type?

src/components/PassphraseDialog/PassphraseModal/PassphraseModal.scss
```diff
@@ -0,0 +1,54 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.password-modal__container {
+  padding: $spacing--lg;
+  background: #ffffff;
+  border-radius: 0.5em;
+  box-shadow: 0 10px 20px rgba(black, 0.2);
```
Schwehn42, line 21: you can use colors from `colors.scss` instead of hardcoded hex values or named standard colors (white being `$gray--000`, for example).

src/components/PassphraseDialog/PassphraseModal/PassphraseModal.scss
```diff
@@ -0,0 +1,54 @@
...
+  left: 50%;
+  max-width: 90%;
+  position: absolute;
+  top: 50%;
+  transform: translate(-50%, -50%);
```
Schwehn42, line 26: so I'm assuming you centered the container here. Can you make the the parent container flex instead and use `align-items` and `justify-content` instead, or is there any specific reason to do it this way?

src/components/PassphraseDialog/PassphraseModal/PassphraseModal.scss
```diff
@@ -0,0 +1,54 @@
...
+}
+
+.password-modal__header span {
+  margin: 0;
+}
+
+[theme="dark"] {
+  .password-modal__container {
+    background: $color-form-background--dark;
+    color: white;
```
Schwehn42, line 52: again, both are hardcoded colors

src/constants/nameList.ts
Schwehn42, line 1: This is so the other authors keep the same name, right? This is why I was a proponent of keeping the real names, but oh well can't do anything about it 😅 

Please move this to `utils/random` though

src/routes/NewBoard/NewBoard.tsx
```diff
@@ -19,8 +22,109 @@ export const NewBoard = () => {
...
+    setFile(file);
+    reader.readAsText(file);
+  };
+
+  const onImportBoard = async () => {
+    if (completeBoard && accessPolicy === AccessPolicy.BY_PASSPHRASE && passphrase) {
+      completeBoard.board.passphrase = passphrase;
+    }
+
+    const boardId = await API.importBoard(JSON.stringify(completeBoard));
```
Schwehn42, line 110: direct API calls like that are not in the spirit of our architecture. Please use a thunk for this

src/routes/NewBoard/NewBoard.scss
```diff
@@ -33,6 +33,10 @@
     }
   }
 }
+.new-board__mode:last-child {
+  grid-column: 1 / -1;
+  max-width: 100%;
+}
```
Schwehn42, line 39: not sure about this, looks kinda weird as well 👀 should probably ask @SelinaBuff what's better

src/routes/NewBoard/NewBoard.tsx
```diff
@@ -1,26 +1,118 @@
...
+      name: string;
+      description?: string;
+      accessPolicy: string;
+      passphrase?: string;
+    };
+    columns: Column[];
+    notes: Note[];
+    participants: Participant;
+    voting: Voting;
+  };
```
Schwehn42, line 45: I think it should be made more clear that this type is for an imported board. Also, it might make sense to move this to the `types` file for boards.

src/routes/NewBoard/NewBoard.tsx
```diff
@@ -1,26 +1,118 @@
...
+    if (!file) return;
+
+    const reader = new FileReader();
+    reader.onload = (e) => {
+      const content = e.target?.result as string;
+      try {
+        const data = JSON.parse(content) as BoardContent;
+        if (data.board.accessPolicy === "BY_PASSPHRASE") {
+          setShowPasswordModal(true);
+          setAccessPolicy(1);
```
Schwehn42, line 83: instead of `1`, use `AccessPolicy.BY_PASSPHRASE`

Schwehn42: What happens if the value for accessPolicy is `AccessPolicy.BY_INVITE`? can you set the access policy directly to the value of `data.board.accessPolicy`?

src/routes/NewBoard/NewBoard.tsx
```diff
@@ -1,26 +1,118 @@
...
+      } catch (error) {
+        Toast.error({title: t("Toast.failedImport")});
+        setFile(new File([], ""));
+      }
+    };
+    setFile(file);
+    reader.readAsText(file);
+  };
+
+  const onImportBoard = async () => {
```
Schwehn42, line 95: since we're not calling the API directly anymore the `async` can be removed

src/routes/NewBoard/NewBoard.scss
mateo-ivc: I talked to @SelinaBuff  about this, and the plan was to have the import button across the entire width so that it lines up nicely with the two rows.

Schwehn42:
Hi @mateo-ivc please change the label to "Review required" if you think this is ready :)

src/routes/NewBoard/NewBoard.tsx
mateo-ivc: If accessPolicy has the value `AccessPolicy.BY_INVITE` it will be imported as a board with this accessPolicy. The reason I am checking if the policy is set to `PASSPHRASE` is to open the PasswordModal

Schwehn42 (approved):
LGTM, you just need to change variables using the `AccessPolicy` type which was changed with #4601 ✌️

PR #4397, opened 2024-08-08
feat: nginx sec headers

Lennart01:
## Checklist


security-headers.conf
Schwehn42, line 1: I'm usually a proponent of splitting stuff into multiple files, however I'm not sure in this case because it adds another file to the project root. Could you add in into the main file or if you want to split it maybe add a subdirectory?

security-headers.conf
```diff
@@ -0,0 +1,8 @@
+# Security headers
+add_header X-XSS-Protection "1; mode=block" always;
```
SimonRocker, line 2: Isn't this deprecated?

security-headers.conf
```diff
@@ -0,0 +1,8 @@
...
+add_header X-Content-Type-Options "nosniff" always;
+add_header Referrer-Policy "no-referrer-when-downgrade" always;
+add_header Content-Security-Policy "default-src 'self' http: https: ws: wss: data: blob: 'unsafe-inline'; frame-ancestors 'self';" always;
+add_header Permissions-Policy "interest-cohort=()" always;
```
SimonRocker, line 6: Isn't floc gone?

security-headers.conf
```diff
@@ -0,0 +1,8 @@
...
+add_header X-Frame-Options "DENY" always;
+add_header Expect-CT "max-age=86400, enforce" always;
```
SimonRocker, line 8: Still relevant? https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Expect-CT

PR #4406, opened 2024-08-13
build: bump postgres to 16.4

BenedictHomuth:
## Description

With our current version we are vulnerable to [CVE-2024-7348](https://www.cve.org/CVERecord?id=CVE-2024-7348), which was ranked of "high" severity. This vulnerability is fixed in version 16.4, hence we should update to this version shortly.

## Changelog

- Bumped postgres in all related files


Schwehn42 (approved):
lgtm 😎

PR #4428, opened 2024-08-23
feat: add new column menu

laila-rin:
## Description
Changed column menu layout to new design.

## Changelog
- added colorpicker dropdown
- added MiniMenu and removed hardcoded items in column settings menu
- added horizontally/vertically flipped version for mobile
 
## Checklist

## (Optional) Visual Changes

image
image


src/components/Column/Column.scss
```diff
@@ -229,6 +229,7 @@
   height: 24px;
   width: 24px;
   transition: all 0.08s ease-out;
+  transform: rotate(90deg);
```
brandstetterm, line 232: The icon is turned 90 degrees. This doesn't seem right 😄 
Screenshot 2024-08-24 at 11 23 13

src/components/Column/ColumnSettings.scss
```diff
@@ -1,108 +1,10 @@
...
-
-  &:hover {
-    transform: scale(1.1);
-  }
-
-  &:focus-visible {
-    box-shadow: 0 0 0 2px $navy--300;
-  }
+.column_settings {
+  z-index: 220;
```
brandstetterm, line 4: z-index doesn't have an effect here since the element isn't positioned anymore (since you've removed `position: absolute`)

src/components/Column/ColumnSettings.scss
```diff
@@ -1,108 +1,10 @@
...
 }
 
 // Hacky way to not show delete column button if it's the last column
 .column:only-of-type .column__header-menu-dropdown > ul > li:nth-child(5) {
   display: none;
 }
```
brandstetterm, line 10: Don't think that has an effect anymore too

src/components/Column/ColumnSettings.scss
brandstetterm, line 1: Since the z-index doesn't have an effect here, and the hacky css to hide the delete button shouldn't work anymore too, I think you can delete this file

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,59 @@
...
+  onClose?: () => void;
+  colors: Color[];
+};
+
+export const ColorPicker: React.FC<ColorPickerProps> = ({id, name, visible, index, color, onClose, colors}) => {
+  const dispatch = useDispatch();
+  const colorsWithoutSelectedColor = colors.filter((curColor) => curColor !== color);
+  const primColorAnchor = uniqueId(`color-picker-${color.toString()}`);
+  return (
+    <ul className="column__header-color-picker-dropdown">
```
brandstetterm, line 23: Since it's a standalone component, I'd choose the classname `color-picker`

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <Tooltip anchorSelect={`#${primColorAnchor}`} content={color.toString()} />
+      </li>
+      {colorsWithoutSelectedColor.map((item) => {
+        const anchor = uniqueId(`color-picker-${item.toString()}`);
+        return (
+          <li>
+            <button
+              id={anchor}
+              aria-label={item.toString()}
+              title={item.toString()}
```
brandstetterm, line 45: I don't find e.g. `planning-pink` as a label so nice here. Much better would be `Planning pink`

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,272 @@
...
+
+  .online-orange_selected {
+    background: #ffb933;
+    border: 3px solid #ffdc99;
+  }
+
+  .online-orange:hover {
+    border: 3px solid #ffeecc;
+  }
+}
```
brandstetterm, line 261: Please use color variables

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,272 @@
...
+
+.planning-pink_selected {
+  background: #ff3387;
+  border: 3px solid #ff99c3;
+}
+
+.planning-pink:hover {
+  border: 3px solid #ffcce1;
+}
```
brandstetterm, line 86: Would the following be a nice simplification? 
```suggestion
.planning-pink {
  background-color: $pink--400;
  border: 3px solid $gray--300;
  
  &-selected {
    border-color: $pink--200;
  }
  &:hover {
    border-color: $pink--100;
  }
}
```

Also, the styling for focus is missing
```
.planning-pink:focus {
  ...
}
```

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,272 @@
...
+
+.column__header-color-picker-dropdown > li {
+  all: unset;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  //width: 40px;
+  //height: 40px;
+  cursor: pointer;
```
brandstetterm, line 55:
```suggestion
```
You have it already on the button element

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,272 @@
...
+  padding: 0;
+  width: 40px;
+  height: 40px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.column__header-color-picker-dropdown > li {
+  all: unset;
```
brandstetterm, line 49: Why is it necessary?

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,272 @@
...
+
+[theme="dark"] {
+  .column__header-color-picker-dropdown {
+    background-color: $navy--600;
+  }
+}
+
+.column__header-color-option {
+  width: 23px;
+  height: 23px;
```
brandstetterm, line 70: 23 px seems like a odd number 😆 Why not 24?

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,272 @@
...
+  border-radius: 100px;
```
brandstetterm, line 71:
```suggestion
  border-radius: $rounded--full;
```

src/components/Column/ColumnSettings.tsx
```diff
@@ -36,109 +40,66 @@ export const ColumnSettings: FC<ColumnSettingsProps> = ({id, name, color, visibl
...
+              onNameEdit?.();
+              onClose?.();
+            },
+          },
+          {
+            label: t("Column.resetName"),
+            icon: <Close />,
+            onClick: () => (setOpenColumnSet ? setOpenColumnSet((o) => !o) : () => {}),
+          },
+        ]}
```
brandstetterm, line 101: I think it could be beneficial if you extract the items array to a constant. I think it would improve the readability and make the JSX less bloated

src/components/MiniMenu/MiniMenu.scss
```diff
@@ -0,0 +1,92 @@
...
+  min-height: 48px;
+  padding-left: $spacing--xxs 2px;
+
+  border-radius: $rounded--full;
+  box-shadow: 0 0 20px rgba($navy--400, 0.12);
+
+  background-color: $gray--000;
+}
+.mini-menu__item {
+  all: unset;
```
brandstetterm, line 22: Why is it necessary here?

src/components/Column/ColumnSettings.tsx
```diff
@@ -36,109 +40,66 @@ export const ColumnSettings: FC<ColumnSettingsProps> = ({id, name, color, visibl
     dispatch(Actions.createColumnOptimistically({id: TEMPORARY_COLUMN_ID, name: "", color: randomColor, visible: false, index: columnIndex}));
   };
 
+  const renderColorPicker = () =>
+    openedColorPicker ? (
+      <ColorPicker id={id} name={name} visible={visible} index={index} color={color} onClose={onClose} colors={COLOR_ORDER} />
+    ) : (
+      <span className={`column__header-color-option ${color}_selected`} />
```
brandstetterm, line 47: `${color}_selected` doesn't look like valid BEM

I think something like `column__header-color-option--${color}-selected` would be valid

brandstetterm (commented):
Hey, great work! Some comments:

The new column settings don’t seem to be accessible for keyboard users. Some suggestions:
- The visual of the three dots button should be improved. I’d suggest, instead of changing the color from black to the accent color on focus, we show an outline in the accent color of the column.
- If you’re pressing enter on the three dots button, the column settings open and the focus should automatically shift to the first settings item.
- Pressing TAB selects the focuses the next settings item. 
- A focus lock should be used. Pressing TAB on the last settings item shifts the focus to the first settings item again. You are “trapped” inside of the column settings. Pressing ESC or pressing Enter on a settings item (e.g. to hide the column) should close the settings menu and return the focus to the three dots button again. Pressing Enter on the color settings item should open up the color picker, focus the first (or maybe current?) color and trap the focus within the color picker. Selecting a new color or pressing ESC (to cancel) should return the focus to the color settings item of the column settings again.

I was only shown a tooltip for the button to close the settings, but not for the other buttons. 

https://github.com/user-attachments/assets/40d72b70-a499-4e3b-8712-312f6c92a665

src/components/MiniMenu/MiniMenu.scss
```diff
@@ -0,0 +1,92 @@
...
+        background-color: $navy--600;
+      }
+    }
+  }
+  .mini-menu {
+    background-color: $navy--400;
+  }
+}
+
+// Hacky way to not show delete column button if it's the last column
```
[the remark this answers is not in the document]
laila-rin: nope, fair point :)

src/components/Column/Column.scss
laila-rin: haha that was supposed to only affect the three dots icon:D just rotated the html element directly now

src/components/Column/ColumnSettings.scss
laila-rin: there's no need for it in the desktop view, in the mobile view however the "add note" input field overlaps the menu otherwise

src/components/ColorPicker/ColorPicker.scss
laila-rin: i thought it was necessary to reset the list elements properties to their default values here but it actually works without, good point! :)

src/components/ColorPicker/ColorPicker.scss
laila-rin: it's 23px in the design :D should I change it to 24?

src/components/MiniMenu/MiniMenu.scss
laila-rin: to override the native styles of mini-menu__items

src/components/Column/ColumnSettings.scss
laila-rin: The z-index is needed for the mobile view - should i keep the file just for that of have the styling in the html element?

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,364 @@
...
+      &--goal-green {
+        border: 3px solid $green--100;
+      }
+      &--yielding-yellow {
+        border: 3px solid $yellow--100;
+      }
+      &--online-orange {
+        border: 3px solid $orange--100;
+      }
+    }
```
Schwehn42, line 52: by adding the color class name to this or a parent element, you can replace all of this with `border: 3px: var(--accent-color--100)`. It will automatically pick the correct one :)

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,364 @@
+@import "../../constants/style";
+
+.color-picker {
+  z-index: 220;
```
Schwehn42, line 4: hard-coded

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,364 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  padding-bottom: $spacing--xxs;
+  box-shadow: 0 0 20px rgba(55, 62, 79, 0.12);
```
Schwehn42, line 10: hard-coded, this color probably exists in the color definitions. do the following:
1. find out hex code of color by converting from rgb
2. find the color in `colors.scss` that has this value
3. use `rgba($the-right-color, 0.12)`

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,364 @@
...
+
+    &:focus-visible {
+      background-color: $blue--400;
+      border: 3px solid $blue--100;
+    }
+  }
+
+  &--poker-purple {
+    background-color: $purple--400;
+    border: 3px solid $gray--300;
```
Schwehn42, line 124: like above, you can remove all of the different color definitions with just one by using the accent color

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,364 @@
...
+
+    &--poker-purple {
+      border: 3px solid $navy--600;
+
+      &-selected {
+        border: 3px solid $purple--200;
+      }
+
+      &:hover {
+        border: 3px solid $purple--100;
```
Schwehn42, line 272: and again

src/components/ColorPicker/ColorPicker.scss
Schwehn42: of course, these also exist, for `--200`, `--300`, and so on up to `--800`. In addition to that, `--accent-color--xxx-rgb` also exists, like `--accent-color--100-rgb`, if you ever need to adjust things like the alpha value.

Oh and also there are the shortcuts `--accent-color--light` and `--accent-color--dark`, which equate to `--500` and `--400` (also available as `-rgb`) ✌️

laila-rin: didn't know that was a thing, thanks!

src/components/Column/ColumnSettings.tsx
```diff
@@ -36,109 +41,89 @@ export const ColumnSettings: FC<ColumnSettingsProps> = ({id, name, color, visibl
     dispatch(Actions.createColumnOptimistically({id: TEMPORARY_COLUMN_ID, name: "", color: randomColor, visible: false, index: columnIndex}));
```
Schwehn42, line 41: this code isn't by you, but still wanted to comment that this actually doesn't lead to a fair distribution of colors, since 100 isn't divisible by 7.
```ts
const randomColor = getColorForIndex(Math.floor(Math.random() * COLOR_ORDER.length));
```
would probably be better.

src/components/Column/ColumnSettings.tsx
```diff
@@ -36,109 +41,89 @@ export const ColumnSettings: FC<ColumnSettingsProps> = ({id, name, color, visibl
...
+        name={props.name}
+        visible={props.visible}
+        index={props.index}
+        color={props.color}
+        onClose={props.onClose}
+        colors={COLOR_ORDER}
+        closeColorPicker={() => setOpenedColorPicker(false)}
+      />
+    ) : (
+      <span className="column__header-color-option column__header-color-option-selected" />
```
Schwehn42, line 68: because of BEM, the correct classname is `--selected`

src/components/Column/ColumnSettings.tsx
```diff
@@ -36,109 +41,89 @@ export const ColumnSettings: FC<ColumnSettingsProps> = ({id, name, color, visibl
...
+    );
+
+  const menuItems = [
```
Schwehn42, line 71:
```suggestion
  const menuItems: MiniMenuItem[] = [
```

of course the type needs to exported from `MiniMenu` for this to work, but I prefer the compiler knowing the correct type

src/components/Column/ColumnSettings.tsx
```diff
@@ -36,109 +41,89 @@ export const ColumnSettings: FC<ColumnSettingsProps> = ({id, name, color, visibl
...
+      icon: <Edit />,
+      onClick: () => {
+        props.onNameEdit?.();
+        props.onClose?.();
+      },
+    },
+    {
+      label: t("Column.resetName"),
+      icon: <Close />,
+      onClick: () => (props.setOpenColumnSet ? props.setOpenColumnSet((o) => !o) : () => {}),
```
Schwehn42, line 120:
```suggestion
      onClick: () => props.setOpenColumnSet?.((o) => !o),
```

src/components/Column/ColumnSettings.tsx
```diff
@@ -36,109 +41,89 @@ export const ColumnSettings: FC<ColumnSettingsProps> = ({id, name, color, visibl
...
-          <button
-            aria-label="Online Orange"
-            title="Online Orange"
-            className={classNames(getColorClassName("online-orange"), "column__color-button")}
-            onClick={() => dispatch(Actions.editColumn(id, {name, color: "online-orange", index, visible}))}
-          />
-        </li>
-      </ul>
+    <div ref={columnSettingsRef} className="column_settings">
+      <MiniMenu className="" items={menuItems} />
```
Schwehn42, line 126: I'd add a className, something like `column-settings__mini-menu`

src/components/Column/ColumnSettings.tsx
```diff
@@ -36,109 +41,89 @@ export const ColumnSettings: FC<ColumnSettingsProps> = ({id, name, color, visibl
...
-          />
-          <button
-            aria-label="Online Orange"
-            title="Online Orange"
-            className={classNames(getColorClassName("online-orange"), "column__color-button")}
-            onClick={() => dispatch(Actions.editColumn(id, {name, color: "online-orange", index, visible}))}
-          />
-        </li>
-      </ul>
+    <div ref={columnSettingsRef} className="column_settings">
```
Schwehn42, line 125: BEM once again, `column-settings` looks better

Schwehn42 (commented):
just some nitpicks :)

Schwehn42 (commented):
If there's not enough space, the menu looks like this:

Screenshot 2024-09-11 at 15 46 20

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,116 @@
...
+
+  > li {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+
+    > button {
+      cursor: pointer;
+      border: none;
+      background: none;
```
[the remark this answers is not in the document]
brandstetterm: Imo, not every element needs to have a dedicated classname (and it's also overkill sometimes). So for me it's fine to use css selectors.

src/components/Column/ColumnSettings.tsx
brandstetterm: Just adding a classname to prevent it from being an empty string also seems unnecessary. If the classname wouldn't be used, just remove it by making the property optional.

laila-rin: That was a relic tbh. I removed it, it's already an optional property

src/components/Column/Column.tsx
```diff
@@ -175,7 +175,11 @@ export const Column = ({id, name, color, visible, index}: ColumnProps) => {
       )}
       {!isTemporary && (
         <button title={t("Column.settings")} className="column__header-edit-button" onClick={() => setOpenedColumnSettings((o) => !o)}>
-          {openedColumnSettings ? <Close className="column__header-edit-button-icon" /> : <ThreeDots className="column__header-edit-button-icon" />}
+          {openedColumnSettings ? (
+            <Close className="column__header-edit-button-icon" />
+          ) : (
+            <ThreeDots className="column__header-edit-button-icon" style={{transform: "rotate(90deg)"}} />
```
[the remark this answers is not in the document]
laila-rin: having the icon rotated as part of the design is only done once in the whole project and only for this one icon here. All other icons are using the column__header-edit-button-icon class without any changes to it. i kind of thought it overkill to have a class dedicated only for the one use

src/components/Column/Column.tsx
```diff
@@ -175,7 +175,11 @@ export const Column = ({id, name, color, visible, index}: ColumnProps) => {
       )}
       {!isTemporary && (
         <button title={t("Column.settings")} className="column__header-edit-button" onClick={() => setOpenedColumnSettings((o) => !o)}>
-          {openedColumnSettings ? <Close className="column__header-edit-button-icon" /> : <ThreeDots className="column__header-edit-button-icon" />}
+          {openedColumnSettings ? (
+            <Close className="column__header-edit-button-icon" />
```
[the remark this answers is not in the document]
laila-rin: True, fair point

src/components/Column/ColumnSettings.tsx
laila-rin: sure, we're all for color equality here:D

src/components/ColorPicker/ColorPicker.scss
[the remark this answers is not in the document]
laila-rin: in this case I'd stick to css selectors just because fitting it into BEM’s naming convention would just mean calling the classes __item or __button instead of having a list item or button css selector, which i don't find too beneficial here tbh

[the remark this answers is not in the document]
Schwehn42: while it doesn't make a syntactical difference per se, it's more about consistency among all files for me. Also in the file, the `&`-selector is used as well so it's kind of a mix. But if you think this is more readable, I'm fine with it.

src/components/Column/Column.tsx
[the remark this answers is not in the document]
Schwehn42: that's true in this case, but mixing inline styles and stylesheets can lead to confusion and doesn't seem like a good practice to me. you could change the class name to `column__header-edit-button-icon column__header-edit-button-icon--menu` and add 
```scss
&--menu {
  transform: rotate(90deg);
}
```
in the stylesheet.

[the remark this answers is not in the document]
laila-rin: as discussed, leaving it to avoid funky behavior while hovering

src/components/ColorPicker/ColorPicker.scss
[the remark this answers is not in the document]
laila-rin: fair point, inconsistent naming isn't the way to go

Schwehn42 (approved):
LGTM! 🥳  
just one small thing, tabbing doesn't work in Firefox, however, I just noticed it also doesn't seem to work for note reactions, so I propose we create a separate issue to resolve both (_edit: #4500_).

PR #4488, opened 2024-09-11
test: fix cypress CI

Schwehn42:
## Description

Cypress, which uses the root ts config, didn't work anymore because of a configuration which isn't allowed anymore.

## Changelog

- remove suppressImplicitAnyIndexErrors from TS config


mateo-ivc (approved):
lgtm :shipit:

PR #4523, opened 2024-10-02
Mi/refactor logging

mateo-ivc:
## Description

Creating a more structured and less complex logging layout
## Changelog

`api/*`
- logging in front of every .decode(...) 
`services/*`
- logging after every database call
- removed logging after every realtime.Boradcast... since they get logged when publishing the event 
- using logger.FromContext() to log more information obtain the request
## Checklist


server/src/logger/logger.go
```diff
@@ -26,7 +26,8 @@ func init() {
 // EnableDevelopmentLogger constructs a development logger and overwrites the default production logger
 // this is only for development
 func EnableDevelopmentLogger() {
```
Schwehn42, line 28: This function is called from `main.go` if a docker environment is found. I don't comprehend how that is an indicator of the server being run in dev mode, as the server is usually called within a docker environment, isn't it?

server/src/logger/logger.go
```diff
@@ -26,7 +26,8 @@ func init() {
...
-	loggerConfig := zap.NewDevelopmentConfig()
+	loggerConfig := zap.NewProductionConfig()
+	loggerConfig.EncoderConfig.StacktraceKey = "" // remove this line and add to init() for production
```
Schwehn42, line 30: Can you explain why this is needed here but not in production level? Since it's the only difference between the two

server/src/logger/logger.go
```diff
@@ -46,6 +47,7 @@ func FromRequest(r *http.Request) *zap.SugaredLogger {
 // to the pre initialized logger.
 func FromContext(ctx context.Context) *zap.SugaredLogger {
 	l, ok := ctx.Value(ctxRequestLogger).(*zap.SugaredLogger)
+	l.WithOptions()
```
Schwehn42, line 50: Why are you applying empty options?

server/src/services/notes/notes.go
```diff
@@ -112,65 +113,56 @@ func (s *NoteService) Delete(ctx context.Context, body dto.NoteDeleteRequest, id
...
 	return err
 }
 
 func (s *NoteService) UpdatedNotes(board uuid.UUID) {
-	notes, dbErr := s.database.GetNotes(board)
-	if dbErr != nil {
-		logger.Get().Errorw("unable to retrieve notes in UpdatedNotes call", "err", dbErr)
+	notes, err := s.database.GetNotes(board)
+	if err != nil {
+		logger.Get().Errorw("unable to retrieve notes in UpdatedNotes call", "boardID", board, "err", err)
```
Schwehn42, line 134: What's the difference between `logger.Get()` and `logger.FromContext()`?

Schwehn42 (commented):
rest looks good, just have a couple questions:

server/src/logger/logger.go
mateo-ivc: I think this is somewhat specific to our deployment. As we use Kubernetes for deployment and only Docker for local development.
But I agree that this should be done differently by adding a flag.

server/src/logger/logger.go
mateo-ivc: You're right, that line makes no sense.
To be honest, I didn't even realize I added that line.
Will remove it!

server/src/services/notes/notes.go
mateo-ivc: Since we store the logger in go's context, we are able to automatically log data that was afterwards added to the context.
An example would be if we get a HTTP request we also store data like request-id inside the context.
`{"level":"info","ts":...,"caller":"logger/logger.go:84","msg":"Served","requestID":"microwave/V3oqqt2ri5-000004","method":"POST","proto":"HTTP/1.1","address":"\"http://localhost:8080/boards\"","lat":0.010658187,"status":201,"size":278}
`

server/src/logger/logger.go
mateo-ivc: If you mean the stacktrace, I wanted to disable the stacktrace to make the logs more readable for humans. Enabling the stacktrace leads to a very long and confusing log. But with the updated code, it's only disabled in production.

Schwehn42 (approved):
lgtm

PR #4556, opened 2024-10-16
fix: column header overlay

### LINKED ISSUE #4538: [Chrome] Colour Picker is hidden behind the Notes
### The problem

After adding notes if I try to change the colour of the notes , then the colour picker dropdown is getting hidden behind the notes

### Browser

Chrome

1. Create a board.
2. Add couple of notes.
3. Edit the board (from the 3-dot menu)
4. Select Colour-picker

### Screenshots

Screenshot 2024-10-13 at 1 05 50 PM

### Additional context

_No response_

Schwehn42:
## Description
Right now, the color picker isn't correctly rendered in Chrome.  
Reason for this is a wrong z-index for column header, which would lead to these problems.
The z-index was introduced in #4391, fixing another issue concerning the emoji overlay.
However, the value itself was less than the notes z-index.

The proposed changes here adjust the z-index of column headers to always be greater than the notes z-index, so all components are explicitly overlayed.

Resolves #4538.

## Changelog
- Added new z-index var `$column-header-z-index` which is greater than `$note-z-index`
- Applied  `$column-header-z-index` to column header
- Change z-indices in components related to the color picker; if not required removed entirely

## Checklist

### DIFF
--- a/src/components/ColorPicker/ColorPicker.scss
+++ b/src/components/ColorPicker/ColorPicker.scss
@@ -1,7 +1,6 @@
 @import "constants/style";
 
 .color-picker {
-  z-index: $menu-item-z-index;
   display: flex;
   flex-direction: column;
   justify-content: center;
--- a/src/components/Column/Column.scss
+++ b/src/components/Column/Column.scss
@@ -49,10 +49,11 @@
 }
 
 .column__header {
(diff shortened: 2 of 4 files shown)


mateo-ivc (approved):
lgtm 👍🏽

PR #4254, opened 2024-06-21
feat: template and sessions (framework/layout)

Schwehn42:
## Description

Resolves #4201 and #4253 👀 

## Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] The light- and dark-theme are both supported and tested
- [ ] The design was implemented and is responsive for all devices and screen sizes
- [ ] The application was tested in the most commonly used browsers (e.g. Chrome, Firefox, Safari)

PR #4584, opened 2024-10-24
fix: local user setting change

Schwehn42:
## Description

With the introduction of the newly refactored settings dialog (#4265), it was made possible to add settings screens outside a board. 

However, using this in other views, like the WIP template view (#4254) revealed that changing the user name or avatar would not be applied to the state until reloading the page.  

The reason for this is simple: We use the features `participants` and `auth` separately, both of which contain information about the name and avatar of a user. The difference lies in where the information is available: 
- `auth.user` is always available after the user has logged in
- `participants.self` is only available after joining a board.

The way these two change is also different:
- `auth.user` is initially retrieved after logging in and doesn't change.
- `participant.self` is updated when a server event is sent.

Thus, the solution is to apply the changes to the user (name and avatar specifically) so we aren't dependent on being in a board session. In places where user information is required, using `auth` instead of `participants` is required.

## Changelog
- Use user state from `auth` feature instead of `participants` in settings.
- When editing the user in the settings dialog, changes are applied instantly to `auth`.
  - add `auth` action `editUserOptimistically`
  - adjust `participants` thunk `editSelf`

## Checklist

### DIFF
--- a/src/components/SettingsDialog/Components/AvatarSettings.tsx
+++ b/src/components/SettingsDialog/Components/AvatarSettings.tsx
@@ -4,7 +4,6 @@ import {Avatar, generateRandomProps} from "components/Avatar";
 import {Fragment, useEffect, useState} from "react";
 import {useTranslation} from "react-i18next";
 import {useAppDispatch, useAppSelector} from "store";
-import {isEqual} from "underscore";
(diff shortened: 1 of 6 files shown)


laila-rin (approved):
lgtm :)

PR #4600, opened 2024-10-30
fix: admit longer session cookies by using a larger store

kedorlaomer:
## Description
Add an explicit cookie store with a bigger limit so that we can accept larger session cookies. Some OIDC providers require that.

## Changelog
- server/src/api/login.go: in `beginAuthProviderVerification`, set a cookie store

## Checklist

- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] The light- and dark-theme are both supported and tested
- [ x] The design was implemented and is responsive for all devices and screen sizes
- [ x] The application was tested in the most commonly used browsers (e.g. Chrome, Firefox, Safari)

## (Optional) Visual Changes
none

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "react-scripts": "^5.0.1",
     "redux-mock-store": "^1.5.4",
     "resize-observer-polyfill": "^1.5.1",
-    "sass": "^1.80.4",
+    "sass": "^1.80.5",
     "stylelint": "^16.10.0",
     "stylelint-config-standard": "^36.0.1",
     "ts-jest": "^29.2.5",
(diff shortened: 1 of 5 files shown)


brandstetterm:
Hey @kedorlaomer

brandstetterm:
u

kedorlaomer:
We were using Microsoft's OIDC provider. It was sending session cookies which were larger than the default size of 4k. The backend wouldn't accept the login. Therefore, we had to increase the limit in our own fork.

Note that this does not mean that 1e9 bytes of memory and/or storage are used. It is just a limit. The original example (which I'd need to dig out) used the largest positive integer 😸.

Schwehn42:
@kedorlaomer we can definitely increase the limit, I would prefer some logical value over an arbitrary one though. Something like 16k or 32k should be more than enough, shouldn't it?

kedorlaomer:
changed to 32k

Schwehn42 (approved):
LGTM. Here's the PR you were referring to btw: https://github.com/markbates/goth/pull/141

Note that before releasing, we will wrap this around a feature flag which is deactivated by default, as `NewFilesystemStore` is still an experimental feature. I'll tag you in the related PR. Thanks for your contribution!

PR #4601, opened 2024-10-30
refactor: access policy enum

Schwehn42:
## Description

Problem: Since the very beginning, `AccessPolicy` was an enum. This might look nice to use but actually causes some trouble, especially when trying to cast json data from an endpoint. Sure you can use `typeof keyof AccessPolicy` but that has other disadvantages.

So, this PR simply replaces the enum in favour of an union type.

Further reading: https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh

## Changelog

- change `AccessPolicy` from enum to union type
- adjust code that's impacted by the change

## Checklist


mateo-ivc (approved):
lgtm :shipit:

PR #4674, opened 2024-12-03
chore: docker improvements

Planlos5000:
## Description

Add healthchecks to the docker compose and add some default values that can be changed through the .env file.
Add the versions of each container to the .env file.
Update caddy configuration.

## Changelog
- Update docker compose with health checks and container names
- Update .env file for docker compose 
- Update caddy configuration to respond on localhost
- Fix documentation

## Checklist


deployment/.gitignore
```diff
@@ -1,2 +1,2 @@
-.env
-jwt.key
\ No newline at end of file
+**/.env
+**/jwt.key
```
HannesHil, line 2: please add line break
```suggestion
**/jwt.key

```

.gitignore
```diff
@@ -36,7 +36,6 @@ ui-debug.log
 
 # env files
 server/.env
```
HannesHil, line 38:
```suggestion
**/.env
```

HannesHil: use one **/.env only please

deployment/docker/docker-compose.yml
```diff
@@ -35,15 +35,22 @@ services:
       # SCRUMLR_SERVER_REDIS_HOST: "${REDIS_HOST}"
       # SCRUMLR_SERVER_REDIS_USERNAME: "${REDIS_USERNAME}"
       # SCRUMLR_SERVER_REDIS_PASSWORD: "${REDIS_PASSWORD}"
-    ports:
-      - "8080:8080"
```
HannesHil, line 39: good catch! But what happens if somebody wants to do API requests directly against the backend or is this not a thing?

deployment/docker/docker-compose.yml
```diff
@@ -35,15 +35,22 @@ services:
...
     depends_on:
-      - postgres
-      - nats
+      postgres:
+        condition: service_healthy
+      nats:
+        condition: service_started
+    healthcheck:
+      test: ["CMD-SHELL", "wget --no-verbose --tries=1 --spider http://localhost:8080/api/health || exit 1"]
```
HannesHil, line 44:
```suggestion
      test: ["CMD-SHELL", "wget --no-verbose --tries=1 --spider http://localhost:${SCRUMLR_SERVER_PORT}/api/health || exit 1"]
```
could be a thing

deployment/docker/Caddyfile
```diff
@@ -1,14 +1,14 @@
-0.0.0.0:80 {
-    log {
-        output stdout
-    }
-    @api {
-        path /api*
-    }
-    reverse_proxy @api scrumlr-backend:8080
+0.0.0.0:80, localhost:80, 127.0.0.1:80 {
```
HannesHil, line 1: was this done do be more verbose regarding the bindings because
`0.0.0.0:80`
AND
`:80`

would be enough to create the same listenings

deployment/docker/docker-compose.yml
Planlos5000: You can still call the api through caddy on e.g. <code>localhost/api/...</code>

deployment/docker/.env.example
```diff
@@ -1,51 +1,63 @@
...
+SCRUMLR_SHOW_LEGAL_DOCUMENTS=true
+## Analytics (Plausible)
+ANALYTICS_DATA_DOMAIN=
+ANALYTICS_SRC=
 SCRUMLR_CLARITY_ID=
+
+## Postgres
+POSTGRES_VERSION=17.2
+POSTGRES_USER=
+POSTGRES_PASSWORD=iefe4eiciequeDavolalai1ooheeFohRoh7Ahng9loofaeveiSa8pheV1kee8ooc
```
mateo-ivc, line 56: Password should be empty since its an .example file

deployment/docker/docker-compose.yml
```diff
@@ -2,7 +2,7 @@
 services:
   scrumlr-backend:
     restart: always
-    image: "ghcr.io/inovex/scrumlr.io/scrumlr-server:${SCRUMLR_BACKEND_VERSION:-3.10.0}"
+    image: "ghcr.io/inovex/scrumlr.io/scrumlr-server:${SCRUMLR_BACKEND_VERSION:-3.10.2}"
```
Schwehn42, line 5: here and in the other images, would it make sense to use `latest` as the fallback version if the specific env var isn't set? then we wouldn't need to adjust the version value with every release

Lennart01: latest would potentially pull development images.
We probably dont want that.

Schwehn42 (approved):
lgtm

PR #4756, opened 2025-01-07
add case to note reducer when voting gets updated

mateo-ivc:
## Description

When ending a voting the user expects to get its note sorted by votes from high to low
## Changelog

``notes/reducer.ts``
- added a case that checks if the `updateVoting` event has been called and updates the note state

## Checklist


Schwehn42 (approved):
LGTM, also tested with multiple columns and note stacks.

PR #4759, opened 2025-01-10
ci: Add action to clean packages

Planlos5000:
## Description

Add action to remove old, untagged packages. The action runs ever day from monday to friday at 7 am.

## Changelog

- add action to remove packages

## Checklist

### DIFF
--- a/.github/workflows/clean_packages.yml
+++ b/.github/workflows/clean_packages.yml
@@ -1,0 +1,29 @@
+---
+name: Remove old unused scrumlr packages
+
+on:
+  schedule:
+    - cron: "0 7 * * 1-5"
+
+jobs:
+  clean_scrumlr_server_packages:
+    name: "clean scrumlr server packages"
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/clean_packages.yml
brandstetterm, line 1: Please add a comment in which interval the packages are going to be deleted

.github/workflows/clean_packages.yml
```diff
@@ -0,0 +1,29 @@
+---
+name: Remove old unused scrumlr packages
+
+on:
+  schedule:
+    - cron: "0 7 * * 1-5"  # every monday through friday at 7 am
```
brandstetterm, line 6: Thank you for adding the interval as comments.
Out of curiosity: Wouldn't it be enough to run the workflow just once a week?

Planlos5000: The action only deletes 100 packages per run. Once the most of the untagged packages are deleted, we could switch to run once a week.

brandstetterm (approved):

PR #4760, opened 2025-01-10
test: mock system time for snowfall

Schwehn42:
## Description

In the months of December and January, snowfall is rendered including a setting in the `Appearance` component.
An accompanying test would compare the rendered component to a snapshot. However, since the test is reliant on time, the test will fail if the time's not right. Additionally, CI/CD runners don't necessarily synchronize their time with the real world, leading to more confusion. Hence, these tests mock the system time in order to achieve consistent results.

## Changelog

- `Appearance.test.tsx`
  - set mocked system time
  - add test to test both with and without snowfall
- `Appearance.test.tsx.snap`: update snapshot

### DIFF
--- a/src/components/SettingsDialog/Appearance/__tests__/Appearance.test.tsx
+++ b/src/components/SettingsDialog/Appearance/__tests__/Appearance.test.tsx
@@ -5,6 +5,8 @@ import getTestStore from "utils/test/getTestStore";
 import {Appearance} from "../Appearance";
 import {getByLabelText} from "@testing-library/dom";
 import i18n from "i18nTest";
+import {within} from "@testing-library/react";
+import {t} from "i18next";
 
 const createAppearance = () => (
   <Provider store={getTestStore()}>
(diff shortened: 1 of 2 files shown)


src/components/SettingsDialog/Appearance/__tests__/Appearance.test.tsx
brandstetterm, line 1: Thank you for creating some tests for it!
I'm not a big fan of snapshots. Instead of using them, we could check if a specific element is / isn't in the rendered output. I think that would be better, since we're not required to check the snapshot if the snowfall setting is / isn't in there.

Schwehn42: I think both strategies have their pros and cons, because snapshots have to be updated if anything around it changes, but they have the advantage you don't need to specifically select elements from the dom, where class names and the like can also change. So at the end of the day it doesn't really matter in terms of maintainability I guess. But I can still add some checks for the element

brandstetterm (approved):

PR #4761, opened 2025-01-10
chore: update homepage copyright year

Schwehn42:
## Description
Sets the year displayed in the homepage footer to the current year. Note this PR does not apply to the Landingpage used on scrumlr.io, ~~for that see the linked PR~~

## Changelog

- set copyright year to dynamic current year (2021 - `{{currentYear}}`)

### DIFF
--- a/src/i18n/de/translation.json
+++ b/src/i18n/de/translation.json
@@ -320,7 +320,7 @@
     "english": "Englisch"
   },
   "Homepage": {
-    "developers": "© 2021-2023, mit ❤️ von <inovex>inovex</inovex>",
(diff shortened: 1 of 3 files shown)


brandstetterm:
This won't work if the user has disabled JavaScript in his browser, right? @Schwehn42

Schwehn42:
@brandstetterm well the application as a whole won't load without javascript so it doesn't matter

brandstetterm:
> @brandstetterm well the application as a whole won't load without javascript so it doesn't matter

True, haven't even thought about that :D

brandstetterm (approved):

Issue #4736, opened 2024-12-30
serve static content under dedicated assets route


Schwehn42:
In order to allow for some Nginx configs, we should serve static content (`timer_finished.mp3`, `hotkeys.pdf`) under a dedicated subdirectory. Could be called `assets`.

_Originally posted by @kersten in https://github.com/inovex/scrumlr.io/discussions/4724_

PR #4775, opened 2025-01-14
refactor: static content assets directory

Schwehn42:
## Description

Resolves https://github.com/inovex/scrumlr.io/issues/4736

## Changelog

- Move `timer_finished.mp3` and `hotkeys.pdf` from `public` to `public/assets`
- Adjust path refs


nginx.conf
```diff
@@ -30,6 +30,11 @@ server {
       add_header Cache-Control "no-store, no-cache, must-revalidate";
     }
 
+   location /assets {
+      try_files $uri $uri/ =404;
+      add_header Cache-Control "public, max-age=3600";
```
brandstetterm, line 35: We could consider a higher cache duration, e.g. 1 day to 1 week, since these assets are likely to not change that frequently

Schwehn42: agreed 👍 @Planlos5000 and I were still looking into it since on the dev deployment, this location rule doesn't seem to work right now

Schwehn42:
Q: is `/assets` or `/static` better fitting?

brandstetterm:
> Q: is `/assets` or `/static` better fitting?

Fine with both

Schwehn42:
closing as this is probably not needed anymore right now

PR #4788, opened 2025-01-17
refactor: refactor event_filter

wischoepke:
## Description
Event filter parsing methods refactored into a more generic solution. A small step to reduce the complexity of the event filter and improve testability.

## Checklist

## (Optional) Visual Changes
No visual changes. Just a small refactoring


server/src/votes/service.go
```diff
@@ -0,0 +1,143 @@
...
+	v.ShowVotesOfOthers = voting.ShowVotesOfOthers
+	v.Status = voting.Status
+	v.VotingResults = getVotingWithResults(voting, votes)
+	return v
+}
+
+func (*Voting) Render(_ http.ResponseWriter, _ *http.Request) error {
+	return nil
+}
+
```
mateo-ivc, line 25: This method is not used and always returns nil.
Is there a specific purpose for this method?

mateo-ivc (approved):
lgtm :shipit:

PR #4799, opened 2025-01-22
fix: Stack View Animation

Resaki1:
## Description
Due to changes in the stack view layout, the animation that appears when navigating between notes was a little broken. This has been fixed. Includes minor code improvements.

## Changelog
There were three issues with the animation:
1. The width of the container was limited with overflow: hidden, meaning the notes were cut off left and right during the animation.
2. When navigating between stacks with a different number of stacked notes, the height of the container would change instantly at the start of the animation, cutting the notes off at the bottom.
3. When navigating between columns, the previous stack would instantly disappear.

These issues have been fixed, as shown below:

### Before
https://github.com/user-attachments/assets/74f60b9c-542f-41a3-a4cf-14479ec5f269

### After
https://github.com/user-attachments/assets/e044806f-6fe2-4371-9920-183db6cc3bc1


Schwehn42 (commented):
throws an error in **Firefox 134.0.2** when trying to transition between columns:

Firefox column transition error

src/components/StackNavigation/StackNavigation.tsx
```diff
@@ -10,8 +10,8 @@ import {StackNavigationDots} from "./Dots/StackNavigationDots";
 interface StackNavigationProps {
   stacks: Note[];
   currentStack: string;
-  prevColumnStack?: string;
-  nextColumnStack?: string;
```
brandstetterm, line 14: Why?

brandstetterm: Wouldn't `null` be better than `undefined`?

Resaki1: can't remember, changed it back

Schwehn42 (commented):
Found a bug; In mobile view, a stack with multiple notes is no longer scrollable

Resaki1:
> Found a bug; In mobile view, a stack with multiple notes is no longer scrollable

thanks, should be fixed now

Schwehn42 (approved):
works now, lgtm 👍

PR #4854, opened 2025-02-17
refactor: overwritable settings config

Schwehn42:
## Description

The last couple of refactors concerning the settings dialog (https://github.com/inovex/scrumlr.io/pull/4265) were a great upgrade allowing for configuration of settings items, including name, color etc. as well as when these are enabled (settings were only used on boards before, now they're also available for every other route).

While this is quite convenient, the menu items themselves are still constant, with the only possibility to toggle which ones are displayed, but not for the properties themselves.

Hence, with this PR it's now also possible to override settings at runtime by passing an `override` prop to the `SettingsDialog` component.

## Example
```tsx
<Route path="settings" element={<SettingsDialog 
  enabledMenuItems={{...ENABLE_ALL, feedback: feedbackEnabled}}
  overwrite={{appearance: {color: "goal-green"}}} />}
>
  <Route path="appearance" element={<Appearance />} />
  {/* other settings routes here */}
</Route>
```

## Changelog
- add new prop `?overwrite` of new type `MenuItemConfigOverride` to `SettingsDialog`
- add typesafe function to merge with `MENU_ITEM_CONFIG`
- internal changes to `SettingsDialog` where the merged config is used as state in order to render the settings

## Checklist


Schwehn42:
closed since this currently doesn't seem required

PR #4883, opened 2025-02-26
Feature/non anonymous voting

Yaazizi:
Database Changes
	•	Added is_anonymous column to the votings table (21_add_anonymous_votings_.up.sql).
	•	Migration rollback support: Dropping is_anonymous column when rolling back (21_add_anonymous_voting.down.sql).

Backend Changes
	•	Updated Voting model (server/src/database/votings.go)
	•	Added isAnonymous field in Voting and VotingInsert.
	•	Modified CreateVoting method to include isAnonymous.
	•	Modified Voting Service (server/src/services/votings/votings.go)
	•	CreateVoting method now includes isAnonymous.
	•	Updated Data Transfer Objects (server/src/votes/dto.go)
	•	Added isAnonymous field to VotingCreateRequest and Voting.
	•	Updated Votes Service (server/src/votes/service.go)
	•	Voting.From() now includes isAnonymous from database.
	•	Ensured non-anonymous votes display user names.

Frontend Changes
	•	Updated Voting UI (src/components/Infobar/Infobar.tsx)
	•	Displays an icon ( Hidden or Visible) indicating if voting is anonymous or not.
	•	Modified Vote Buttons (src/components/Votes/VoteButtons/RemoveVoteButton.tsx)
	•	Now shows participant names when voting is non-anonymous.
	•	Updated Voting Dialog (src/components/VotingDialog/VotingDialog.tsx)
	•	Added a toggle button to switch between anonymous and non-anonymous voting.
	•	Updated Redux Store (src/store/features/votings/types.ts)
	•	Added isAnonymous to the Voting state.
	•	Updated Test Utilities (src/utils/test/getTestApplicationState.ts & src/utils/test/getTestVoting.ts)
	•	Adjusted mock data to support anonymous voting.


src/components/Votes/VoteButtons/RemoveVoteButton.tsx
```diff
@@ -43,7 +46,7 @@ export const RemoveVoteButton: FC<RemoveVoteProps> = ({noteId, disabled, numberO
         setDoBump(false);
       }}
       dataTooltipId="scrumlr-tooltip"
-      dataTooltipContent={disabled ? t("Votes.VotesOnNote", {count: numberOfVotes}) : t("Votes.RemoveVote")}
+      dataTooltipContent={isAnonymous ? t("Votes.VotesOnNote", {count: numberOfVotes}) : participantNames}
```
mateo-ivc, line 49: image

Wrong usernames are being displayed.
Not even sure if it's necessary to display names during voting phase

server/src/services/votings/votings.go
```diff
@@ -50,6 +50,7 @@ func (s *VotingService) Create(ctx context.Context, body votes.VotingCreateReque
 		VoteLimit:          body.VoteLimit,
 		AllowMultipleVotes: body.AllowMultipleVotes,
 		ShowVotesOfOthers:  body.ShowVotesOfOthers,
+		IsAnonymous:        body.IsAnonymous,
```
mateo-ivc, line 53: Those are basically the same.
Consider removing ShowVotesOfOthers since it's not beeing used

mateo-ivc (changes requested):
IsAnonymouse and ShowVotesOfOthers are basically the same. Chose on of them and remove the other

server/src/services/votings/votings.go
Schwehn42: is it though? seems to me one is to filter out certain votes in order to only see votes by yourself, while the other option is whether you can see the names of who has cast votes in general.

though the former could be done using a pure appearance option and the latter has to be set before starting a voting

mateo-ivc: > though the former could be done using a pure appearance option and the latter has to be set before starting a voting

This would make more sense imo. Also we currently don't have the option to toggle ShowVotesOfOthers in the frontend right?

server/src/api/votings_test.go
Schwehn42, line 1: we need something like husky to format the code uniformly for backend

src/components/Infobar/Infobar.tsx
```diff
@@ -27,6 +28,16 @@ export const InfoBar = () => {
...
+      {state.activeVoting && state.isVotingAnonymous && (
+        <span data-tooltip-id="info-bar__tooltip" data-tooltip-content={t("InfoBar.VotingIsAnonymous")}>
+          <Hidden />
+        </span>
+      )}
+      {state.activeVoting && !state.isVotingAnonymous && (
+        <span data-tooltip-id="info-bar__tooltip" data-tooltip-content={t("InfoBar.VotingIsNotAnonymous")}>
+          <Visible />
+        </span>
+      )}
```
Schwehn42, line 40: the `gap` of `.info-bar` should be 8px according to design

Also, make sure to set the `height` of the span (give it a class first) to `$icon--large`, which is equal to `32px`, so it's vertically centered

src/components/Votes/Votes.tsx
```diff
@@ -32,6 +32,19 @@ export const Votes: FC<VotesProps> = (props) => {
...
+      return {participantsNames: "", isAnonymous: true};
+    }
+    const others = state.participants?.others ?? [];
+    const participants = [state.participants?.self, ...others];
+    const names = state.votings.past[0]?.votes?.votesPerNote[props.noteId]?.userVotes?.map(
+      (userVote) => participants.find((p) => p?.user.id === userVote.id)?.user.name as string
+    ) as string[];
+    return {participantsNames: names?.join(", "), isAnonymous: false};
+  })!;
+
```
Schwehn42, line 47: I think it would be good to first add a small comment explaining what's going on here, and maybe also refactor this a bit. For instance, `state.votings?.past[0]` should be a const var named `lastVoting` or something along those lines.

Also, if possible we want to avoid type assertions using `!` and `as string`, because if a user isn't found `undefined` will be treated like a string. Maybe instead try [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) or something like that instead?

src/store/features/votes/types.ts
```diff
@@ -1,6 +1,7 @@
 export interface Vote {
   voting: string;
   note: string;
+  user?: string;
```
Schwehn42, line 4: maybe a more type safe approach would be something like this:
```ts
type VoteBase = {
  voting: string;
  note: string;
  isAnonymous: true; // user is not available
};

type VoteWithUser = {
  voting: string;
  note: string;
  isAnonymous: false; // user is available
  user: string; // user is required
};

type Vote = VoteBase | VoteWithUser; // Union type

```

That way `user` can only be set if `isAnonymous` is false, retaining type safety.

Schwehn42 (commented):
I want to address the elephant in the room: this is not anonymous. You can still see who voted what in the network tab and draw a conclusion who is who quite easily.

So if we want actual anonymous votings, changes that are more complex are required

src/store/features/votes/types.ts
Yaazizi: please see the commit

src/components/Votes/Votes.tsx
Yaazizi: please see the commit

src/components/Infobar/Infobar.tsx
Yaazizi: please see the commit

src/store/features/votes/types.ts
Schwehn42: well this isn't quite what i meant, you've added the new types but aren't using it anywhere ^^ the problem is that the types for votings are insanely bad structured, with a million optional subtypes (not your fault, it's always been that way unfortunately)

I think for now you can revert these type changes, and with the votes refactor we can figure a better type system out together where votings, votes and respective users are easier and safer to handle.

src/components/Votes/Votes.tsx
```diff
@@ -32,6 +34,29 @@ export const Votes: FC<VotesProps> = (props) => {
...
+    const userVotes = votesPerNote[props.noteId]?.userVotes ?? [];
+    // Retrieve every participant, including oneself
+    const others = state.participants?.others ?? [];
+    const selfParticipant = state.participants?.self;
+    const participants = selfParticipant ? [selfParticipant, ...others] : others;
+    // Extract participant names based on votes for the given noteId
+    const names = userVotes
+      .map((userVote) => participants.find((p) => p?.user.id === userVote.id)?.user.name ?? null)
+      .filter((name): name is string => name !== null) // Ensure only valid strings remain
+      .join(", ");
```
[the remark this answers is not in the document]
Yaazizi: i have made the changes in this commit:

[ref_Sort participant names alphabetically](https://github.com/inovex/scrumlr.io/pull/4883/commits/9f700c83137b004a9a24a0111824cd9fe95a6324)

src/components/Votes/Votes.tsx
```diff
@@ -32,6 +34,29 @@ export const Votes: FC<VotesProps> = (props) => {
...
+    if (!lastVoting || lastVoting.isAnonymous) {
+      return {participantsNames: "", isAnonymous: true};
+    }
+    // Extract votes per note safely
+    const votesPerNote = lastVoting.votes?.votesPerNote ?? {};
+    const userVotes = votesPerNote[props.noteId]?.userVotes ?? [];
+    // Retrieve every participant, including oneself
+    const others = state.participants?.others ?? [];
+    const selfParticipant = state.participants?.self;
+    const participants = selfParticipant ? [selfParticipant, ...others] : others;
```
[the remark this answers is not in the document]
Yaazizi: ref_simplify participant merging logic

Schwehn42 (commented):
hi, I found one more small bug: A stacked displays the aggregated vote count of parent and children, but does not display all names in case of an non-anonymous voting, see screenshots:

user ‚work it‘ is part of the voting
Screenshot 2025-04-15 at 13 06 37

but not displayed on the parent note
Screenshot 2025-04-15 at 13 06 51

do note when implementing this, make sure to filter out duplicates

Yaazizi:
> hi, I found one more small bug: A stacked displays the aggregated vote count of parent and children, but does not display all names in case of an non-anonymous voting, see screenshots:
> 
> user ‚work it‘ is part of the voting Screenshot 2025-04-15 at 13 06 37
> 
> but not displayed on the parent note Screenshot 2025-04-15 at 13 06 51
> 
> do note when implementing this, make sure to filter out duplicates

Yaazizi:
> fix (show all voters)

Hey, I have made the required changes see [fix (show all voters)](https://github.com/inovex/scrumlr.io/pull/4883/commits/50654efeb0fbe93c7109eadb8a7c5b4e0a61a5b9)

src/components/Votes/Votes.tsx
```diff
@@ -32,6 +32,44 @@ export const Votes: FC<VotesProps> = (props) => {
...
+        .reduce(
+          (acc, curr) => {
+            const childUserVotes = (votesPerNote[curr.id]?.userVotes ?? []).map((v: any) => ({id: v.id}));
+            return acc.concat(childUserVotes);
+          },
+          [] as Array<{id: string}>
+        );
+      userVotes = userVotes.concat(childVotes);
+    }
+    const uniqueUserVotes = _.uniq(userVotes, (vote) => vote.id);
```
Schwehn42, line 57: Could you add a little comment explaining what this part does (display unique user names for stacked notes)

and also fix the two linter errors regarding `any` type

Schwehn42 (commented):
I'm assuming you closed the PR on accident, I reopened it and added one small nitpick, after that it should be fine :)

src/components/Votes/Votes.tsx
Yaazizi: i have add some changes :)

Yaazizi:
> hi, I found one more small bug: A stacked displays the aggregated vote count of parent and children, but does not display all names in case of an non-anonymous voting, see screenshots:
> 
> user ‚work it‘ is part of the voting Screenshot 2025-04-15 at 13 06 37
> 
> but not displayed on the parent note Screenshot 2025-04-15 at 13 06 51
> 
> do note when implementing this, make sure to filter out duplicates

Schwehn42 (approved):
LGTM 🥳

mateo-ivc (approved):

PR #4886, opened 2025-02-27
feat: linkify note dialog note content

### LINKED ISSUE #4885: Inline URLs in detail view
As of now, URLs are rendered on the board. We use [Linkify](https://linkify.js.org/docs/linkify-react.html) for this.

In the stack view, this hasn't been implemented yet, mainly due to the fact that `TextAreas` don't support html content, but only text.

AC:
- make inline URLs in stack possible
- should be clickable (new tab)
- should be editable (if note is editable for that person)
- should be truncated if too long if viewing stack, and full length if editing

Schwehn42:
## Description

Resolves #4885

Makes URLs visible and clickable in stack view, as is already the case for board view.  
URLs should be editable and clickable after saving the changes again.

## Change log

some refactoring was required to get this to work
- Note Content: replace `TextArea` with contentEdtiable `div`
- conditional content (linkified or pure)
- URL style mixin
- emoji suggestions adjustments (wip)

## TODOs
- cursor location when clicking div, changing to edit mode
- emoji autocomplete (probably need to mock change event)

## Checklist

### DIFF
--- a/src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
+++ b/src/components/NoteDialogComponents/NoteDialogNoteComponents/NoteDialogNoteContent.tsx
@@ -141,6 +141,7 @@ export const NoteDialogNoteContent: FC<NoteDialogNoteContentProps> = ({noteId, a
       tabIndex={0}
       onFocus={onFocus}
       onBlur={onBlur}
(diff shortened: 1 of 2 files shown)


Schwehn42:
closing this: it might be better to just add inline markdown, similar to what github does. using linkify seems to be a little troublesome

PR #4898, opened 2025-03-04
Ga sec tools

Yaazizi:
the test  include two   gosec and go critic  (for security and quality)


docs/yarn.lock
Schwehn42, line 1: we use yarn over npm, please revert this file and use `yarn install`

PR #4949, opened 2025-03-25
fix: require auth on legacy new board

Schwehn42:
## Description

Users were able to go to the legacy template view (if enabled in env) without being logged in. This caused an error when trying to create a new board.

## Changelog

- `Router.tsx`->`/new` route element: Wrap `LegacyNewBoard` with `RequireAuthentication`


laila-rin (approved):
lgtm :)

PR #4960, opened 2025-03-28
refactore notes

mateo-ivc:
## Checklist


server/src/main.go
```diff
@@ -371,7 +372,7 @@ func run(c *cli.Context) error {
 	boardSessionService := boards.NewBoardSessionService(dbConnection, rt)
 	votingService := votings.NewVotingService(dbConnection, rt)
 	userService := users.NewUserService(dbConnection, rt)
-	noteService := notes.NewNoteService(dbConnection, rt)
+	noteService := notes2.NewNoteService(dbConnection, rt)
```
Planlos5000, line 375: The service needs a `NotesDatabase` and here it gets the old database struct.
Consider using the `initialize` package and add there a `InitializeNoteService` method that creates the `NotesDatabase` and then the service

server/src/main.go
```diff
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	notes2 "scrumlr.io/server/notes"
```
Planlos5000, line 9: Remove the old note package in line 25

server/src/notes/database.go
```diff
@@ -1,63 +1,28 @@
...
-	Position      *NoteUpdatePosition `bun:",embed"`
-}
-
-type NoteUpdatePosition struct {
-	Column uuid.UUID
-	Rank   int
-	Stack  uuid.NullUUID
-}
+func NewNotesDatabase(database *bun.DB) *DB {
+	db := DB{db: database}
```
Planlos5000, line 19: We should agree on one method to create new objects. Now we have at least three methods to create new Object: `o = Object{}; return &o `, `o = &Object{}; return o` and `o := new(Object); return o`

server/src/notes/service.go
```diff
@@ -1,108 +1,216 @@
...
+  ImportNote(insert NoteImport) (NoteDB, error)
+  GetNote(id uuid.UUID) (NoteDB, error)
+  GetNotes(board uuid.UUID, columns ...uuid.UUID) ([]NoteDB, error)
+  UpdateNote(caller uuid.UUID, update NoteUpdate) (NoteDB, error)
+  DeleteNote(caller uuid.UUID, board uuid.UUID, id uuid.UUID, deleteStack bool) error
+  GetChildNotes(parentNote uuid.UUID) ([]NoteDB, error)
+}
 
-	Edited bool `json:"edited"`
+func NewNoteService(db NotesDatabase) NotesService {
```
Planlos5000, line 31: The broker is missing

server/src/notes/service_test.go
```diff
@@ -122,3 +125,215 @@ func buildNote(authorId uuid.UUID, columnId uuid.UUID) *Note {
...
+
+	req := NoteCreateRequest{
+		Column: uuid.UUID{},
+		Text:   "Test content",
+		Board:  uuid.UUID{},
+		User:   uuid.UUID{},
+	}
+	noteDB := makeTestNoteDB(uuid.New())
+
+	mockDB.On("CreateNote", mock.Anything).Return(noteDB, nil)
```
Planlos5000, line 169: Use the `mockDB.EXPECT().CreateNote().Return()`. This does not based on the method name as a string and will raise errors when writing the tests.

server/src/notes/service_test.go
```diff
@@ -43,7 +46,7 @@ func TestShouldShowNotesBecauseAuthorIdIsEqual(t *testing.T) {
 func TestMapping(t *testing.T) {
 	databaseNote := buildDatabaseNote()
 
-	note := Notes([]database.Note{databaseNote})[0]
+	note := Notes([]NoteDB{databaseNote})[0]
```
Planlos5000, line 49: Maybe move the Tests for the mapping, unamrshall and NoteSlice into a `dto_test` file

Planlos5000 (changes requested):
There is a test file in `services/notes`. What is up with this file?
Could you add the old database tests?

server/src/api/feedback.go
```diff
@@ -52,11 +24,11 @@ func (s *Server) createFeedback(w http.ResponseWriter, r *http.Request) {
 		str := "/"
 		body.Text = &str
 	}
-	if body.Type == FeedbackTypeFeatureRequest && *body.Text == "/" {
+	if body.Type == feedback.FeedbackTypeFeatureRequest && *body.Text == "/" {
```
wischoepke, line 27: Wenn möglich die Konstanten FeedbackTypeBugReport etc noch umbenennen. Die befinden sich schon im Feedback Package da ist es nicht notwendig das präfix feedback vorne dran zu haben. Beispiel:
FeedbackTypeBugReport -> BugReportType

server/src/notes/database.go
mateo-ivc: Agreed! What do you prefer here? I am fine with everything

Planlos5000: I'm also fine with all three. 
I would suggest to use the `o = new(Object)` because it is mostly used in this project.

Is there a way to enforce one style? Linter?

PR #5005, opened 2025-04-11
refactor: add board reaction package

Planlos5000:
## Description

Add a board reaction package

## Changelog

- Add board reaction package
- Add tests for board reaction package

## Checklist


server/src/boardreactions/reaction_test.go
```diff
@@ -0,0 +1,15 @@
+package boardreactions
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUnmarshalReaction(t *testing.T) {
```
mateo-ivc, line 9: We should clarify whether the use of “testify/suits” makes sense in our case.

Since I use it and you don't, we should come to a unified decision.

PR #5043, opened 2025-04-29
refactor: Add columns package

Planlos5000:
## Description

Adds a columns package and tests for the columns package

## Changelog

- Add columns package
- Add tests for the columns package

## Checklist


server/src/columns/service.go
```diff
@@ -1,74 +1,197 @@
...
+}
 
-	// The column id.
-	ID uuid.UUID `json:"id"`
+func NewColumnService(db ColumnDatabase, rt *realtime.Broker, noteService notes.NotesService, votingService voting.VotingService) ColumnService {
+	service := new(Service)
+	service.database = db
+	service.realtime = rt
+	service.noteService = noteService
+	service.votingService = votingService
```
mateo-ivc, line 38: I thought the idea was to move everything that would have needed an other service into the api layer

mateo-ivc (approved):

server/src/columns/service.go
Planlos5000: This was only the idea if it was not possible to use the needed service e.g. cycle import.

PR #5140, opened 2025-06-10
Feat/unleash integration

Yaazizi:
Frontend (scrumlr-frontend) requires:
	        REACT_APP_UNLEASH_FRONTEND_URL
	        REACT_APP_UNLEASH_FRONTEND_TOKEN

Backend (scrumlr-backend) requires:
	        SCRUMLR_UNLEASH_BACKEND_URL
                 SCRUMLR_UNLEASH_BACKEND_TOKEN

Secret Management
		Secrets are managed using [Sealed Secrets]in the GitLab infrastructure repo.Tokens are also stored in Bitwarden.


.env
```diff
@@ -1,3 +1,9 @@
 PUBLIC_URL=/
 REACT_APP_PUBLIC_URL=/
-REACT_APP_VERSION=$npm_package_version  
+REACT_APP_VERSION=$npm_package_version
+
+REACT_APP_UNLEASH_FRONTEND_TOKEN=default:development.15302e96feac7443a5cc159c9e722eab755da21addf09fe2a8246937
+REACT_APP_UNLEASH_FRONTEND_URL=https://unleash.dev.scrumlr.fra.ics.inovex.io/api/frontend/
+
+SCRUMLR_UNLEASH_BACKEND_TOKEN=default:development.c084c0095aea0f07aca45341142f8b80199c908d1298c0527ba894fc
+SCRUMLR_UNLEASH_BACKEND_URL=https://unleash.dev.scrumlr.fra.ics.inovex.io/api
```
Schwehn42, line 9: `.env` is checked in! tokens are sensitive and should never be viewable in git. I suggest regenerating the tokens, removing this and using the env vars like any other. the setup with the deployment vars looks good, so you just need to use them now.

server/src/api/router.go
```diff
@@ -225,6 +230,12 @@ func (s *Server) protectedRoutes(r chi.Router) {
 	})
 }
 
+func (s *Server) getFeatureFlag(w http.ResponseWriter, r *http.Request) {
+	flagName := chi.URLParam(r, "name")
+	status := unleash.IsEnabled(flagName)
+	render.JSON(w, r, map[string]bool{"enabled": status})
+}
+
```
Schwehn42, line 238: I understand correctly that this is just for testing purposes for the time being?

server/src/database/migrations/sql/21_add_anonymous_voting.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE IF EXISTS votings DROP COLUMN IF EXISTS is_anonymous;
```
Schwehn42, line 1: why are all these files from https://github.com/inovex/scrumlr.io/pull/4883 in this PR?

server/src/main.go
```diff
@@ -23,12 +24,44 @@ import (
...
 	"scrumlr.io/server/services/boards"
 	"scrumlr.io/server/services/feedback"
+	"scrumlr.io/server/services/health"
 	"scrumlr.io/server/services/notes"
 	"scrumlr.io/server/services/reactions"
 	"scrumlr.io/server/services/users"
 	"scrumlr.io/server/services/votings"
 )
 
+func init() {
```
Schwehn42, line 34: what exactly is `init` initializing? the function name should reflect that

server/src/main.go
```diff
@@ -23,12 +24,44 @@ import (
...
+			}
+			return
+		}
+		// Abbruch, wenn wir im Wurzel-Verzeichnis angekommen sind
+		if parent := filepath.Dir(dir); parent == dir {
+			break
+		}
+	}
+	log.Println(".env nicht gefunden – Umgebungsvariablen fehlen möglicherweise")
+}
```
Schwehn42, line 63: we use https://github.com/urfave/cli-altsrc for reading the env file. all of this can be removed, look into the `main` function how it can be replaced with just
```go
if c.String("SCRUMLR_UNLEASH_BACKEND_URL") != ""
```

server/src/main.go
```diff
@@ -216,7 +249,7 @@ func main() {
 			altsrc.NewStringFlag(&cli.StringFlag{
 				Name:    "auth-oidc-user-name-scope",
 				EnvVars: []string{"SCRUMLR_AUTH_OIDC_USER_NAME_SCOPE"},
-				Usage:   "JWT claim to request for the user name",
+				Usage:   "JWT claim to request für den user name",
```
Schwehn42, line 252: keep it English

server/src/main.go
```diff
@@ -245,7 +278,7 @@ func main() {
 			&cli.StringFlag{
 				Name:     "config",
 				EnvVars:  []string{"SCRUMLR_CONFIG_PATH"},
-				Usage:    "TOML `filepath` to be loaded ",
+				Usage:    "TOML `filepath` to be geladen ",
```
Schwehn42, line 281: keep it English

server/src/main.go
```diff
@@ -268,7 +301,29 @@ func run(c *cli.Context) error {
...
+		unleash.WithEnvironment("development"),
+	); err != nil {
+		return fmt.Errorf("failed to initialize Unleash: %w", err)
+	}
+
+	unleash.WaitForReady()
+	if unleash.IsEnabled("enable-welcome-message") {
+		logger.Get().Info("Feature 'enable-welcome-message' ist ENABLED!")
+	} else {
+		logger.Get().Info("Feature 'enable-welcome-message' ist DISABLED!")
```
Schwehn42, line 326: why is this nested here? this should probably be inside the `init` function you added earlier in this file.

server/src/services/votings/votings.go
```diff
@@ -50,6 +49,7 @@ func (s *VotingService) Create(ctx context.Context, body votes.VotingCreateReque
 		VoteLimit:          body.VoteLimit,
 		AllowMultipleVotes: body.AllowMultipleVotes,
 		ShowVotesOfOthers:  body.ShowVotesOfOthers,
+		IsAnonymous:        body.IsAnonymous,
```
Schwehn42, line 52: again, I don't understand why all these changes from #4883 are here. did you forget to merge the changes from main into your branch?

src/index.tsx
```diff
@@ -18,6 +20,14 @@ import {SHOW_LEGAL_DOCUMENTS, ANALYTICS_DATA_DOMAIN, ANALYTICS_SRC, CLARITY_ID}
 import {initAuth} from "./store/features";
 import "react-tooltip/dist/react-tooltip.css";
 
+const unleash = new UnleashClient({
+  url: process.env.REACT_APP_UNLEASH_FRONTEND_URL!,
+  clientKey: process.env.REACT_APP_UNLEASH_FRONTEND_TOKEN!,
+  appName: "scrumlr-frontend",
+  environment: "development",
```
Schwehn42, line 27: need two things here:
1. fallback if unleash isn't set  (should be optional to use imo, or we need to at least consider that)
2. development / production mode (for this purpose multiple `.env` can be used)

Schwehn42 (changes requested):
hey, thanks for the PR. there are quite a few issues to be addressed; the most prominent ones being:
1. the branch is probably split from your #4883 dev branch and hasn't been updated in a while. make sure to merge the current `main` into your branch.
2. env var should be
    1. secret (`.env` is checked in)
    2. accessed using https://github.com/urfave/cli-altsrc in the backend
    3. have a fallback / be checked if exist / be optional
3. when using AI in German, make sure not to accidentally translate our code to German.
4. make sure to make your IDE use the file `.editorconfig` for uniform formatting.

.env
Yaazizi: Alright, I have checked the .gitignore file — it includes the .env files in the service and deployment folders, but not in the root directory. I will work on it and add a new tokens

server/src/api/router.go
Yaazizi: Yes, I wrote this for the test backend in case someone wants to test it.

Schwehn42:
@Lennart01 why the force push, what happened to the branch 😭

Schwehn42:
see #5167

PR #5178, opened 2025-06-20
Refactoring

mateo-ivc:
## Description

Structural changes to backend code -> Moving slowly away from a layered architecture (api, service, database) we are now implementing a component based architecture while still keeping the three layers for each package.  
 
## Changelog

Legend: A - Added, M - Moved, D - Deleted, E - Edited
List of components: boards, boardreactions, boardtemplates, columns, columtemplates, notes, reactions, sessionrequests, sessions, votings (includes votes)
`service/`
- M&E .../* -> Moved all items into a designated component package

`database/`
- M .../`<component>`.go -> Moved all items into their designated component package
- E .../`<component>`_test.go 

`api/`
- E .../`<component>`.go

`src/`
- E .../main.go -> Initializing new services
- A .../.mockery.yaml

`src/initialize`
- A .../* initialization methods for all services 

`common/`
- M .../dto.go -> Moved all items into their designated component package
- A .../models.go -> BunModels to evade cyclic import

`<component/>`
- A .../api.go -> currently only interface for service-layer, later implement api logic here 
- A .../database_dto.go -> seperation between dto`s for web and database
- A .../mock_`<componentService>`.go -> mocks for testing
- A .../`<service>`_test.go 
- A .../helpers.go -> Not for all components, includes helper functions 
## Checklist


server/src/api/boards.go
```diff
@@ -430,22 +434,22 @@ func (s *Server) importBoard(w http.ResponseWriter, r *http.Request) {
 
 	body.Board.Owner = owner
 
-	columns := make([]dto.ColumnRequest, 0, len(body.Notes))
+	importColumns := make([]columns.ColumnRequest, 0, len(body.Notes))
```
Planlos5000: Why is it `len(body.Notes)` when it is loops through the columns?
The array does not need to be initialized first

server/src/boardtemplates/database.go
```diff
@@ -0,0 +1,135 @@
...
+		return DatabaseBoardTemplate{}, err
+	}
+
+	return tBoard, err
+}
+
+func (db *DB) GetAll(user uuid.UUID) ([]DatabaseBoardTemplateFull, error) {
+	var tBoards []DatabaseBoardTemplate
+
+	// note that if no templates match, nil is returned
```
Planlos5000: Change the methods so that `nil` will never be returned

server/src/sessionrequests/database.go
```diff
@@ -0,0 +1,122 @@
...
+
+func NewSessionRequestDatabase(database *bun.DB) SessionRequestDatabase {
+	db := new(SessionRequestDB)
+	db.db = database
+
+	return db
+}
+
+// CreateBoardSessionRequest creates a new board session request with the state 'PENDING' and does nothing if it already exists
+func (database *SessionRequestDB) Create(request DatabaseBoardSessionRequestInsert) (DatabaseBoardSessionRequest, error) {
```
Planlos5000, line 25: This is not a create, its an upsert method. Check if the update method is needed

Planlos5000 (changes requested):
From the review with @wischoepke 
Maybe rewrite the `UnmarschalJSON` methods that unmarschal to a const string like the `FeedbackType`. Maybe make the method generic

server/src/boardtemplates/database.go
Planlos5000: The comment is wrong. `tBoards` is initialized with an empty slice and it is still an empty list after the query is executed and no results match. Because `temaplates` is also initialized with an empty slice, an empty slice is returned and not `nil`

server/src/sessionrequests/database.go
Planlos5000: I tested this very short. The `Create` and `Update` are slightly different and are not doing the exact same thing. I would prefer to change this into an `Upsert` once the database tests are refactored

Planlos5000 (approved):

PR #5210, opened 2025-07-02
fix: dependabot spam

brandstetterm:


Schwehn42:
context please, what exactly is the issue right now and how does this attempt to fix it

brandstetterm:
There are too many pull requests opened by dependabot
I want to minimize them by hopefully having one group PR for frontend dependencies and one group PR for backend dependencies

Planlos5000:
I don't think this is an issue. I prefer to have all dependency updates as a separate PR.
If one dependency breaks the build you have to manually look for the dependency and revert it. If they are all separate you can see which dependency breaks the build.
I think we need a better strategy to merge these PR

Schwehn42:
I see, I for one would definitely welcome grouped deps PRs, but also as @Planlos5000 said we'd need to be able to circumvent the problem if one single dependency fails the pipeline for multiple deps

PR #5219, opened 2025-07-08
feat: added favourit star button to all template cards (issue 5218)

jeancloudvanlang:
## Description
Added favourite star button to recommended template cards (see custom cards)

## Changelog
* added favourite buttons
* added sort of favourites first

## Checklist

- [x] All TemplateCards have favourite button
- [x] Recommended favourtite star buttons are stored locally (in contrast to Custom ones)
- [x] favourite Template Cards appear at first in lists


src/routes/Boards/Templates/Templates.tsx
```diff
@@ -80,7 +82,14 @@ export const Templates = () => {
     dispatch(getTemplates());
   }, [dispatch]);
 
-  const toggleFavourite = (templateId: string, favourite: boolean) => dispatch(setTemplateFavourite({id: templateId, favourite: !favourite}));
+  const recommendedFavourites = useAppSelector((state) => state.recommendedFavourites);
+  const toggleFavourite = (templateId: string, favourite: boolean, type: "RECOMMENDED" | "CUSTOM") => {
+    if (type === "RECOMMENDED") {
+      dispatch(toggleRecommendedFavourite(templateId));
+    } else {
+      dispatch(setTemplateFavourite({id: templateId, favourite: !favourite}));
```
Schwehn42: seems like the backend flips the boolean again, leading to no change when attempting to favourite a custom template

src/routes/Boards/Templates/Templates.tsx
```diff
@@ -184,37 +193,42 @@ export const Templates = () => {
...
+            ).map((templateFull: TemplateWithColumns) => (
+              <TemplateCard
+                templateType="RECOMMENDED"
+                template={{
+                  ...mergeTemplateWithColumns(templateFull.template, templateFull.columns),
+                  template: {
+                    ...templateFull.template,
+                    favourite: recommendedFavourites.includes(templateFull.template.id),
+                  },
+                }}
```
Schwehn42: instead of merging the favourite state here, one can also change the type of `ImportReducedTemplateWithColumns` by adding `favourite` (default `false`), and adjust `convertJsonToFullTemplates()` to update the state respectively. that would be a little cleaner I think

src/store/features/templates/recommendedFavourite.ts
Schwehn42, line 1: an extra reducer just for the favourite state of recommended templates is a bit awkward imo, I think it would be preferable to make the templates reducer support recommended templates. 

here's how this could be achieved:
1. type `Template` gets the type `TemplateType` differentiating between "CUSTOM" and "RECOMMENDED".
2. the code importing `recommendedTemplates.json` could be moved to the reducer, in order to add it to the initial state.
3. `toggleRecommendedFavourite` can be moved to the reducer as an action too, and edit the `favourite` state locally there

what do you think?

Schwehn42 (commented):
thanks for the PR, you've implemented the general working idea pretty well.

I commented some suggestions that I think would improve the structure and readability of the changes.

One big point that needs to be addressed is that as of now, the local favourite is not persisted! that's definitely one acceptance criterion, as otherwise the changes are lost with a reload.

src/store/features/templates/reducer.ts
```diff
@@ -1,22 +1,45 @@
...
+import {TemplatesState, Template} from "./types";
 import {createTemplateWithColumns, deleteTemplate, editTemplate, getTemplates} from "./thunks";
 
 // both this and templateColumns reducer don't have an empty state.
 // the reason for this is the fact the template/templateColumn logic for reordering is handled partly by the reducer.
 // therefore, they need to be in the state at all times so they can be accessed
-const initialState: TemplatesState = [DEFAULT_TEMPLATE.template];
+
+// Helper to generate recommended templates with ids, type, etc.
+const generateRecommendedTemplates = (): Template[] => (recommendedTemplatesJson as any[]).map((tpl, idx) => ({
```
Schwehn42: never cast as `any`, as it thwarts type safety. actually it will also be flagged by the linter so the pipeline fails because of it.

Instead, cast as the already existing type `ImportReducedTemplateWithColumns[]`

src/store/features/templates/reducer.ts
```diff
@@ -1,22 +1,45 @@
...
+    id: `recommended-${idx}`,
+    creator: "scrumlr",
+    name: tpl.name,
+    description: tpl.description,
+    favourite: tpl.favourite ?? false,
+    type: "RECOMMENDED",
+  })) as Template[];
+
+const initialState: TemplatesState = [{...DEFAULT_TEMPLATE.template, type: "CUSTOM"}, ...generateRecommendedTemplates()];
```
Schwehn42: I guess one could also add the template type to the default template constant itself, or did you decide against it?

Schwehn42: nvm, you did add it, so there's no reason to override is there?

src/store/features/templates/types.ts
```diff
@@ -19,6 +22,7 @@ export type TemplateWithColumns = {template: Template; columns: TemplateColumn[]
 export type TemplatesState = Template[];
 
 // used for importing, where information like ids will be set dynamically
-export type ImportReducedTemplateWithColumns = Omit<Template, "id" | "creator" | "favourite"> & {
+export type ImportReducedTemplateWithColumns = Omit<Template, "id" | "creator"> & {
+  favourite?: boolean;
```
Schwehn42: I think the property `favourite` doesn't have to be manually re-added, as it is already part of `Template` -- and is no longer omitted

src/routes/Boards/Templates/Templates.tsx
```diff
@@ -53,36 +52,29 @@ export const Templates = () => {
...
+      id: `column-${idx}-${indexColumn}`,
+      template: templateId,
+      index: indexColumn,
+      name: t(column.name, {ns: "templates"}),
+      description: t(column.description, {ns: "templates"}),
     }));
-
-  const reducedRecommendedTemplates: ImportReducedTemplateWithColumns[] = templatesJsonRaw as ImportReducedTemplateWithColumns[];
-  const fullRecommendedTemplates = convertJsonToFullTemplates(reducedRecommendedTemplates);
+  };
```
Schwehn42: since we've already added the recommended templates to our app state, maybe we should also add their associated columns to the `templateColumn` state. what do you think?

src/store/features/templates/types.ts
jeancloudvanlang: done

src/store/features/templates/reducer.ts
jeancloudvanlang: done

src/routes/Boards/Templates/Templates.tsx
jeancloudvanlang: would be a lot cleaner and simpler. on it

src/store/features/templates/reducer.ts
```diff
@@ -1,22 +1,66 @@
...
+const RECOMMENDED_FAVOURITES_KEY = "recommendedTemplateFavourites";
+
+function getRecommendedFavouritesFromStorage(): string[] {
+  try {
+    return JSON.parse(localStorage.getItem(RECOMMENDED_FAVOURITES_KEY) || "[]");
+  } catch {
+    return [];
+  }
+}
+
```
Schwehn42: - use `constants/storage.ts` and `utils/storage.ts` instead 
- maybe implement a defaultValue there as well like `getNumberFromStorage` (but use [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)

Schwehn42: sorry @jeancloudvanlang I worded that adversely, I meant to generalize `getRecommendedFavouritesFromStorage` as `getFromStorage` with a default value inside `utils/storage.ts`, then use the new storage key to use :D

jeancloudvanlang: refactored by merging those functions and having the default value as an optional parameter. what do you think?

Schwehn42: looks sensible 👍

Schwehn42 (approved):
great work :)

Issue #5184, opened 2025-06-23
Design Fixes: Template Page


SelinaBuff:
### **General – Template Page**:

1. **Search Bar and Switch:**

   * The toggle switch is temporarily hidden.
   * The search bar is extended to utilize the freed-up space.

2. **Typography:**

   * The main heading “Erstelle eine Vorlage” and card titles are reduced from 24px to 20px.

3. **Scroll Behavior:**

   * Only the card section is scrollable.
   * The section headers “Unsere Empfehlungen” and “Deine Vorlagen” remain sticky while scrolling.

4. **Card Styling:**

   * Card padding is set to 16px.

5. **Favorite Star on Recommendation Cards:**

   * A favorite star is also displayed on cards in the “Recommendations” section. Selection is stored locally.
   * The star shows a visible focus state (e.g., an ellipse around the icon) **only** when accessed via keyboard navigation.

6. **Logout Function:**

   * A logout option is available in the user menu (see Figma design: 1).
   * After logout, the user is redirected to the login page.

---

### **Create Template**:

7. **Title:**

   * The page title and input field headings are reviewed for correct English–German translation.

8. **Layout Adjustment:**

   * The horizontal width of the column section is expanded to reach halfway between the logo (left) and user avatar (right).

9. **Helper Texts:**

   * Explanatory helper text is added below the columns, according to the Figma design.

10. **Placeholder Texts for Column Headers:**

* If a counter is functional:
  → Placeholder: “Column 1” / “Spalte 1” and “Actions” / “Aktionen”
* If no counter is available:
  → Unified placeholder: “Title” / “Titel”

---

### **Access Settings**:

11. **Alignment:**

* All content in this section is left-aligned.

Schwehn42:
all taks/subtasks done

PR #5231, opened 2025-07-15
scrolling in template container #5184

jeancloudvanlang:
## Description
enabled scrolling in template container with sticky headers and adjusted padding of cards as well

## Changelog
enabled scrolling in template container with sticky headers and adjusted padding of cards as well (according to change request)

## Checklist

### DIFF
--- a/src/routes/Boards/Templates/Templates.scss
+++ b/src/routes/Boards/Templates/Templates.scss
@@ -21,11 +21,7 @@ $max-card-width: 500px;
   flex: 1;
   gap: $spacing--lg;
 
-  padding: $spacing--xl;
   border-radius: $rounded--default;
-
-  overflow-y: scroll;
-
   background-color: $blue--50;
 
   &--saved {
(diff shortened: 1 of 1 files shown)


src/routes/Boards/Templates/Templates.scss
```diff
@@ -67,6 +67,15 @@ $max-card-width: 500px;
...
   gap: $spacing--md;
+  overflow-y: scroll;
+  scrollbar-width: none;
+
+  margin-bottom: -$spacing--xl;
+  margin-left: -$spacing--xl;
+  margin-right: -$spacing--xl;
+  padding-bottom: $spacing--xl;
+  padding-left: $spacing--xl;
+  padding-right: $spacing--xl;
```
Schwehn42: generally, you want to avoid negative spacings as much as possible. I think to get the same effect of the bottom card not clipping into the bottom, one extra container around might do the trick

jeancloudvanlang: it was done like this so the user can already scroll the container when inside the parent element. Might be also sufficient that the users mouse has to be inside the card elements..

jeancloudvanlang: used paddings of child elements instead

PR #5248, opened 2025-07-24
moved template columns to app state 

jeancloudvanlang:
moved template columns to app state


src/store/features/templates/reducer.ts
```diff
@@ -64,3 +64,12 @@ export const templatesReducer = createReducer(initialState, (builder) => {
...
 });
+
+export const recommendedTemplateColumns: Record<string, ImportReducedTemplateWithColumns["columns"]> = (recommendedTemplatesJson as ImportReducedTemplateWithColumns[]).reduce(
+  (acc, tpl, idx) => {
+    const id = `recommended-${idx}`;
+    acc[id] = tpl.columns;
+    return acc;
+  },
+  {} as Record<string, ImportReducedTemplateWithColumns["columns"]>
+);
```
Schwehn42: I don't understand why you're constructing an object out of the array here, wouldn't an array be easier or was there a specific reason?

jeancloudvanlang: more or less an deprecated idea. I thought with an Object you can retrieve corresponding Columns faster via object id. But in our current situation we should use the array. thx!

PR #5263, opened 2025-07-29
feat: logout button

jeancloudvanlang:
## Description
Added Logout button in settings dialog

## Changelog
Added Logout button in settings dialog with redirects the user to the login page afterwards and removes it from the store

## Checklist

### DIFF
--- a/src/components/SettingsDialog/SettingsDialog.scss
+++ b/src/components/SettingsDialog/SettingsDialog.scss
@@ -30,9 +30,9 @@ $settings-dialog-container--bottom: 62px;
 }
 
 .settings-dialog__logout.navigation__item {
-  margin-top: auto;
   margin-bottom: $spacing--xl;
   border: none;
+  background-color: transparent;
 
   &:hover {
     background-color: $navy--200;
@@ -41,11 +41,6 @@ $settings-dialog-container--bottom: 62px;
   .navigation-item__icon {
     background: $navy--400;
     color: $gray--000;
-
-    svg {
-      width: 2rem;
-      height: 2rem;
-    }
   }
 }
 
@@ -246,6 +241,7 @@ $settings-dialog-container--bottom: 62px;
   display: flex;
   flex-direction: column;
   justify-content: center;
+  align-self: center;
   gap: 2px;
   overflow: hidden;
 }
@@ -284,6 +280,9 @@ $settings-dialog-container--bottom: 62px;
   border-radius: $rounded--full;
   background: var(--accent-color--light);
   color: $gray--000;
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
 
 .settings-dialog__container {
--- a/src/components/SettingsDialog/SettingsDialog.scss
+++ b/src/components/SettingsDialog/SettingsDialog.scss
@@ -211,6 +211,7 @@ $settings-dialog-container--bottom: 62px;
   flex-direction: row;
   gap: 16px;
   padding: 8px;
+  cursor: pointer;
 
   &:hover {
     background: rgba(var(--accent-color--light-rgb), 0.8);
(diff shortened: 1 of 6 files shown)


src/components/SettingsDialog/SettingsDialog.scss
```diff
@@ -28,6 +28,27 @@ $settings-dialog-container--bottom: 62px;
...
+  }
+
+  .navigation-item__icon {
+    background: $navy--400;
+    color: $gray--000;
+
+    svg {
+      width: 2rem;
+      height: 2rem;
+    }
```
Schwehn42:
```suggestion
display: flex;
justify-content: center;
align-items: center;
```

this is better (you can also apply this to the general `.navigation-item__icon` in L283 instead of here)

src/components/SettingsDialog/SettingsDialog.scss
```diff
@@ -28,6 +28,27 @@ $settings-dialog-container--bottom: 62px;
   display: flex;
   flex-direction: row;
 }
+
+.settings-dialog__logout.navigation__item {
+  margin-top: auto;
```
Schwehn42: removing `margin-top` doesn't make a difference for me

src/components/SettingsDialog/SettingsDialog.scss
```diff
@@ -28,6 +28,27 @@ $settings-dialog-container--bottom: 62px;
...
+  margin-bottom: $spacing--xl;
+  border: none;
+
+  &:hover {
+    background-color: $navy--200;
+  }
+
+  .navigation-item__icon {
+    background: $navy--400;
```
Schwehn42: background should be transparent by default according to design, and imo `navy-200` is better as hover color (my sense for UI/UX is debatable though)

src/components/SettingsDialog/SettingsDialog.scss
```diff
@@ -28,6 +28,27 @@ $settings-dialog-container--bottom: 62px;
   display: flex;
   flex-direction: row;
 }
+
+.settings-dialog__logout.navigation__item {
```
Schwehn42: I would move this all the way to the bottom (around L290) below the general CSS styling rules of `navigation-item__*`. That way, its easier to see that these rules are overriding the general ones by using specificity

Schwehn42: alternatively, I also often like to add `--specifier` to class names, e.g. `my-class__part-of-class--specifier`, or `navigation__item navigation__item--logout` in our case, and add the specific rules inside like this:
```scss
.navigation__item {
  some-rule: value;

  &--logout {
    some-rule: other-value; // overrides if element also has this class
  }
}
```

both work fine though, up to you how you want to do it

src/components/SettingsDialog/SettingsDialog.scss
jeancloudvanlang: nice!

src/components/SettingsDialog/SettingsDialog.scss
jeancloudvanlang: you mean the button should be transparent and not the icon, right?

src/components/SettingsDialog/SettingsDialog.scss
```diff
@@ -213,12 +220,29 @@ $settings-dialog-container--bottom: 62px;
   &:focus-visible {
     box-shadow: 0 0 0 2px var(--accent-color--light);
   }
+
+  &--logout {
+    background-color: $navy--200;
+    margin-bottom: $spacing--xl;
+    border: none;
+    background-color: transparent;
```
Schwehn42: rm first background-color rule as its overridden below

Schwehn42 (commented):
one small thing, could you also add `cursor: pointer;` to `.navigation__item`

that way the logout button is correctly rendered as clickable

src/components/SettingsDialog/SettingsDialog.scss
jeancloudvanlang: thx. done

Issue #5266, opened 2025-07-30
Theme not stored properly


Schwehn42:
### Description
Theme is not stored properly, causing the wrong theme to be used

### Reproduce
1. Change theme light
2. Reload page
3. Theme is set to auto

### Context
Maybe caused by local storage changes introduced in #5219

PR #5276, opened 2025-07-31
fixed theme not parsing correctly from local storage #5266

jeancloudvanlang:
## Description
Fixed theme not parsing correctly from local storage 

## Checklist

### DIFF
--- a/src/utils/storage.ts
+++ b/src/utils/storage.ts
@@ -6,15 +6,7 @@ export const saveToStorage = (key: string, value: string): void => localStorage.
 /**
  * Retrieves value from local storage by key.
  */
-export function getFromStorage<T>(key: string, defaultValue: T = null as T): T {
-  try {
-    const value = localStorage.getItem(key);
-    if (value == null) return defaultValue;
-    return (typeof value === "string" ? value : JSON.parse(value)) as T;
-  } catch {
-    return defaultValue;
-  }
-}
+export const getFromStorage = <T>(key: string): T | null => localStorage.getItem(key) as T | null;
 
 /**
  * Retrieves number from local storage by key. Returns alternative value on error.
--- a/src/store/features/templates/reducer.ts
+++ b/src/store/features/templates/reducer.ts
@@ -12,7 +12,7 @@ import {createTemplateWithColumns, deleteTemplate, editTemplate, getTemplates} f
 
 // Helper to generate recommended templates with ids, type, etc.
 const generateRecommendedTemplates = (): Template[] => {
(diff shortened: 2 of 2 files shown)


src/utils/storage.ts
```diff
@@ -8,7 +8,9 @@ export const saveToStorage = (key: string, value: string): void => localStorage.
  */
 export function getFromStorage<T>(key: string, defaultValue: T = null as T): T {
```
Schwehn42: casting null as T is not a good idea I think

```suggestion
export function getFromStorage<T>(key: string, defaultValue?: T): T {
```

this might be better, returning undefined if no value is found and no default value is set (note this makes the defaultvalue optional to the developer, so you gotta be aware of that when using it)

also note I did not test this

Schwehn42: actually its probably to better to either
- have defaultValue be optional but then also return type optional (i.e. can be T or undefined)
- or have defaultValue be required, then return value can stay T

src/utils/storage.ts
```diff
@@ -8,7 +8,9 @@ export const saveToStorage = (key: string, value: string): void => localStorage.
...
   try {
-    return JSON.parse(localStorage.getItem(key) ?? JSON.stringify(defaultValue)) as T;
+    const value = localStorage.getItem(key);
+    if (value == null) return defaultValue;
+    return (typeof value === "string" ? value : JSON.parse(value)) as T;
```
Schwehn42: localStorage returns type `string` or `null`, thus we don't need the type check nor the conditional return


```suggestion
    return JSON.parse(value) as T;
```

jeancloudvanlang: actually this will cause the error again because `JSON.parse("light")` would not be a valid json

src/utils/storage.ts
jeancloudvanlang: this implies changes in the function signature, means that the current implementations using this function have to be adjusted as well. I would tend to sth not changing existing code^^

Schwehn42: alrighty, I actually think there would be no change required for the callers but we will leave it like this, also works 👍

src/utils/storage.ts
Schwehn42: okay, but still this doesn't make sense as typeof value is always string

as TS is not able to parse typeof T directly (T is compile time while typeof is runtime) one could try checking defaultValue instead

although that will also run into some issues with this func signature for sure

jeancloudvanlang: refactored so that no default param is supported because it was only used once

Schwehn42 (approved):
lgtm 🥇

PR #5286, opened 2025-08-05
fixed #5265 by adding missing column localization for recommended templates 

jeancloudvanlang:
## Description
added missing localization of column names of recommended templates

## Checklist

### DIFF
--- a/src/i18n/de/translation.json
+++ b/src/i18n/de/translation.json
@@ -35,8 +35,8 @@
     },
 
     "TemplateEditor": {
-      "create": "Erstelle eine Vorlage",
-      "edit": "Bearbeite eine Vorlage",
+      "createTitle": "Erstelle eine Vorlage",
+      "editTitle": "Bearbeite eine Vorlage",
       "cancel": "Abbrechen",
       "saveEdit": "Speichern",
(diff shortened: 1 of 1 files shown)


src/store/features/templateColumns/reducer.ts
```diff
@@ -14,6 +15,7 @@ function withMeta(columns: Omit<TemplateColumn, "id" | "template" | "index">[],
     id: `${templateId}-${idx}`,
     template: templateId,
     index: idx,
+    name: t(col.name, {ns: "templates"}),
```
jeancloudvanlang, line 18: the description is not needed, right?

Schwehn42 (approved):
works, lets keep in mind though that we are essentially double translating in Templates.tsx: 
https://github.com/inovex/scrumlr.io/blob/aae9b7d7d5f5e194e9fedb051f5ed8956804041a/src/routes/Boards/Templates/Templates.tsx#L69

some parts are translated only temporarily in that component for viewing, some others (like column names) are saved the store. so, when we are going to change this logic in the future (e.g. by also persisting recommended templates), we should keep this in mind.

PR #5291, opened 2025-08-06
fix: template editor settings

### LINKED ISSUE #5290: Template Editor Settings 404
Right now, opening the settings on /boards/create/ or /boards/edit/* will result in a 404

Schwehn42:
## Description

Fixes #5290  
Previously, opening settings by clicking on the user pill top right on the templates view would not work within the edit template view. this is mostly because these routes were never added in the router -- but also some other small change concerning the navigation to the settings from the user pill had to be adjusted so that the uuid is included in the location prefix for the router navigate.

## Changelog

- add settings routes (appearance, profile, optional feedback) to template editor (/boards/create, /boards/edit/:id)
- add outlet to template editor so settings are shown
- retrieve id param to be used for navigation location prefix 

## Checklist

### DIFF
--- a/src/routes/Boards/Boards.tsx
+++ b/src/routes/Boards/Boards.tsx
@@ -1,5 +1,5 @@
 import {useTranslation} from "react-i18next";
-import {Outlet, useLocation} from "react-router";
+import {Outlet, useLocation, useParams} from "react-router";
 import {useEffect, useState} from "react";
 import {ScrumlrLogo} from "components/ScrumlrLogo";
 import {UserPill} from "components/UserPill/UserPill";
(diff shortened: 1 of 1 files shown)


src/routes/Boards/Boards.tsx
```diff
@@ -15,12 +15,15 @@ type BoardView = "templates" | "sessions" | "create" | "edit";
...
   const location = useLocation();
+  const {id: editTemplateId} = useParams();
   // const navigate = useNavigate();
   const dispatch = useAppDispatch();
 
   const [boardView, setBoardView] = useState<BoardView>("templates");
   // a simplification of BoardView in order to change some render behaviour (e.g. conditional render of SearchBar)
   const viewType = ["templates", "sessions"].includes(boardView) ? "overview" : "edit";
+  // for edit route, expand location prefix used for settings with the edit template uuid
+  const locationPrefix = boardView === "edit" ? `edit/${editTemplateId}` : boardView;
```
jeancloudvanlang, line 26: what happens in case the `editTemplateId` is undefined? checking if it is available could make sense.

jeancloudvanlang: and redirecting to the template overview page? (atm e.g. in case the user is in on the template creation page, logs out and then logs in, the edit template page is appearing)

Schwehn42: in that case this error will be shown:
Screenshot 2025-08-06 at 16 28 51

I guess one could add a guard to the router to check whether the template exists (i.e. is in the store) and redirect to the template overview if not?

jeancloudvanlang: let's keep it at it is since a change would cover a super rare use case

PR #5304, opened 2025-08-08
refactor: emoji autocomplete

### LINKED ISSUE #5275: Column Description: Emojis
Emoji Autocompletion should work for both column name and description

Schwehn42:
## Description

Refactors emoji autocomplete hook. Closes #5275.

Initially just meant to implement emojis for the new columns details (#4280), some more work was required for our functionality:
1. cursor not setting at correct place after inserting an emoji
2. value state and getter couldn't be delegated
3. confusing to use altogether

This PR should solve all of these issues by refactoring parts of the hook.  
This not only adds emojis to columns details, but also modifies how other components using it (namely NoteInput and NoteDialogNoteContent) interact with the hook.

additional features/changes:
- cmd/ctrl+enter to submit changes of both name and description input
- clicking outside of column name details will reset changes rather than submitting (streamlined with template editor)

## Changelog

- `useEmojiAutocomplete` hook
  - changed options object, e.g. adding input ref
  - update cursor position dynamically after inserting emoji
  - delegate value and value setter to outside, such that changes can be controlled
- `EmojiAutosuggestions`
  - can be set to relative positioning
-  add hook `useSubmitOnShortcut`
  - delegates callback function for inputs upon pressing ctrl/cmd+enter
- `TextArea`
  - add emoji suggestions
  - can be turned off using prop option
  - submit on shortcut
- `ColumnHeader`
  - add emoji suggestions for name input
  - submit on shortcut
  - cancel update / reset changes upon clicking outside

## Checklist


src/components/Column/ColumnDetails/ColumnDetails.tsx
```diff
@@ -203,12 +214,13 @@ export const ColumnDetails = (props: ColumnDetailsProps) => {
...
 
   return (
-    <div className="column-details" ref={ref}>
+    <div className="column-details" ref={containerRef}>
       <div className="column-details__name-and-settings-wrapper">
         <div className="column-details__name-wrapper">{renderName()}</div>
         {renderSettings()}
       </div>
       <div className={classNames("column-details__description-wrapper", `column-details__description-wrapper--${props.mode}`)}>{renderDescription()}</div>
+      <EmojiSuggestions positionRelative {...emoji.suggestionsProps} /> {/* suggestions for name input, as textarea has its own */}
```
jeancloudvanlang, line 223: instead of spreading the props one could list them explicitly

src/components/TextArea/TextArea.tsx
```diff
@@ -19,26 +22,41 @@ type TextAreaProps = {
...
+    inputRef: internalRef,
+    value: props.input,
+    onValueChange: props.setInput,
+    maxInputLength: props.maxLength,
+    suggestionsHidden: !props.emojiSuggestions,
+  });
+
   return (
     <>
       <style>{`.text-area { --text-area-rows: ${rows} }`}</style>
```
jeancloudvanlang, line 56: does it make sense to set the rows as a prop of the TextareaAutosize component and get rid of the inline style?

jeancloudvanlang (approved):

src/components/Column/ColumnDetails/ColumnDetails.tsx
Schwehn42: true, though its debatable whether this has a readability impact as you basically just pass on the return value of the `useEmojiAutocomplete` hook

src/components/TextArea/TextArea.tsx
Schwehn42: the style is needed so it can be used as a CSS variable. CSS rules are basically pre-compiled, where the only variables are such as the screen width and so on. If you want to change styles based on arbitrary values, such as the amount of rows, one has two possibilities:
1. use JS to update styles dynamically
2. pass CSS variables

I know it looks a little silly, but because of the TS type checker this seemed to be the easiest way to pass CSS variables

PR #5311, opened 2025-08-12
added feature toggle for en/disabling template creation in UI

jeancloudvanlang:
## Description
Added toggle for en/disabling template creation in UI by introducing an env variable

## Changelog
added env variable for en/disabling of template creation
added disabling of the button and showing a tooltip (in case it is disabled)

## Checklist


src/store/features/participants/types.ts
```diff
@@ -10,6 +10,7 @@ export interface Participant {
   showHiddenColumns: boolean;
   role: ParticipantRole;
   banned?: boolean;
+  isAnonymous?: boolean;
```
Schwehn42: not needed here, participants state is only filled _after_ joining a board. this is before a board, i.e. template screen, therefore use `state.auth.user?.isAnonymous` instead

src/utils/test/getTestApplicationState.ts
```diff
@@ -183,7 +183,8 @@ export default (overwrite?: Partial<ApplicationState>): ApplicationState => ({
     moderating: false,
     serverTimeOffset: 0,
     anonymousLoginDisabled: false,
-    allowAnonymousCustomTemplates: true,
+    allowAnonymousCustomTemplates: false,
+    allowAnonymousCreateBoards: false,
```
Schwehn42: should be allowed by default for testing purposes

Schwehn42:
will either be re-opened or superseded later, not sure yet

PR #5389, opened 2025-09-09
ci: migrate ARM builds to native GitHub ARM runners

Yaazizi:
## Description
This PR updates the CI/CD pipeline to build ARM container images on GitHub's new native ARM runners.  
By removing emulation (QEMU) and targeting `ubuntu-24.04-arm`, ARM builds are now faster and more efficient.  

## Changelog
- Updated `.github/workflows/package.yml`
- ARM jobs now run on native GitHub ARM runners
- Explicitly target `linux/arm64` platform

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -9,6 +9,12 @@ on:
       server-image-tag:
         description: The tag of the server image
         value: ${{ jobs.package-server.outputs.server-image-tag }}
+      frontend-image-tag-arm64:
+        description: The tag of the frontend arm64 image
+        value: ${{ jobs.package-frontend-arm64.outputs.frontend-image-tag-arm64 }}
+      server-image-tag-arm64:
+        description: The tag of the server arm64 image
+        value: ${{ jobs.package-server-arm64.outputs.server-image-tag-arm64 }}
 
 jobs:
   determine-platforms:
@@ -19,11 +25,11 @@ jobs:
       - name: Set platforms
         id: platforms
         run: |
-          if [ "${{ github.event_name }}" == "release" ]; then
-            echo "platforms=linux/amd64,linux/arm64" >> $GITHUB_OUTPUT
-          else
-            echo "platforms=linux/amd64" >> $GITHUB_OUTPUT
-          fi
+          # CHANGED: always build amd64 here to avoid emulation on x86 runners
+          # (arm64 will be handled by dedicated native ARM jobs below)
+          echo "platforms=linux/amd64" >> $GITHUB_OUTPUT
+
+
   package-frontend:
     needs: determine-platforms
     outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/package.yml
```diff
@@ -19,11 +25,11 @@ jobs:
...
-          if [ "${{ github.event_name }}" == "release" ]; then
-            echo "platforms=linux/amd64,linux/arm64" >> $GITHUB_OUTPUT
-          else
-            echo "platforms=linux/amd64" >> $GITHUB_OUTPUT
-          fi
+          # CHANGED: always build amd64 here to avoid emulation on x86 runners
+          # (arm64 will be handled by dedicated native ARM jobs below)
+          echo "platforms=linux/amd64" >> $GITHUB_OUTPUT
+
+
```
Planlos5000, line 16: It looks like this job can be deleted

.github/workflows/package.yml
```diff
@@ -146,4 +152,129 @@ jobs:
...
+      - name: Set outputs (arm64)
+        id: outputs
+        run: |
+          echo $DOCKER_METADATA_OUTPUT_TAGS
+          echo _____________________________________________________
+          # Example result: sha-1a2b3c4-arm64 (because of the flavor suffix)
+          echo frontend-image-tag-arm64=$(echo "$(echo $DOCKER_METADATA_OUTPUT_TAGS | grep -oP 'sha-[a-f0-9]+-arm64')" ) >> $GITHUB_OUTPUT
+
+  package-server-arm64:
```
Planlos5000: The not arm version for the server is missing

.github/workflows/package.yml
```diff
@@ -101,49 +124,104 @@ jobs:
...
+          echo "frontend_image_tag=${fe_amd64}"       >> "$GITHUB_OUTPUT"
+          echo "server_image_tag=${sv_amd64}"         >> "$GITHUB_OUTPUT"
+          echo "frontend_image_tag_arm64=${fe_arm64}" >> "$GITHUB_OUTPUT"
+          echo "server_image_tag_arm64=${sv_arm64}"   >> "$GITHUB_OUTPUT"
+
+          if [ -z "$fe_amd64" ] || [ -z "$sv_amd64" ] || [ -z "$fe_arm64" ] || [ -z "$sv_arm64" ]; then
+            echo "ERROR: One or more image tags are empty. Aborting to avoid deploying with an empty tag."
+            ls -l artifacts || true
+            exit 1
+          fi
```
Planlos5000: Why is this job needed?

.github/workflows/package.yml
```diff
@@ -4,31 +4,38 @@ on:
...
+      matrix:
+        include:
+          - arch: amd64
+            runner: ubuntu-latest
+            platform: linux/amd64
+            suffix: ""
+          - arch: arm64
+            runner: ubuntu-24.04-arm
+            platform: linux/arm64
+            suffix: "-arm64"
```
Planlos5000: Maybe define the matrix like in the [rebuild workflow](https://github.com/inovex/scrumlr.io/blob/main/.github/workflows/rebuild_images.yaml). I think it is easier to read

`matrix:`
`    build_info:`
`        - {platform: 'linux/amd64', runner: 'ubuntu-latest'}`
`        - {platform: 'linux/arm64', runner: 'ubuntu-24.04-arm'}`

.github/workflows/package.yml
```diff
@@ -4,31 +4,38 @@ on:
...
-        description: The tag of the server image
-        value: ${{ jobs.package-server.outputs.server-image-tag }}
+        description: The tag of the server image (amd64)
+        value: ${{ jobs.export-outputs.outputs.server-image-tag }}
+      frontend-image-tag-arm64:
+        description: The tag of the frontend image (arm64)
+        value: ${{ jobs.export-outputs.outputs.frontend-image-tag-arm64 }}
+      server-image-tag-arm64:
+        description: The tag of the server image (arm64)
+        value: ${{ jobs.export-outputs.outputs.server-image-tag-arm64 }}
```
Planlos5000: Are all these outputs needed?

.github/workflows/package.yml
```diff
@@ -39,58 +46,74 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
-      - name: Log in to GitHub container registry
+      - name: Default platform env (only for arm64)
+        if: matrix.arch == 'arm64'
+        run: echo "DOCKER_DEFAULT_PLATFORM=${{ matrix.platform }}" >> $GITHUB_ENV
```
Planlos5000: I think this value is never used

.github/workflows/package.yml
Yaazizi: I originally added the frontend-image-tag-arm64 and server-image-tag-arm64 outputs because we are building both amd64 and arm64 images, and I thought it might be useful to expose the tags for both architectures.

But you’re right – since our current deploy workflow only uses the amd64 tags, the extra outputs are not needed. I’ll remove the arm64 outputs to keep the workflow interface clean.

PR #5394, opened 2025-09-10
feat: Add integration tests for note service

Planlos5000:
## Description

Add integration tests for note service and some API changes to the note service, so that values are extracted from the context in the api layer and not in the service layer.

## Changelog

- Change Service API
- Recreate mocks
- Fix tests
- Add integration tests
## Checklist

### DIFF
--- a/server/src/notes/service_integration_test.go
+++ b/server/src/notes/service_integration_test.go
@@ -1,0 +1,445 @@
+package notes
+
+import (
+	"context"
+	"log"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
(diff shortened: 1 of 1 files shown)


server/src/api/notes.go
```diff
@@ -107,7 +110,10 @@ func (s *Server) deleteNote(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := s.notes.Delete(r.Context(), body, note); err != nil {
+	body.ID = note
+	body.Board = board
```
mateo-ivc, line 114: Is fine by me, but it looks unintuitive overwriting the request body

Planlos5000: Yes, but for the update method it is done in the same way.  But i think it is better in that way than getting the note id and the board id in the service from the context.

Issue #5380, opened 2025-09-08
Create a board on mobile


Planlos5000:
### The problem

On small mobile devices you can't create a board because the buttons for starting a session are not displayed and the UI does not allow to scroll to the buttons.
On bigger mobile devices the dialog for starting a session is slightly cut of.

### Browser

Safari

### Steps to reproduce the behavior

Open scrumlr.io on your phone and create a new board

### Screenshots

IPad Air 11' with Safari Browser
![Image](https://github.com/user-attachments/assets/12350995-a35c-4704-8ac1-35813260dea8)
![Image](https://github.com/user-attachments/assets/279bc40e-4d96-4672-8751-58e20ea9dfe5)

IPhone 12 with Safari Browser
<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/540b0ddc-ed6a-4b8e-ab1f-409fa3a22cc5" />

### Additional context

This was tested with an IPhone 12 and an IPad Air 11'

ghost:
We have another viewport issue on mobile devices in our "accept cookies" modal: 

<img width="412" height="685" alt="Image" src="https://github.com/user-attachments/assets/2b21e777-5498-4823-957d-227f1782d302" />

ghost:
Both fixed in this merge request: https://github.com/inovex/scrumlr.io/pull/5399

PR #5399, opened 2025-09-11
feat: fix out of display components on mobile #5380

ghost:
## Description
These changes fix mobile UI issues where the cookie notice buttons and access settings modal were partially or completely hidden on smaller screens. Both components now use responsive layouts with proper scrolling behavior to ensure all interactive elements remain accessible on mobile devices.

## Changelog
- Cookie Notice: Added responsive flexbox layout that stacks buttons vertically on mobile screens (≤768px) and reorders them for better UX
- Access Settings Modal: Made modal scrollable and repositioned to top of screen on mobile to prevent content from extending beyond viewport

## Checklist

## (Optional) Visual Changes

Made template settings scrollable:

https://github.com/user-attachments/assets/325abe25-b70c-425f-b2ec-74ef72e98246

Cookie layout fix: 
image

### DIFF
--- a/src/components/Templates/AccessSettings/AccessSettings.scss
+++ b/src/components/Templates/AccessSettings/AccessSettings.scss
@@ -48,11 +48,11 @@
 
 @media screen and (max-width: $breakpoint--smartphone) {
   .access-settings__wrapper {
-    align-items: flex-start;
     padding: $spacing--xs;
   }
 
   .access-settings {
+    height: 90%;
     width: 85%;
     gap: $spacing--md;
 


src/components/CookieNotice/CookieNotice.scss
Schwehn42, line 1: instead of setting the order specifically, why not just use `flex-direction: column-reverse`?

src/components/Templates/AccessSettings/AccessSettings.scss
Schwehn42, line 1: on some screen sizes, e.g. 920x475, the modal still looks questionable as it clips out of the screen, but that's probably not so bad because who has that kind of screen.

However, on common screen sizes, e.g. 402x874, which is the default iPhone Pro portrait size, the modal should be centered imo.

ghost: What do you mean with "the modal should be centered"? Lets have a quick call about this.

Schwehn42: I just mean the horizontal alignment (its already vertically aligned) on mobile screens so its totally centered

Screenshot 2025-09-17 at 15 47 23

this is the how it looks currently, would look better if it was also horizontally centered imo :)

Schwehn42 (approved):
lgtm🏚️

PR #5403, opened 2025-09-12
refactor: seperate user from session

mateo-ivc:
## Description

Changing the frontend api requests that a separation of user and session in the backend is possible
## Changelog

`backend`
 - Separating user from session -> moving files into their own package
 
 `frontend`
 - added extra API request to receive the missing Auth object
 - `board/thunk.js` -> biggest problem was here since (user)sessions now only contains the userID but the frontend needed the full user object => added extra API request where needed
 
Keep in mind that the frontend changes should not be kept like this. The goal is here to also refactor the frontend api in order to achieve the separation  

## Checklist

### DIFF
--- a/server/src/users/dto.go
+++ b/server/src/users/dto.go
@@ -1,45 +1,41 @@
 package users
 
 import (
-	"net/http"
+  "net/http"
 
-	"github.com/google/uuid"
-	"scrumlr.io/server/common"
+  "github.com/google/uuid"
+  "scrumlr.io/server/common"
 )
 
 // User is the response for all user requests.
 type User struct {
-	// The id of the user
-	ID uuid.UUID `json:"id"`
+  // The id of the user
+  ID uuid.UUID `json:"id"`
 
-	// The user's display name
-	Name string `json:"name"`
+  // The user's display name
+  Name string `json:"name"`
 
-	// The user's avatar configuration
-	Avatar *common.Avatar `json:"avatar,omitempty"`
+  // The user's avatar configuration
+  Avatar *common.Avatar `json:"avatar,omitempty"`
 
-	// The user's account type configuration
-	AccountType common.AccountType `json:"accountType"`
+  // The user's account type configuration
+  AccountType common.AccountType `json:"accountType"`
 }
 
 type UserUpdateRequest struct {
-	ID     uuid.UUID      `json:"-"`
-	Name   string         `json:"name"`
-	Avatar *common.Avatar `json:"avatar,omitempty"`
+  ID     uuid.UUID      `json:"-"`
+  Name   string         `json:"name"`
+  Avatar *common.Avatar `json:"avatar,omitempty"`
 }
 
 func (u *User) From(user DatabaseUser) *User {
-	u.ID = user.ID
-	u.Name = user.Name
-	u.Avatar = user.Avatar
-	u.AccountType = user.AccountType
-	return u
+  u.ID = user.ID
+  u.Name = user.Name
+  u.Avatar = user.Avatar
+  u.AccountType = user.AccountType
+  return u
 }
 
 func (*User) Render(_ http.ResponseWriter, _ *http.Request) error {
-	return nil
-}
-
-type UserID struct {
-	UserID uuid.UUID `json:"id"`
+  return nil
 }


server/src/sessions/dto.go
```diff
@@ -2,40 +2,108 @@ package sessions
...
 	"github.com/google/uuid"
 	"scrumlr.io/server/common"
 )
 
-// User is the response for all user requests.
-type User struct {
-	// The id of the user
+// BoardSession is the response for all participant requests.
+type BoardSession struct {
 	ID uuid.UUID `json:"id"`
```
Planlos5000: Rename to User, because the session has a combined primary key of board and user

server/api.postman_collection.json
Planlos5000, line 1: Maybe a new test for the `GET /user/{id}` route

server/src/api/users.go
```diff
@@ -1,77 +1,97 @@
...
-		span.RecordError(err)
-		common.Throw(w, r, err)
-		return
-	}
+  userParam := chi.URLParam(r, "user")
+  requestedUserId, err := uuid.Parse(userParam)
+  if err != nil {
+    common.Throw(w, r, err)
+  }
+  user, err := s.users.Get(r.Context(), requestedUserId)
```
Planlos5000: Add the otel span and use the context from the span

server/src/sessions/database.go
```diff
@@ -2,175 +2,212 @@ package sessions
...
+	"errors"
 
 	"github.com/google/uuid"
 	"github.com/uptrace/bun"
 	"scrumlr.io/server/common"
+	"scrumlr.io/server/identifiers"
 )
 
-type DB struct {
+type SessionDB struct {
```
Planlos5000: Rename to `DB` to be consistent

server/src/sessions/service.go
```diff
@@ -4,378 +4,535 @@ import (
...
+	Get(ctx context.Context, board, user uuid.UUID) (DatabaseBoardSession, error)
+	GetAll(ctx context.Context, board uuid.UUID, filter ...BoardSessionFilter) ([]DatabaseBoardSession, error)
+	GetUserConnectedBoards(ctx context.Context, user uuid.UUID) ([]DatabaseBoardSession, error)
 }
 
-type Service struct {
-	database       UserDatabase
-	sessionService SessionService
-	realtime       *realtime.Broker
+type BoardSessionService struct {
```
Planlos5000: Rename to `Service` to be consistent

src/store/features/board/thunks.ts
```diff
@@ -51,6 +55,27 @@ export const leaveBoard = createAsyncThunk("board/leaveBoard", async () => {
...
+};
+
+const mapParticipantsWithUsers = async (message: BoardInitEvent): Promise<Participant[]> => {
+  const {participants} = message.data;
+  const asDTOs = participants as unknown as ParticipantDTO[];
+  const mapped = await Promise.all(asDTOs.map(dtoToParticipant));
+  message.data.participants = mapped;
+  return mapped;
+};
+
```
ghost: Why don't we use batch fetching here instead of querying every user one after the other? With 10 participants, that's 10 separate API calls that could potentially be just one. What do you think about
   the performance impact when boards get busy?

  Similar pattern in the websocket handlers around line 178-186
  I see we're doing the same sequential fetching in multiple places. Have you considered if there's a way we could consolidate this logic?

src/api/participant.ts
```diff
@@ -121,4 +121,19 @@ export const ParticipantsAPI = {
       throw new Error(`unable to update ready states: ${error}`);
     }
   },
+
+  getUserById: async (userId: string) => {
```
ghost, line 125: What happens if getUserById fails for one participant? Will it break the entire participant list update?

src/store/features/participants/types.ts
```diff
@@ -12,6 +12,16 @@ export interface Participant {
...
+export interface ParticipantDTO {
+  id: string;
+  connected: boolean;
+  ready: boolean;
+  raisedHand: boolean;
+  showHiddenColumns: boolean;
+  role: ParticipantRole;
+  banned?: boolean;
+}
+
```
ghost, line 24: What do you think about refactoring this to:   
```
// (currently ParticipantDTO)
export interface ParticipantWithUserId {
    userId: string;  
    connected: boolean;
    ready: boolean;
    raisedHand: boolean;
    showHiddenColumns: boolean;
    role: ParticipantRole;
    banned?: boolean;
  }

// currently Participant
  export interface ParticipantWithUser {
    user: Auth;  // I also don't like the interface naming "Auth" but this isn't your problem.
    connected: boolean;
    ready: boolean;
    raisedHand: boolean;
    showHiddenColumns: boolean;
    role: ParticipantRole;
    banned?: boolean;
  }
```

I think, this is easier to understand.  What is you opinion about this @Schwehn42?

src/store/features/board/thunks.ts
ghost: If we going over to a batch API we could do something like this, to fix both, the performance and possible issues when participants aren't loading. 

```
const mapParticipantsOptimized = async (dtos: ParticipantDTO[]): Promise<Participant[]> => {
    try {
      const userIds = dtos.map(dto => dto.id);
      const users = await API.getUsersByIds(userIds); 

      return dtos.map(dto => {
        const user = users.find((u: Auth) => u.id === dto.id);
        if (!user) {
          console.warn(`User not found for participant ${dto.id}`);
          return null;
        }
        return { user, ...dto };
      }).filter(Boolean) as Participant[];
    } catch (error) {
      console.error('Failed to map participants with users:', error);
      return [];
    }
  };
```

mateo-ivc: Good point, didn't think of this one. 
I would still keep fetch single user since there are situations where only one user has to be fetched

server/src/-d "postgres:/admin:supersecret@localhost:5432/scrumlr?sslmode=disable" -v --disable-check-origin --insecure --allow-anonymous-custom-templates/go_build_scrumlr_io_server
Planlos5000, line 1: Please remove the binary

server/src/api/router.go
```diff
@@ -260,6 +262,8 @@ func (s *Server) protectedRoutes(r chi.Router) {
 		r.Route("/user", func(r chi.Router) {
 			r.Get("/", s.getUser)
 			r.Put("/", s.updateUser)
+			r.Get("/{user}", s.getUserByID)
+			r.Post("/", s.getMultipleUserByID)
```
Planlos5000, line 266: Why is this a `POST` request. This should be a `GET` because you are not adding new users

server/src/api/users.go
```diff
@@ -33,14 +36,66 @@ func (s *Server) getUser(w http.ResponseWriter, r *http.Request) {
...
+		common.Throw(w, r, err)
+		return
+	}
+
+	render.Status(r, http.StatusOK)
+	render.Respond(w, r, user)
+}
+
+func (s *Server) getMultipleUserByID(w http.ResponseWriter, r *http.Request) {
+	ctx, span := tracer.Start(r.Context(), "getMultipleUser")
```
Planlos5000: Change the span name to `scrumlr.users.api.get.all`

server/src/api/users.go
```diff
@@ -33,14 +36,66 @@ func (s *Server) getUser(w http.ResponseWriter, r *http.Request) {
 	render.Respond(w, r, user)
 }
 
+func (s *Server) getUserByID(w http.ResponseWriter, r *http.Request) {
+	ctx, span := tracer.Start(r.Context(), "scrumlr.users.api.update")
```
Planlos5000: Change the span name to `scrumlr.users.api.get`

server/src/sessions/database.go
```diff
@@ -1,176 +1,213 @@
...
-func (db *DB) CreateAnonymousUser(ctx context.Context, name string) (DatabaseUser, error) {
-	var user DatabaseUser
-	insert := DatabaseUserInsert{Name: strings.TrimSpace(name), AccountType: common.Anonymous}
-	_, err := db.db.NewInsert().
-		Model(&insert).
-		Returning("*").
-		Exec(ctx, &user)
-
-	return user, err
+func (database *DB) Create(ctx context.Context, boardSession DatabaseBoardSessionInsert) (DatabaseBoardSession, error) {
```
Planlos5000, line 24: The create method has been changed with #5446

server/src/users/database.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (db *DB) GetMultipleUsers(ctx context.Context, ids []uuid.UUID) ([]DatabaseUser, error) {
+	var users []DatabaseUser
+
+	query := db.db.NewSelect().
+		Model(&users)
+
+	query = query.Where("id IN (?)", bun.In(ids))
+
+	err := query.Scan(ctx, &users)
```
Planlos5000: Can this be one statement?

server/src/users/dto.go
```diff
@@ -0,0 +1,45 @@
...
+	return u
+}
+
+func (*User) Render(_ http.ResponseWriter, _ *http.Request) error {
+	return nil
+}
+
+type UserID struct {
+	UserID uuid.UUID `json:"id"`
+}
```
Planlos5000: I think this struct is not needed

server/src/users/service.go
Planlos5000, line 1: The OTel stuff is missing in this file

server/src/api/router.go
mateo-ivc: RFC 9110 states that request bodys on GET, HEAD, and DELETE are not interoperable. While it is possible to use it on a GET with our implementation, there are scenarios where clients/proxies drop the request body entirely

Planlos5000: Good point. But i would prefer to make it a GET request. What about putting the ids in the URL. This would result in a very long URL that could be to long, but if we chunk the request in like chunks of 30 ids per requests, it should work and we have enough room for the rest of the url

mateo-ivc: That would mean that we have to do multiple requests. Which leads to the problem, how we act if one request fails?
I think is approach is more make the code more complex than it has to be.

mateo-ivc:
because of huge merge conflicts moved to #5510

Issue #5404, opened 2025-09-12
Cards not sorted after end of voting


mateo-ivc:
### The problem

After testing some other features, it was discovered that the cards are not sorted by number of votes after a poll has been completed. We have encountered this behaviour before.

I'm not entirely sure whether this is a backend or frontend issue, but I have observed that the cards are updated in the correct order when the website is reloaded or when another user joins the board. This could lead to the conclusion that the cards are not being returned in the correct order (by the backend) when a vote has been updated.

### Browser

firefox, chromium

### Steps to reproduce the behavior

1. Open board
2. add a few notes
3. start voting and add votes
4. close voting

### Screenshots

<img width="975" height="400" alt="Image" src="https://github.com/user-attachments/assets/e9ef4df7-d84a-4716-bb78-43aa79dd6f1d" />

Schwehn42:
did some digging:
1. frontend doesn't seem to explicitly sort by rank, but adheres to the notes state order
2. backend does set the rank correctly after a voting concludes
3. no `NOTES_SYNC` ~or `NOTES_UPDATED`~ is sent after a voting concludes

note: when a participant joins/reloads, `NOTES_SYNC` is sent, so it will actually sort correctly for all connected participants as a side effect.

~possible solution: broadcast `NOTES_SYNC` with updated rank after voting concludes successfully~

edit: actually `NOTES_UPDATED` **is sent** after a voting concludes. but it doesn't contain the correct ranks.

PR #5443, opened 2025-09-22
fix: Sort cards after voting

mateo-ivc:
## Description

Description of the problem can be found here: #5404

Fix inside of `votings/service.go`
added logic to sort the notes before sending the results to the frontend 
## Checklist


mateo-ivc:
> Are the tests now running reliably?

Tbh i am now sure why they failed. The only difference between the passing branch and failing branch were some package updates. 
And the test only failed because the response was in the wrong order 🤷‍♂️

PR #5464, opened 2025-09-29
refactor: move logic from note database to note service

Planlos5000:
## Description

Move some logic from the note database layer to the note service and add more unit tests.
Remove vote service as a dependency from the note service.
## Changelog

- move logic from database to service
- add more unit tests
- remove vote service dependency
- adjust database tests
- adjust integration tests
## Checklist


server/src/notes/service.go
```diff
@@ -108,9 +124,46 @@ func (service *Service) Update(ctx context.Context, user uuid.UUID, body NoteUpd
...
+	)
+
+	precondition, err := service.database.GetPrecondition(ctx, body.ID, body.Board, user)
+	if err != nil {
+		span.SetStatus(codes.Error, "failed to get preconditions")
+		span.RecordError(err)
+		return nil, common.InternalServerError
+	}
+
+	if user != precondition.Author && precondition.CallerRole == common.ParticipantRole {
```
mateo-ivc: This would also mean users are not allowed to move notes right?
In our own retros we as normal users also move notes from time to time (duplicates). Do we want to keep that?

mateo-ivc:
While dragging a note (from another user) onto another note, we expect that either the note gets stacked or stays where it was since you are not allowed to modify the not of another user. 
What actually happens when you don't have the permission -> the note vanishes. It looks like a the frontend expects the note to be stacked but the backend doesn't allow it. So either frontend changes are needed or we allow the notes of other users to be moved.

server/src/notes/service.go
Planlos5000: I have updated the check so that participants can move notes

Schwehn42: this should be a board option maybe

Planlos5000: I created an the issue #5495 for that

Issue #5445, opened 2025-09-22
Delete Account


SelinaBuff:

jmoehler:
change check button to radio button

PR #5534, opened 2025-10-22
feat: delete user

Planlos5000:
## Description

Add endpoint to delete a user for #5445. This will delete the user and cascading the board templates, sessions, notes and votes.

## Changelog

- Add delete user method
- Add tests for delete user method
- Add delete endpoint 

## Checklist

### DIFF
--- a/server/src/api/router.go
+++ b/server/src/api/router.go
@@ -262,6 +262,7 @@ func (s *Server) protectedRoutes(r chi.Router) {
 			r.Get("/", s.getUser)
 			r.Get("/{user}", s.getUserByID)
 			r.Put("/", s.updateUser)
+			r.Delete("/", s.deleteUser)
 			r.With(s.BoardParticipantContext).Get("/board/{id}", s.getUsers)
 		})
 	})
--- a/server/src/api/users.go
+++ b/server/src/api/users.go
@@ -121,3 +121,23 @@ func (s *Server) updateUser(w http.ResponseWriter, r *http.Request) {
 	render.Status(r, http.StatusOK)
 	render.Respond(w, r, updatedUser)
 }
+
+func (s *Server) deleteUser(w http.ResponseWriter, r *http.Request) {
+	ctx, span := tracer.Start(r.Context(), "scrumlr.users.api.delete")
+	defer span.End()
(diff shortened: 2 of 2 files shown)


server/src/sessions/database_test.go
```diff
@@ -208,6 +208,17 @@ func (suite *DatabaseUserTestSuite) Test_Database_GetUser_NotFound() {
...
+	t := suite.T()
+	database := NewUserDatabase(suite.db)
+
+	userId := suite.users["Delete"].ID
+
+	err := database.DeleteUser(context.Background(), userId)
+
+	assert.Nil(t, err)
+}
+
```
BingeCode: I think it would make sense to also test whether the user and all their related data does not exist in the db anymore - not just if the DeleteUser call did not throw an error. For all we know in this test, anything could happen in DeleteUser. 
So I would set up the test such that a bunch of related data is inserted at the beginning and then check if it got deleted as expected. Perhaps also multiple scenarios that reflect common real-world usage (for example having notes/votes on other boards). This way we can catch some more bugs should anything change in the future (in the db schema for example).

Planlos5000: Yes i see your point. Unfortunately  the test structure does not allow to test for such things.

BingeCode (approved):

server/src/users/service.go
```diff
@@ -1,407 +1,429 @@
...
+
+	err := service.database.Delete(ctx, id)
+	if err != nil {
+		span.SetStatus(codes.Error, "failed to delete user")
+		span.RecordError(err)
+		log.Errorw("failed to delete user", "user", id, "err", err)
+		return common.InternalServerError
+	}
+
+	deletedUserCounter.Add(ctx, 1)
```
mateo-ivc, line 313: do we want to send a user_delete event to the clients ?
I think the only advantage would be, that the boards get update automatically

Planlos5000: Could be a good idea to send some kind of event.
If a user is on a board and deletes the account, the session should be disconnected. Also each board, where the user was connected to, should be reloaded, because the notes of the user are also deleted.

Planlos5000:
Will be reopened by @mateo-ivc

Issue #5353, opened 2025-09-01
Server component


SelinaBuff:
Talk to @Schwehn42 (his idea: Vite Framework)

brandstetterm:
Please also think about if migrating to Next.js would be an option

Planlos5000:
The reason this came up was that we need some place to store some environment variables.
Currently we store env variables in the cookies but there the can be read from everybody, which is not that great for secret tokens like the Unleash token. Is it possible for that reason to introduce server components to store the tokens on the server and not on the client

PR #5580, opened 2025-11-11
refactor: migrate legacy SASS code and refactor SASS file structure

### LINKED ISSUE #5569: SASS improvements
Working on #5548 I noticed a few things that could be improved with our current SASS setup:

1. Migrate to `@use` as `@import` is [deprecated and will be removed with Sass v3](https://sass-lang.com/documentation/breaking-changes/import/) (this will also get rid of some warnings on `yarn start`)
2. Separate the constants (aka variables) from mixins and functions (static code vs. dynamic code) to avoid code duplication
3. Convert style.scss and colors.scss to [Partials](https://sass-lang.com/documentation/at-rules/import/#partials) as well as any new files.
4. Solve `Deprecation The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.`
5. Solve deprecation of red(), blue() and green()

BingeCode:
Closes #5569

## Description
This is a DX-oriented refactor.

1) It migrates all legacy SASS code:
- Converts all [`@import`](https://sass-lang.com/documentation/at-rules/import/) statements to `@use` statements
- The import/use URLs were streamlined as well
- Converts all deprecated color functions

2) It reorganizes the sass file structure for better performance and maintainability:
- Rename `style.scss` to `styles.scss` for consistency
- `styles.scss` and `colors.scss` were made [Partials](https://sass-lang.com/documentation/at-rules/import/#partials) as is convention within Sass (e.g. `_styles.scss` and `_colors.scss`)
- Mixins and functions (dynamic code) were extracted from `_colors.scss` and `_styles.scss` into a new `_mixins.scss`
- Some variables were converted to mixins where it made sense

## Changelog
See commits

## Checklist

## (Optional) Visual Changes
None


src/constants/_mixins.scss
```diff
@@ -0,0 +1,170 @@
+@use "sass:list";
+@use "sass:color";
+
+@use "constants/styles";
+@use "constants/colors";
```
Schwehn42: maybe this is an IDE issue but it reprimands that here, the path cannot be resolved (can be fixed by either removing `"constants/"`) or prefixing `"src/"`

BingeCode: If we're talking about file resolution with the editor (clicking on the URL to get to the imported file), then only omitting `constants/` works in VSCode. Prefixing `src/` is not recognized by VSCode (VSCode apparently [uses relative imports to resolve SASS URLs](https://github.com/microsoft/vscode/issues/153237); [source code](https://github.com/microsoft/vscode-css-languageservice/blob/main/src/services/scssNavigation.ts)).

However, omitting `constants/` in any other importing file under `components/` leads to `SassError: Can't find stylesheet to import.`

Also, when switching to the "modern" dart implementation of SASS, the URLs are not being recognized anyway ([webpack config](https://webpack.js.org/loaders/sass-loader/#api)). So this is something we will have to take a look at during migration to Vite ([#5353](https://github.com/inovex/scrumlr.io/issues/5353)).

We should later find a solution how to configure Vite such that both the SASS engine and the IDE resolve @use URLs correctly.

Schwehn42: should be fixed with the migration then 👍

PR #5584, opened 2025-11-12
fix: note deletion

### LINKED ISSUE #5559: Clicking on "Delete note" button does not trigger UI to refresh
### The problem

When trying to delete a note, clicking on the note does not refresh the UI (closing the dialogs, refreshing the view). However, when closing the dialogs manually and refreshing the page, the note is gone.

This bug seems to be introduced with the latest version 2.4.1.

### Browser

Firefox 143.0.4, Safari 18.6

- Create new board
- Add note
- Click on note
- Click on trash icon
- Click on "Delete note" button

### Screenshots

Image

### Additional context

_No response_

Schwehn42:
## Description
Fixes #5559

As of recent releases, deleting notes would not result in a correct state update.  
This is caused by backend changes introduced with #5464, where instead of a struct in the form of 
```typescript
{
  noteId: string;
  deleteStack: boolean;
}
```

we now have a simple
```typescript
notes: string[]
```

Meaning that in case the user requested a stack deletion, the backend will now return all notes affected by this.  
Thus, the deletion had to be refactored accordingly.

The reducer will now iteratively delete notes one by one.
Redux batches this as one operation, so the state only updates after the fact.

## Changelog

- adjust action data type to match with backend
- synchronously calls delete action for each note
- adjusted and fixed deletion logic
- refactored payload type where needed
- test: make reducers available in mock store
- test: add reducer tests

## Checklist


Planlos5000:
How complicated would it be to write tests for the store, so that we would notice such bugs earlier?

Schwehn42:
> How complicated would it be to write tests for the store, so that we would notice such bugs earlier?

ill have a look into it @Planlos5000

mateo-ivc (approved):
lgtm

PR #5640, opened 2025-12-03
feat: delete user

mateo-ivc:
## Description

Added an API entrypoint and logic to delete a user account. **only backend**
## Changelog

`realtime/boards.go`
- added new websocket event -> USER_DELETED

`user/*`
- api entrypoint
- service logic
- database query to delete a user
- also added tests
## Checklist

### DIFF
--- a/server/src/users/api.go
+++ b/server/src/users/api.go
@@ -1,335 +1,335 @@
 package users
 
 import (
-	"context"
-	"errors"
-	"net/http"
+  "context"
+  "errors"
+  "net/http"
 
-	"github.com/go-chi/chi/v5"
-	"github.com/go-chi/render"
-	"github.com/google/uuid"
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/codes"
-	"scrumlr.io/server/common"
-	"scrumlr.io/server/identifiers"
-	"scrumlr.io/server/logger"
-	"scrumlr.io/server/sessions"
+  "github.com/go-chi/chi/v5"
+  "github.com/go-chi/render"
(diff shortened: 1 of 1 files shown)


server/src/users/api.go
```diff
@@ -1,278 +1,338 @@
...
+func (api *API) isAuthorizedUser(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx, span := tracer.Start(r.Context(), "scrumlr.user.api.context.is_authorized_user")
+		defer span.End()
+		log := logger.FromContext(ctx)
+
+		userIDValue := ctx.Value(identifiers.UserIdentifier)
+		userID, ok := userIDValue.(uuid.UUID)
+		userIDString := userID.String()
+		fmt.Printf("%s", userIDString)
```
Planlos5000: Either use the logger or remove

server/src/users/routes.go
```diff
@@ -14,6 +14,7 @@ type UsersApi interface {
 	Update(w http.ResponseWriter, r *http.Request)
 	Delete(w http.ResponseWriter, r *http.Request)
 
+	isAuthorizedUser(next http.Handler) http.Handler
```
Planlos5000: I don't know if this is the right name for this method

server/src/users/api.go
mateo-ivc: ups, was for debugging purposes only :D

server/src/users/routes.go
mateo-ivc: True didn't have a better name in my mind, but i am open to suggestions

Issue #5635, opened 2025-11-27
"Export PDF" broken


SelinaBuff:
### The problem:

The PDF-download also doesn't work (like the Markdown). It shows only a white screen (see screenvideo):

### Browser

Chrome, Safari 

https://github.com/user-attachments/assets/94538f7d-6ac3-4133-81da-3f83d03ae5d0

Schwehn42:
this error (and likely #5627, too) are caused by the user/session split (#5510). since the `exportBoard` api call does not return full users with names and such anymore, but just the IDs. 

easiest solution is probably to fetch the after the fact, using the `mapSingleParticipant` or `mapMultipleParticipants` functions

PR #5641, opened 2025-12-03
fix: export of board #5635

mateo-ivc:
## Description

Problem for the export was, that the backend is now only sending the userId for a session, frontend is missing on username and avatar.
## Changelog

added a fix that fetches the corresponding user objects, merges them correctly together and returns the correct object
## Checklist


mateo-ivc:
@Schwehn42 Do we have a shared place for functions that are needed in multiple parts of the code?
Fetching and merging the user is now happening in several places.

src/components/SettingsDialog/ExportBoard/PrintView/PrintView.tsx
```diff
@@ -7,9 +7,10 @@ import {Close, Printer} from "components/Icon";
 import {useTranslation} from "react-i18next";
 import classNames from "classnames";
 import {getColorClassName} from "constants/colors";
-import {ExportBoardDataType, compareNotes, getAuthorName, getChildNotes, getNoteVotes} from "utils/export";
+import {compareNotes, ExportBoardDataType, getAuthorName, getChildNotes, getNoteVotes} from "utils/export";
 import {DEFAULT_URL} from "constants/misc";
 import "./PrintView.scss";
+import {Auth, ParticipantWithUserId} from "../../../../store/features";
```
Schwehn42: relative path

src/components/SettingsDialog/ExportBoard/PrintView/PrintView.tsx
```diff
@@ -36,7 +37,21 @@ export const PrintView = ({boardId, boardName}: PrintViewProps) => {
 
   const getBoardData = async () => {
     const response = await API.exportBoard(boardId, "application/json");
-    return response.json();
+    const jsonResponse = await response.json();
+    const userWithUser = await API.getUsers(jsonResponse.board.id);
```
Schwehn42: all API calls should be typed; but `ExportBoardDataType` still has the old data type for participants (`ParticipantWithUser[]` instead of `ParticipantsWithUserId[]`).

so, `API.exportBoard()` should return either return old type, and then be converted to the correct type like you do here, or move the transformation to inside the api call itself

src/components/SettingsDialog/ExportBoard/PrintView/PrintView.tsx
```diff
@@ -36,7 +37,21 @@ export const PrintView = ({boardId, boardName}: PrintViewProps) => {
...
+      const user = userWithUser.find((u: Auth) => u.id === participant.id);
+      return {
+        user: {...user},
+        connected: participant.connected,
+        ready: participant.ready,
+        raisedHand: participant.raisedHand,
+        showHiddenColumns: participant.showHiddenColumns,
+        role: participant.role,
+        banned: participant.banned,
+      };
```
Schwehn42: correct me if I'm wrong, but isn't the following the same?

```suggestion
      return {
        user,
        ...participant
      };
```

src/utils/export.ts
```diff
@@ -110,6 +111,20 @@ const mdTemplate = (boardData: ExportBoardDataType) =>
...
+    return {
+      user: {...user},
+      connected: participant.connected,
+      ready: participant.ready,
+      raisedHand: participant.raisedHand,
+      showHiddenColumns: participant.showHiddenColumns,
+      role: participant.role,
+      banned: participant.banned,
+    };
+  });
```
Schwehn42: duplicate as above, consider moving to the API itself or somewhere else

Schwehn42:
also the test doesn't function anymore, but the test doesn't seem too feasible, so maybe we can consider skipping it and replace it with a proper integration test

src/utils/test/updated-dummy-board-data.json
BingeCode, line 1: Kleinigkeit, aber: Warum das renaming zu _updated_-dummy-board-data?

mateo-ivc: Da sich die Participant Struktur geändert hat, hab ich diese sicherheitshalber in einer neuen Datei angelegt. Damit die alten Tests nicht auf einmal fehlschlagen. Habe das aber zurückgeändert, da die Strukturänderung keine Probleme gemacht hat.

PR #5653, opened 2025-12-09
refactor: new auth provider

### LINKED ISSUE #5645: replace markbates/goth package.
The markbates/goth package depends on an other deprecated package and has not been updated for some time.
Since it is used for authentication, it would be better to switch to a package that is more up to date and maintained more frequently.

Therefore i suggest using https://github.com/go-pkgz/auth

mateo-ivc:
## Checklist


k8s/deployment.yaml
Planlos5000, line 1: Revert this file. We don't want  a valid cert for each deployment

server/docs/development.md
```diff
@@ -54,3 +54,43 @@ go run . --config config_example.toml
 ```
 
 To see all values that can be set and what purpose they serve, take a look at the provided [config_example.toml](./../config_example.toml) file.
+
+## Local testing of auth providers
+
+In order to use of provider in a local test environment following arguments must be set either in the CLI arguments or in the .env file:
```
Planlos5000: `In order to use *a* provider in a local test environment *the* following arguments ...`

server/docs/development.md
```diff
@@ -54,3 +54,43 @@ go run . --config config_example.toml
...
+
+### Needed for all providers
+```
```
Planlos5000, line 68: Add a new line above and below code blocks and annotate them with e.g. `bash`/`sh` or `ini`

server/docs/development.md
```diff
@@ -54,3 +54,43 @@ go run . --config config_example.toml
...
+-auth-callback-host     = http://localhost:8080
+-session-secret         = <own session secret>
```
Planlos5000: How can a session secret be generated?

server/docs/development.md
```diff
@@ -54,3 +54,43 @@ go run . --config config_example.toml
...
+```
+
+### GitHub
+```
+-auth-github-client-id        =	<github client id>
+-auth-github-client-secret    = <github client secret>
+```
+
+### OIDC Dex
+Start the docker with the OIDC profile
```
Planlos5000: `Start the docker compose file`

server/src/api/board_reaction.go
Planlos5000, line 1: Revert the file. It should still be `var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")`

server/src/api/board_session_requests.go
Planlos5000, line 1: Revert the file. It should still be var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")

server/src/api/board_templates.go
Planlos5000, line 1: Revert the file. It should still be var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")

server/src/api/boards.go
Planlos5000, line 1: Revert the file. It should still be var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")

server/src/api/column_templates.go
Planlos5000, line 1: Revert the file. It should still be var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")

server/src/api/columns.go
Planlos5000, line 1: Revert the file. It should still be var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")

server/src/api/feedback.go
Planlos5000, line 1: Revert the file. It should still be var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")

server/src/api/health.go
Planlos5000, line 1: Revert the file. It should still be var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")

server/src/api/notes.go
Planlos5000, line 1: Revert the file. It should still be var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")

server/src/api/reactions.go
```diff
@@ -5,17 +5,13 @@ import (
...
-	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/codes"
-	"go.opentelemetry.io/otel/trace"
 	"scrumlr.io/server/common"
 	"scrumlr.io/server/identifiers"
 	"scrumlr.io/server/logger"
 	"scrumlr.io/server/reactions"
 )
 
-var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")
```
Planlos5000, line 17: Should we remove the tracer from all api files where it is commented out?

server/src/api/votings.go
Planlos5000, line 1: Revert the file. It should still be var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	url := config.AuthCodeURL(state)
+	http.Redirect(w, r, url, http.StatusTemporaryRedirect)
+}
+
+func (api *API) Callback(w http.ResponseWriter, r *http.Request) {
+	log := logger.FromContext(r.Context())
+	ctx, span := tracer.Start(r.Context(), "scrumlr.auth.service.handle_callback")
+	defer span.End()
```
Planlos5000: Create first the span and reuse the `ctx` created from the span for the logger

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	http.Redirect(w, r, targetURL, http.StatusSeeOther)
+}
+
+func NewAuthApi(service AuthService, userService users.UserService) AuthApi {
+	api := new(API)
+	api.service = service
+	api.userService = userService
+	return api
+}
```
Planlos5000: Most of the time we have the constructor at the top of the file

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+	"go.opentelemetry.io/otel/codes"
+	"scrumlr.io/server/common"
+	"scrumlr.io/server/logger"
+	"scrumlr.io/server/users"
+)
+
+type API struct {
+	service     AuthService
+	userService users.UserService
+	basePath    string
```
Planlos5000: Th `basePath` is never set

server/src/auth/api_test.go
```diff
@@ -0,0 +1,153 @@
...
+  "scrumlr.io/server/users"
+)
+
+type AuthTestSuite struct {
+  suite.Suite
+}
+
+func TestAuthTestSuite(t *testing.T) {
+  suite.Run(t, new(AuthTestSuite))
+}
```
Planlos5000: The suite is not used

server/src/auth/router.go
```diff
@@ -0,0 +1,40 @@
...
+	})
+
+	return router
+}
+
+func NewAuthRouter(authAPI AuthApi) *Router {
+	r := new(Router)
+	r.authAPI = authAPI
+	return r
+}
```
Planlos5000: `New` functions at the top

server/src/auth/service.go
```diff
@@ -0,0 +1,437 @@
...
+  }
+  if err := a.initializeJWTAuth(); err != nil {
+    return nil
+  }
+
+  return a
+}
+
+func (a *AuthConfiguration) GetConfig(provider string) *oauth2.Config {
+  return a.oauthConfigs[provider]
```
Planlos5000: What happens if the provider does not exist?

Maybe use the two-value assignment here https://go.dev/blog/maps

server/src/auth/service.go
```diff
@@ -0,0 +1,437 @@
...
+    }
+    a.oauthConfigs[string(common.TypeOIDC)] = &oauth2.Config{
+      ClientID:     p.ClientId,
+      ClientSecret: p.ClientSecret,
+      Endpoint:     endpoint,
+      RedirectURL:  p.RedirectUri,
+      Scopes:       []string{"openid", "profile", "email"},
+    }
+    var err error
+    oidcJWKS, err = keyfunc.NewDefault([]string{"http://127.0.0.1:5556/dex/keys"})
```
Planlos5000: This is now hard coded for dex. What if i want to use keycloak or something else? What if dex is not running at localhost or on another port?

server/src/auth/service.go
```diff
@@ -0,0 +1,437 @@
...
+  }
+
+  return a
+}
+
+func (a *AuthConfiguration) GetConfig(provider string) *oauth2.Config {
+  return a.oauthConfigs[provider]
+}
+
+func (a *AuthConfiguration) initializeProviders() error {
```
Planlos5000, line 79: Add the otel tracer and maybe add some debug log

server/src/auth/service.go
```diff
@@ -0,0 +1,437 @@
...
+        ctx := r.Context()
+
+        var authToken jwx.Token
+        var err error
+
+        if authToken, err = jwtauth.VerifyRequest(a.unsafeAuth, r, jwtauth.TokenFromCookie); err == nil {
+          // check if user tries to authenticate by a prior authentication key
+          // attempt to migrate JWT to new key
+          userID := authToken.PrivateClaims()["id"].(string)
+          var user uuid.UUID
```
Planlos5000: Are there better names? Both value are later the userId

server/src/auth/service.go
```diff
@@ -0,0 +1,437 @@
...
+  var kf jwt.Keyfunc
+  switch provider {
+  case string(common.Google):
+    kf = googleJWKS.Keyfunc
+  case string(common.AzureAd), string(common.Microsoft):
+    kf = microsoftJWKS.Keyfunc
+  case string(common.TypeOIDC):
+    kf = oidcJWKS.Keyfunc
+  case string(common.Apple):
+    return nil, errors.New("not implemented")
```
Planlos5000: Why not add a key func for apple?

server/src/logger/logger.go
Planlos5000, line 1: Why do we need a tmp logger?

Planlos5000 (changes requested):

server/src/auth/service.go
```diff
@@ -0,0 +1,437 @@
...
+  var internalUser *users.User
+  switch userInfo.Provider {
+  case common.Google:
+    internalUser, err = a.userService.CreateGoogleUser(ctx, userInfo.Ident, userInfo.Name, userInfo.AvatarURL)
+  case common.GitHub:
+    internalUser, err = a.userService.CreateGitHubUser(ctx, userInfo.Ident, userInfo.Name, userInfo.AvatarURL)
+  case common.TypeOIDC:
+    internalUser, err = a.userService.CreateOIDCUser(ctx, userInfo.Ident, userInfo.Name, userInfo.AvatarURL)
+  case common.Microsoft:
+    internalUser, err = a.userService.CreateMicrosoftUser(ctx, userInfo.Ident, userInfo.Name, userInfo.AvatarURL)
```
Planlos5000: Missing the apple user

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+	ctx, span := tracer.Start(r.Context(), "scrumlr.auth.api.signin.anonymous")
+	defer span.End()
+	log := logger.FromContext(ctx)
+
+	var body AnonymousSignUpRequest
+	if err := render.Decode(r, &body); err != nil {
+		span.SetStatus(codes.Error, "unable to decode body")
+		span.RecordError(err)
+		log.Errorw("unable to decode body", "err", err)
+		w.WriteHeader(http.StatusBadRequest)
```
wischoepke: error handling is not rly consistent here:
sometimes: "w.WriteHeader(http.StatusBadRequest)"
and then "common.Throw(w, r, common.InternalServerError)"

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+		return
+	}
+
+	user, err := api.userService.CreateAnonymous(ctx, body.Name)
+	if err != nil {
+		span.SetStatus(codes.Error, "failed to create anonyoums user")
```
wischoepke: typo

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+	basePath    string
+}
+
+// AnonymousSignUpRequest represents the request to create a new anonymous user.
+type AnonymousSignUpRequest struct {
+	// The display name of the user.
+	Name string
+}
+
+// signInAnonymously create a new anonymous user
```
wischoepke: i think some comments are not needed because the describe nothing new

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+	"scrumlr.io/server/users"
+)
+
+type API struct {
+	service     AuthService
+	userService users.UserService
+	basePath    string
+}
+
+// AnonymousSignUpRequest represents the request to create a new anonymous user.
```
wischoepke: i think some comments are not needed because the describe nothing new

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+	tokenString, err := api.service.Sign(map[string]interface{}{"id": user.ID})
+	if err != nil {
+		span.SetStatus(codes.Error, "failed to generate token string")
+		span.RecordError(err)
+		log.Errorw("unable to generate token string", "err", err)
+		common.Throw(w, r, common.InternalServerError)
+		return
+	}
+
+	cookie := http.Cookie{Name: "jwt", Value: tokenString, Path: "/", HttpOnly: true, MaxAge: math.MaxInt32}
```
wischoepke: if we create cookies on some other, not anonymous points with the same flags we should think about it (lifetime, CSRF protection -> SameSiteStrictMode etc or Secure: true)

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (api *API) BeginAuth(w http.ResponseWriter, r *http.Request) {
+	provider := strings.ToUpper(chi.URLParam(r, "provider"))
+	config := api.service.GetConfig(provider)
+
+	//get redirect url from query param
+	nonceBytes := make([]byte, 64)
+	if _, err := io.ReadFull(rand.Reader, nonceBytes); err != nil {
+		http.Error(w, "Internal error", http.StatusInternalServerError)
```
wischoepke: Internal error is not a meaningful message

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+	render.Status(r, http.StatusNoContent)
+	render.Respond(w, r, nil)
+}
+
+func (api *API) BeginAuth(w http.ResponseWriter, r *http.Request) {
+	provider := strings.ToUpper(chi.URLParam(r, "provider"))
+	config := api.service.GetConfig(provider)
+
+	//get redirect url from query param
+	nonceBytes := make([]byte, 64)
```
wischoepke, line 135: that is possible critical because CSRF protection is not made here...
You must store the generated nonce in an encrypted, HttpOnly cookie before redirecting the user. Your callback handler will then compare the state in the URL with the state in the cookie.

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+	if _, err := io.ReadFull(rand.Reader, nonceBytes); err != nil {
+		http.Error(w, "Internal error", http.StatusInternalServerError)
+		return
+	}
+	nonce := base64.URLEncoding.EncodeToString(nonceBytes)
+	returnURL := r.URL.Query().Get("state")
```
wischoepke, line 149: Possible Open Redirect attack: You need to validate returnURL before using it. Ensure it only contains a relative path (e.g., starts with /) or matches a strict whitelist of allowed domains.

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+		http.SetCookie(w, &cookieWithSubdomain)
+	}
+
+	render.Status(r, http.StatusNoContent)
+	render.Respond(w, r, nil)
+}
+
+func (api *API) BeginAuth(w http.ResponseWriter, r *http.Request) {
+	provider := strings.ToUpper(chi.URLParam(r, "provider"))
+	config := api.service.GetConfig(provider)
```
wischoepke: can GetConfig return nil pointer (maybe if the provider is wrong)? this will cause a possible nil pointer exception

server/src/auth/api.go
wischoepke: in or case relative check will be enough

server/src/auth/service.go
```diff
@@ -0,0 +1,457 @@
...
+	if !ok {
+		return errors.New("the provided keys are no ecdsa keys")
+	}
+
+	a.auth = jwtauth.New("ES512", privateKey, privateKey.PublicKey)
+	return nil
+}
+
+func (a *AuthConfiguration) HandleCallback(ctx context.Context, provider, code string) (*http.Cookie, *common.APIError) {
+	// Handles the the callback after login at a provider
```
wischoepke: typo

server/src/auth/service.go
```diff
@@ -0,0 +1,457 @@
...
+	ctx, span := tracer.Start(ctx, "scrumlr.auth.service.handle_callback")
+	log := logger.FromContext(ctx)
+	defer span.End()
+
+	span.SetAttributes(
+		attribute.String("scrumlr.auth.service.handle_callback.provider", provider),
+	)
+
+	config := a.GetConfig(provider)
```
wischoepke: if its possible that the config is null the service will crash here! (config.Exchange)

server/src/auth/service.go
```diff
@@ -0,0 +1,457 @@
...
+	if err != nil {
+		span.SetStatus(codes.Error, "failed to parse and verify id_token")
+		span.RecordError(err)
+		log.Errorf("failed to parse and verify id_token: %v", err)
+		return nil, errors.New("failed to verify id_token")
+	}
+
+	// access user info
+	var claims jwt.MapClaims
+	if claims, ok = parsedToken.Claims.(jwt.MapClaims); ok {
```
wischoepke, line 383: gemini answer: 
"The "<nil>" Identity Bug (Authentication Bypass / Data Corruption)
In fetchExternalUser, you use a type assertion to extract jwt.MapClaims. However, if that assertion fails, the claims variable remains nil.
In Go, reading from a nil map doesn't panic; it returns nil. If you pass nil to fmt.Sprint(), it returns the literal string "<nil>".
Because "<nil>" is not an empty string (""), your validation check (if res.Ident == "") will pass! This means malicious or malformed tokens could authenticate as a shared user whose ID is literally the string "<nil>"." 
looks like a real problem

server/src/auth/service_test.go
```diff
@@ -0,0 +1,289 @@
...
+  user, err := service.fetchExternalUser(context.Background(), string(provider), token)
+
+  assert.NoError(t, err)
+  assert.Equal(t, expectedUser.Ident, user.Ident)
+  assert.Equal(t, expectedUser.Name, user.Name)
+  assert.Equal(t, expectedUser.Provider, user.Provider)
+}
+func setupCookie(auth *jwtauth.JWTAuth, userID uuid.UUID) *http.Cookie {
+  _, tokenString, _ := auth.Encode(map[string]interface{}{"id": userID.String()})
+  return &http.Cookie{Name: "jwt", Value: tokenString}
```
wischoepke: a new cookie is created without the strictness of the old one... Secure:   true etc and lifetime

wischoepke: btw why we have so many points in code where we create a new cookie with different parameters etc. i think this should be encapsulated at one point with the important flags like Secure:   true, etc

server/src/auth/service.go
```diff
@@ -0,0 +1,457 @@
...
+var appleJWKS keyfunc.Keyfunc
+
+func NewAuthService(providers map[string]AuthProviderConfiguration, unsafePrivateKey, privateKey string, userService users.UserService) AuthService {
+	a := new(AuthConfiguration)
+	a.providers = providers
+	a.unsafePrivateKey = unsafePrivateKey
+	a.userService = userService
+	a.privateKey = privateKey
+	if err := a.initializeProviders(); err != nil {
+		return nil
```
wischoepke, line 63: why errors "silenced" here? pls add logging for this stuff :)

server/src/auth/service.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+	if err := a.initializeJWTAuth(); err != nil {
+		return nil
```
wischoepke, line 66: the same here

server/src/auth/service_test.go
```diff
@@ -0,0 +1,289 @@
...
+  "context"
+  "crypto/rand"
+  "crypto/rsa"
+  "net/http"
+  "net/http/httptest"
+  "testing"
+
+  "github.com/MicahParks/jwkset"
+  "github.com/MicahParks/keyfunc/v3"
+  "github.com/go-chi/jwtauth/v5"
```
wischoepke: why we using multiple jwt libraries and not only one :). it is a possible problem if we want to update etc

server/src/auth/service.go
```diff
@@ -0,0 +1,457 @@
...
+	if p, ok := a.providers[string(common.Google)]; ok {
+		a.oauthConfigs[string(common.Google)] = &oauth2.Config{
+			ClientID:     p.ClientId,
+			ClientSecret: p.ClientSecret,
+			RedirectURL:  p.RedirectUri,
+			Endpoint:     endpoints.Google,
+			Scopes:       []string{"openid", "email", "profile"},
+		}
+		var err error
+		googleJWKS, err = keyfunc.NewDefault([]string{"https://www.googleapis.com/oauth2/v3/certs"})
```
wischoepke: pls encapsulate the urls for the idps in separate consts

server/src/auth/api.go
```diff
@@ -0,0 +1,159 @@
...
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	state := r.FormValue("state")
+	code := r.FormValue("code")
+	cookie, err := api.service.HandleCallback(ctx, providerStr, code)
+	if err != nil {
+		span.SetStatus(codes.Error, "failed to handle callback")
+		span.RecordError(err)
+		log.Errorw("failed to handle callback", "err", err)
```
wischoepke, line 219: no header is set here. w.WriteHeader() is missing here

server/src/auth/service.go
mateo-ivc: would do this in another PR since it would make this PR more bloated than it already is. But will do debug logs!

server/src/auth/api.go
malterusko-ino: Within `common.Throw`, HTTP headers are set by the `go-chi/render` library.

server/src/auth/api.go
```diff
@@ -0,0 +1,257 @@
...
+		common.Throw(w, r, err)
+		return
+	}
+
+	SealCookie(r, cookie)
+	http.SetCookie(w, cookie)
+
+	targetURL := parts[1]
+	if target, safe := api.isSafeRedirect(targetURL); safe {
+		http.Redirect(w, r, target, http.StatusSeeOther)
```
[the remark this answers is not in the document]
malterusko-ino: The issue was resolved with the introduction of the isSafeRedirect validation logic. This function ensures that redirect targets are either:

1. Safe relative paths (starting with `/`, but excluding `//` to prevent protocol-relative redirects).
2. Are permitted external hosts that are matched against an explicit allowlist (allowRedirectHostnames).

Additionally, backslashes are normalized, and evasion attempts (e.g., leading backslashes) are intercepted.

server/src/auth/service.go
```diff
@@ -0,0 +1,414 @@
...
+
+  if err != nil {
+    span.SetStatus(codes.Error, "failed to create user")
+    span.RecordError(err)
+    return nil, common.InternalServerError
+  }
+
+  tokenString, _ := a.Sign(map[string]interface{}{"id": internalUser.ID})
+
+  cookie := CreateCookie("jwt", tokenString, "/", 121*24*60*60, nil) //maxAge = 21 Days
```
ffischbach: `121*24*60*60` = 121 Days not 21...

server/src/main.go
```diff
@@ -435,31 +453,44 @@ func run(ctx *cli.Context) error {
...
 
 	userService := initializer.InitializeUserService(sessionService, noteService)
-
 	keyWithNewlines := strings.ReplaceAll(ctx.String("key"), "\\n", "\n")
 	unsafeKeyWithNewlines := strings.ReplaceAll(ctx.String("unsafe-key"), "\\n", "\n")
-	authConfig, err := auth.NewAuthConfiguration(providersMap, unsafeKeyWithNewlines, keyWithNewlines, bun, userService)
+	authService := initializer.InitializeAuthService(providersMap, unsafeKeyWithNewlines, keyWithNewlines, userService)
 	if err != nil {
 		return fmt.Errorf("unable to setup authentication: %w", err)
 	}
```
ffischbach, line 461: Does this err check work as expected? initializer.InitializeAuthService returns nil instead of an error - like before.

server/src/auth/api.go
```diff
@@ -0,0 +1,250 @@
...
+	if err != nil {
+		log.Errorw("Missing oauth cookie", "err", err)
+		span.SetStatus(codes.Error, "missing oauth cookie")
+		span.RecordError(fmt.Errorf("missing oauth cookie"))
+		common.Throw(w, r, common.BadRequestError(fmt.Errorf("missing oauth cookie")))
+		return
+	}
+
+	//format: nonce__returnURL
+	parts := strings.Split(state, "__")
```
ffischbach, line 196: Since we expect the state to be split in (at least) three parts, we could add a check like ```  if len(parts) < 2 {common.Throw(w, r, common.BadRequestError(errors.New("invalid state format"))) return} ``` to prevent a index out of bounce exception. Event if it should never happen...

server/src/auth/service.go
```diff
@@ -0,0 +1,414 @@
...
+
+  var internalUser *users.User
+  switch userInfo.Provider {
+  case common.Google:
+    internalUser, err = a.userService.CreateGoogleUser(ctx, userInfo.Ident, userInfo.Name, userInfo.AvatarURL)
+  case common.GitHub:
+    internalUser, err = a.userService.CreateGitHubUser(ctx, userInfo.Ident, userInfo.Name, userInfo.AvatarURL)
+  case common.TypeOIDC:
+    internalUser, err = a.userService.CreateOIDCUser(ctx, userInfo.Ident, userInfo.Name, userInfo.AvatarURL)
+  case common.Microsoft:
```
ffischbach: Is `AZURE_AD` missing here? 
```
const (
	// Anonymous users don't require a registration
	Anonymous AccountType = "ANONYMOUS"

	// Google users registered on Google
	Google AccountType = "GOOGLE"

	// Microsoft users registered on Microsoft
	Microsoft AccountType = "MICROSOFT"

	// AzureAd users registered on Azure AD
	AzureAd AccountType = "AZURE_AD"

	// GitHub users registered on GitHub
	GitHub AccountType = "GITHUB"

	// Apple users registered on Apple
	Apple AccountType = "APPLE"

	// TypeOIDC users registered on OIDC
	TypeOIDC AccountType = "OIDC"
)
```

ffischbach:
```
case common.AzureAd:
   internalUser, err = a.userService.CreateAzureAdUser(ctx, userInfo.Ident, userInfo.Name, userInfo.AvatarURL)
```

server/src/auth/service.go
```diff
@@ -0,0 +1,414 @@
...
+
+  }
+
+  if err != nil {
+    span.SetStatus(codes.Error, "failed to create user")
+    span.RecordError(err)
+    return nil, common.InternalServerError
+  }
+
+  tokenString, _ := a.Sign(map[string]interface{}{"id": internalUser.ID})
```
ffischbach: shouldn't we handle a potential signing err here, to avoid returning 200 with an empty token?

server/src/auth/router.go
```diff
@@ -0,0 +1,40 @@
...
+		router.Post("/anonymous", r.authAPI.SignInAnonymously)
+
+		router.Route("/{provider}", func(router chi.Router) {
+			router.Get("/", r.authAPI.BeginAuth)
+			router.Get("/callback", r.authAPI.Callback)
+		})
+	})
+
+	return router
+}
```
ffischbach, line 41: Are we aware, that we are removing the `AnonymousLoginDisabledContext` middleware?

Image

server/src/main.go
malterusko-ino: For now, I've set the error check to nil. I'll add this to the error-handling ticket and implement an appropriate error.

server/src/auth/api.go
malterusko-ino: Good catch, thanks!

server/src/auth/router.go
malterusko-ino: Good catch! I've added the middleware back.

PR #5675, opened 2025-12-17
feat: 5445 delete account

jmoehler:
## Description

added account deletion functionality
## Changelog
- added user deletion api logic
- added user deletion ui components
- added user deletion text in English, German and French

## Checklist

## Visual Changes

before:
image

after:
image

image


src/components/ConfirmationDialog/ConfirmationDialog.scss
```diff
@@ -69,6 +69,102 @@
   letter-spacing: styles.$letter-spacing--small;
 }
 
+.confirmation-dialog__error-message {
+  display: flex;
+  align-items: center;
+  gap: 4px;
+  margin: 8px 0px 8px 0px;
```
Schwehn42: try using the `styles.$spacing-xyz` variables wherever applicable

src/components/SettingsDialog/Components/UserDeletion.scss
Schwehn42, line 1: - missing dark theme styles
- board delete button is bold, this one isn't -- should be consistent (either both bold or both not) imo

src/components/ConfirmationDialog/ConfirmationDialog.scss
```diff
@@ -69,6 +69,102 @@
...
+  color: styles.$critical-carmine--light;
+}
+.confirmation-dialog__error-message:empty {
+  visibility: hidden;
+}
+.confirmation-dialog__error-message > svg {
+  height: styles.$icon--medium;
+  width: styles.$icon--medium;
+  opacity: 0.8;
+  margin-left: calc(0.5em - (styles.$icon--medium / 2));
```
Schwehn42: scss get confused by the variable in conjunction with `/`, as this is not only used for division -- you should use either
- interpolation: `calc(0.5em - (#{$styles.$icon--medium} / 2))` 
- or possibly even more robust scss inbuilt helper functions from `@use sass:math`: `calc(0.5em - #{math.div(styles.$icon--medium, 2)})`

src/components/ConfirmationDialog/ConfirmationDialog.scss
```diff
@@ -69,6 +69,102 @@
...
+  margin: 0;
+  cursor: pointer;
+  font-size: styles.$text-size--medium;
+  letter-spacing: styles.$letter-spacing--small;
+}
+
+.confirmation-dialog__checkbox {
+  opacity: 0;
+  z-index: -1;
+}
```
Schwehn42, line 168: I'm not sure about this part, there's still an invisible checkbox (which is also clickable) right above the custom one

src/components/ConfirmationDialog/ConfirmationDialog.scss
```diff
@@ -69,6 +69,102 @@
...
+  background: styles.$gray--000;
+  cursor: pointer;
+  border-radius: 100%;
+  transition: all 0.16s ease-out;
+}
+
+.confirmation-dialog__checkbox-wrapper:hover .confirmation-dialog__checkbox-label::before {
+  box-shadow: 0 0 0 4px rgba(styles.$blue--500, 0.2);
+}
+
```
Schwehn42: take advantage of BEM nesting, like
```scss
.class {
  foo: bar;

  &--custom-state {
    some: propery;
  }
  
  &:hover {
    foo: baz;
  }
}
```

makes it often more readable and manageable

src/components/ConfirmationDialog/ConfirmationDialog.tsx
```diff
@@ -17,7 +17,13 @@ type ConfirmationDialogProps = {
   onExtraOptionLabel?: string;
   icon?: React.FunctionComponent<React.SVGProps<SVGSVGElement>>;
   warning?: boolean;
+  text?: string;
   className?: string;
+  // checkbox
+  checkboxLabel?: string;
+  checkboxChecked?: boolean;
+  onCheckboxChange?: (checked: boolean) => void;
+  noCheckboxError?: boolean;
```
Schwehn42: consider grouping these as an object, since they are all used together

src/i18n/de/translation.json
```diff
@@ -143,6 +143,11 @@
     "deleteNoteButton": "Kärtchen löschen",
     "deleteStackButton": "Stapel löschen",
     "deleteOptions": "Was möchtest du löschen?",
+    "deleteAccount": "Bist du sicher, dass du dein Konto löschen möchtest?",
+    "deleteAccountText": "Wenn du dein Konto löschst, werden alle damit verbundenen Daten ebenfalls dauerhaft entfernt, einschließlich: \n\n• deiner eigenen erstellten Vorlagen \n• deiner Notizen & Stimmen auf anderen Sitzungen \n\nDiese Aktion kann nicht rückgängig gemacht werden.",
```
Schwehn42: this works, but isn't the best practice.

maybe consider putting the translations values in an array, then mapping them like this:
```tsx
{t('deleteAccountItems', { returnObjects: true }).map((item) => (
    <li key={item}>{item}</li>
))}
```

src/components/ConfirmationDialog/ConfirmationDialog.scss
jmoehler: I now moved the invisible checkbox more to the center of the visible one. Removing the invisible checkbox is probably not a good idea, because then we would have to build the button manually.

src/components/ConfirmationDialog/ConfirmationDialog.tsx
```diff
@@ -44,7 +55,46 @@ export const ConfirmationDialog: FC<ConfirmationDialogProps> = (props) => {
...
+                        />
+                        <span className="confirmation-dialog__checkbox-label">{props.checkbox.label}</span>
+                      </label>
+                    )}
                   </div>
+                  {props.checkbox && (
+                    <div className="confirmation-dialog__error-message">
+                      {props.checkbox.showError && (
+                        <>
+                          <Warning className="" />
```
ghost: Can we get rid of `classname=""` here?

jmoehler: its gone now :)

PR #5722, opened 2026-01-06
Allow setting listen addr in SCRUMLR_SERVER_PORT

dani:
## Description
Long story short : I'm running scrumlr.io on Nomad, and expose it with the Consul Service Mesh. In this setup, security relies on every service listening only on 127.0.0.1 (to prevent bypassing the ACL defined in the mesh by random other containers which just happen to run on the same node). With this change, the SCRUMLR_SERVER_PORT accepts both a simple port (like before), or a addr:port string (like 127.0.0.1:8080)

## Changelog
Allow setting listen addr for the backend service

## Checklist

### DIFF
--- a/docs/src/content/docs/self-hosting/env-vars.md
+++ b/docs/src/content/docs/self-hosting/env-vars.md
@@ -80,6 +80,15 @@ The port on which the backend should listen for incoming connections.
 SCRUMLR_SERVER_PORT='8080'
 ```
 
+### Server Address
+
+The IP address on which the backend should listen for incoming connections.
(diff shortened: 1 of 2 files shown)


Planlos5000:
Hey @dani,
thanks for the PR.
I think it's not a good idea to configure the listen address through the `SCRUMLR_SERVER_PORT` env variable. I would prefer if the listen address gets its own env variable like `SCRUMLR_SERVER_LISTEN_ADDRESS` and the combine the address and port in the program.

dani:
Would something like this be better ?

Planlos5000 (approved):

PR #5737, opened 2026-01-14
feat: Change websocket lib from gorilla to coder

SimonRocker:
## Description
Changed websocket lib from gorilla to coder. 

## Changelog
- Removed gorilla
- Added coder

## Checklist

### DIFF
--- a/server/src/serviceinitialize/service.go
+++ b/server/src/serviceinitialize/service.go
@@ -1,143 +1,143 @@
 package serviceinitialize
 
 import (
-	"net/http"
-	"scrumlr.io/server/boards"
-	"scrumlr.io/server/hash"
-	"scrumlr.io/server/sessions"
-	"scrumlr.io/server/timeprovider"
-	"scrumlr.io/server/users"
-	"scrumlr.io/server/websocket"
+  "net/http"
+  "scrumlr.io/server/boards"
+  "scrumlr.io/server/hash"
+  "scrumlr.io/server/sessions"
+  "scrumlr.io/server/timeprovider"
+  "scrumlr.io/server/users"
+  "scrumlr.io/server/websocket"
 
-	"scrumlr.io/server/votings"
+  "scrumlr.io/server/votings"
 
-	"scrumlr.io/server/boardtemplates"
-	"scrumlr.io/server/columns"
-	"scrumlr.io/server/columntemplates"
-	"scrumlr.io/server/notes"
+  "scrumlr.io/server/boardtemplates"
+  "scrumlr.io/server/columns"
+  "scrumlr.io/server/columntemplates"
+  "scrumlr.io/server/notes"
 
-	"github.com/uptrace/bun"
-	"scrumlr.io/server/boardreactions"
-	"scrumlr.io/server/feedback"
-	"scrumlr.io/server/health"
-	"scrumlr.io/server/reactions"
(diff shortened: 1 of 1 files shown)


SimonRocker:
The sonarqube failure is quite a pain: I've move code in the test to the suite so that every test can use it. As the call to this method is the same for each test, it is marked as duplicate.Apart from that, it looks good from my side. WDYT?

server/src/websocket/coder_websocket.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (c *coderConnection) Close(reason string) error {
+	return c.conn.Close(websocket.StatusNormalClosure, reason)
+}
+
+type coderWebSocketService struct{}
+
+func NewCoderWebSocketService() WebSocketService {
+	return &coderWebSocketService{}
```
ghost: Since we’ve standardized on the Coder websocket library, is there any remaining value in keeping the 'Coder' prefix in our services, variables, and parameters? It might be cleaner to drop it.

server/src/websocket/websocket.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+	"net/http"
+)
+
+type MessageType int
+
+const (
+	MessageText MessageType = 1
+)
```
ghost, line 13: I suggest we also define MessageBinary here.

Even though we only use text right now, mirroring the constants from the coder package ensures we don't have 'magic numbers' if we ever receive a binary frame.

Reference from the library:

```
// MessageType represents the type of a WebSocket message.
// See https://tools.ietf.org/html/rfc6455#section-5.6
type MessageType int

// MessageType constants.
const (
	// MessageText is for UTF-8 encoded text messages like JSON.
	MessageText MessageType = iota + 1
	// MessageBinary is for binary messages like protobufs.
	MessageBinary
)
```

ghost:
Overall it looks good to me!

server/src/websocket/coder_websocket.go
SimonRocker: But would you do something like a WebsocketService or something? I kind of like that this allows us to see that this is the concrete implementation.

server/src/websocket/websocket.go
SimonRocker: Sure, added :+1:

server/src/websocket/coder_websocket.go
SimonRocker: I've changed it to websocket interface and websocket service. I hope that is a better naming.

server/src/serviceinitialize/service.go
```diff
@@ -42,10 +40,7 @@ func NewServiceInitializer(db *bun.DB, rt *realtime.Broker) ServiceInitializer {
 	initializer.hash = hash.NewHashSha512()
 	initializer.db = db
 	initializer.rt = rt
```
mateo-ivc: in [sessionrequests/service.go](https://github.com/inovex/scrumlr.io/pull/5737/files#diff-372e613cff779380200872038af54fc113f5e4e71c60b266c976ea7d688b3122)
.realtime is named broker, this is fine but i would prefer to have the same naming scheme over the project

SimonRocker: done :heavy_check_mark:

PR #5749, opened 2026-01-20
feat(docker): use distroless non-root base image

kersten:
## Description

switch to gcr.io/distroless/static:nonroot base image in Dockerfile to enhance security by running the application as a non-root user.

## Checklist

### DIFF
--- a/deployment/SKE/scrumlr_backend.yaml
+++ b/deployment/SKE/scrumlr_backend.yaml
@@ -37,7 +37,6 @@ spec:
             allowPrivilegeEscalation: false
           image: ghcr.io/inovex/scrumlr.io/scrumlr-server:sha-71c6812
           args:
-            - "/main"
             - "-disable-check-origin"
           resources:
             requests:
--- a/deployment/helm/scrumlr/values.yaml
+++ b/deployment/helm/scrumlr/values.yaml
@@ -127,7 +127,6 @@ backend:
     pullSecrets: []
     pullPolicy: IfNotPresent
     args:
-      - "/main"
       - "-disable-check-origin"
   ## @param backend.replicaCount Set the replica count for the backend
   ##
--- a/k8s/deployment.yaml
+++ b/k8s/deployment.yaml
@@ -35,7 +35,6 @@ spec:
             allowPrivilegeEscalation: false
           image: ghcr.io/inovex/scrumlr.io/scrumlr-server
           args:
-            - "/main"
             - "-disable-check-origin"
           resources:
             requests:
--- a/deployment/helm/scrumlr/README.md
+++ b/deployment/helm/scrumlr/README.md
@@ -56,7 +56,7 @@
 | `backend.image.tag`                          | The image tag to use                                                                        | `nil`                                     |
 | `backend.image.pullSecrets`                  | Docker registry secret names as an array                                                    | `[]`                                      |
 | `backend.image.pullPolicy`                   | The imagePullPolicy to use                                                                  | `IfNotPresent`                            |
-| `backend.image.args`                         | Aguments for the scrumlr backend                                                            | `["/main","-disable-check-origin"]`    |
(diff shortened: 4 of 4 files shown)


Planlos5000:
We have considered to change to a distroless version for the backend and have currently not decided what to do.
The problem here is that the healthcheck in the docker compose needs the `wget` command.
If we use a use a distroless version we should add a way to execute healthchecks in docker either by adding `wget`, a custom binary or adding a `--healthckeck` flag to the server, so that you could start the application a second time and it executes the health request.

kersten:
If you don't mind, I have a look at it and may come up with an idea to discuss about.

Planlos5000:
Yeah, sure. Thanks for looking into this.

kersten:
@Planlos5000 Could you check my integration? I created a new PR here: https://github.com/inovex/scrumlr.io/pull/5763

Planlos5000:
I will look into the PR

server/src/Dockerfile
```diff
@@ -9,11 +9,12 @@ COPY ./ .
...
-RUN mkdir /app
-WORKDIR /app
-COPY --from=0 /go/src/github.com/inovex/scrumlr.io/main ./
-EXPOSE 8080
+WORKDIR /
 
-CMD ["./main"]
+COPY --from=build /go/src/github.com/inovex/scrumlr.io/main .
+
+USER 65532:65532
```
Planlos5000: This is already the default user an the user does not have to be set (https://stackoverflow.com/questions/73568034/nobody-nonroot-user-in-distroless-images)

Could you add the `EXPOSE` for documentation

Planlos5000:
LGTM. We should first merge #5763 ?

kersten:
Yes, absolutely.

kersten:
@Planlos5000 Hey, I rebased the changes and fixed the conflicts. Also I moved everything to `/` instead `/app`. Therefore I changed the paths in multiple files to reflect the changes.

deployment/docker/docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     image: "ghcr.io/inovex/scrumlr.io/scrumlr-server:${SCRUMLR_BACKEND_VERSION:-4.4.0}"
     container_name: scrumlr-backend
     command:
-      - "/app/main"
+      - "/main"
```
Planlos5000: Now the call to `main` in the command is not needed because the docker image now has an entrypoint

kersten: Fixed it 👍🏻

k8s/deployment.yaml
```diff
@@ -35,7 +35,7 @@ spec:
             allowPrivilegeEscalation: false
           image: ghcr.io/inovex/scrumlr.io/scrumlr-server
           args:
-            - "/app/main"
+            - "/main"
```
Planlos5000: Here we also don't need the call to the binary, or do we?

deployment/helm/scrumlr/values.yaml
```diff
@@ -127,7 +127,7 @@ backend:
     pullSecrets: []
     pullPolicy: IfNotPresent
     args:
-      - "/app/main"
+      - "/main"
```
Planlos5000: Here we also don't need the call to the binary, or do we?

deployment/SKE/scrumlr_backend.yaml
```diff
@@ -37,7 +37,7 @@ spec:
             allowPrivilegeEscalation: false
           image: ghcr.io/inovex/scrumlr.io/scrumlr-server:sha-71c6812
           args:
-            - "/app/main"
+            - "/main"
```
Planlos5000: Here we also don't need the call to the binary, or do we?

kersten:
@Planlos5000 you are right, I cleaned the parts and updated the `README.md`

Planlos5000:
I would change the title of the PR since i don't think that it is a fix.

PR #5757, opened 2026-01-21
feat: 5213 new login page

jmoehler:
## Description
Implementation of the new login page design for scrumlr

## Changelog
- Redesigned `LoginBoard` to match the new UI specifications
- Refactored `HeaderBar` into a standalone reusable component
- Updated the `Boards` page to utilize the new `HeaderBar` component
- Enhanced the `Button` component: added a "ghost" button variant for better flexibility
- Updated `UserPill` to support a non-interactive (unclickable) state
- Added a small size variant to the `Input` bar
- Updated the styling for `loginProvider buttons
- Added standard button heights to global styles
- Updated localization (DE/EN/FR): added new texts, removed old texts
- Added new "Stan" illustrations/assets
- Removed the `HeroIllustration` component

## Checklist

## (Optional) Visual Changes
Desktop before:
image

Phone before:
image

Desktop after:

image

image

Phone after:
image

### DIFF
--- a/src/store/features/columns/reducer.ts
+++ b/src/store/features/columns/reducer.ts
@@ -12,10 +12,10 @@ export const columnsReducer = createReducer(initialState, (builder) =>
     .addCase(updatedColumns, (_state, action) => action.payload)
     // .toSpliced instead .splice because of Immer
     .addCase(createColumnOptimistically, (state, action) => state.toSpliced(action.payload.index, 0, action.payload))
-    .addCase(deleteColumnOptimistically, (state) => state.filter((c) => c.id !== TEMPORARY_COLUMN_ID))
+    .addCase(deleteColumnOptimistically, (state) => state.filter((c) => c.id !== TEMPORARY_COLUMN_ID).map((col, idx) => ({...col, index: idx})))
     .addCase(editColumnOptimistically, (state, action) => {
       const col = state.find((c) => c.id === TEMPORARY_COLUMN_ID);
       if (col) col.name = action.payload.column.name;
     })
-    .addCase(deletedColumn, (state, action) => state.filter((c) => c.id !== action.payload))
+    .addCase(deletedColumn, (state, action) => state.filter((c) => c.id !== action.payload).map((col, idx) => ({...col, index: idx})))
 );


src/assets/stan/Stan_Hanging_With_Coffee_Cropped_Light_Big.svg
Schwehn42, line 1: I don't see the reason why we need a big and normal version of each image, can't this be done with one each and potential css?

src/components/Button/Button.scss
```diff
@@ -27,12 +25,16 @@ $height--small: 46px;
   font-size: styles.$text-size--medium;
   font-weight: 700;
 
+  & > * {
+    pointer-events: none;
+  }
+
```
Schwehn42: what for? usually the universal selector (`*`) should be avoided.

src/components/Button/Button.tsx
```diff
@@ -23,28 +23,45 @@ type ButtonProps = {
   onClick?: (e: MouseEvent<HTMLButtonElement>) => void;
 
   dataCy?: string;
+
+  hideLabel?: boolean;
+
+  /** Used to remove button from keyboard navigation */
+  tabIndex?: number;
+  /** Used to hide button from screen readers */
+  "aria-hidden"?: boolean | "true" | "false";
```
Schwehn42: change name to `ariaHidden`, and type to `?: boolean`.
below, use `aria-hidden={Boolean(props.ariaHidden)}`

src/components/Button/Button.tsx
```diff
@@ -3,7 +3,7 @@ import {Color, getColorClassName} from "constants/colors";
 import {MouseEvent, ReactNode} from "react";
 import "./Button.scss";
 
-type ButtonType = "primary" | "secondary" | "tertiary";
+type ButtonType = "primary" | "secondary" | "tertiary" | "ghost";
```
Schwehn42: I don't understand what this new button type is supposed to achieve

src/components/Input/Input.scss
```diff
@@ -45,7 +49,11 @@ $padding--mobile: 12.5px styles.$spacing--base;
     // alternatively, this could be split up and put in their respective classes,
     // which might be more readable but also requires a little bit more code
     &:hover,
-    &:focus-within {
+    &:focus-within,
+
+    // special case for linked hover state (used in LoginBoard for anonymous name input)
+    [data-linked-hover="true"]:has(button:hover) &,
+    [data-linked-hover="true"]:has(button:active) & {
```
Schwehn42: I have multiple issues with this:
1. firstly, why is the special case needed? I guess because you want the text color to not look focused when it actually is. but this is a weird design consistency in comparison to the other inputs, so maybe it would be worth to ask design about this
2. using an attribute this way seems odd to me, since we generally use props and conditional classes for such things

src/components/Input/Input.scss
```diff
@@ -155,16 +163,27 @@ $padding--mobile: 12.5px styles.$spacing--base;
...
+        @include dark-active-styles;
+      }
+
+      // special case for linked hover state (used in LoginBoard for anonymous name input)
+      @at-root {
+        [theme="dark"] [data-linked-hover="true"]:has(button:hover) .input,
+        [theme="dark"] [data-linked-hover="true"]:has(button:active) .input {
+          @include dark-active-styles;
+        }
+      }
```
Schwehn42: same as above, I don't understand the need for this

src/components/Input/Input.tsx
```diff
@@ -18,13 +18,15 @@ type SearchBarProps = {
   disabled?: boolean;
   input: string;
   setInput: Dispatch<SetStateAction<string>>;
+  onKeyDown?: KeyboardEventHandler<HTMLInputElement>;
```
Schwehn42: consider hiding the pressing enter logic inside the input component, and then emitting a `()=>void` function named `onSubmit` or something instead (only if not disabled). that way not every consumer has to implement this themselves

src/components/LoginProviders/LoginProviders.tsx
```diff
@@ -1,53 +1,134 @@
...
-  }
+  const [isCompact, setIsCompact] = useState(false);
+  const containerRef = useRef<HTMLDivElement>(null);
+  const ghostRef = useRef<HTMLSpanElement>(null);
 
   const signIn = (provider: string) => async () => {
     await Auth.signInWithAuthProvider(provider, originURL);
   };
 
+  const providerConfig = {
```
Schwehn42: please define your own types/interfaces for this to create type safety. `providerConfig` can then be set to `Record<string, YourProviderConfigEntryType>`

src/components/LoginProviders/LoginProviders.tsx
```diff
@@ -1,53 +1,134 @@
...
+  } as const;
+
+  type ProviderKey = keyof typeof providerConfig;
+  const enabledProviders = providers.filter((provider): provider is ProviderKey => provider in providerConfig);
+
+  const primaryKey = enabledProviders[0];
+  const primaryData = primaryKey ? providerConfig[primaryKey] : null;
+
+  // useLayoutEffect prevents visual flickering by calculating width before paint
+  useLayoutEffect(() => {
```
Schwehn42: we have our own `useSize` hook for these cases, can this be applied here?

Schwehn42: also the whole thing seems kinda hacky, if its required fair enough, but maybe look into using container queries to simplify the whole thing?

src/components/UserPill/UserPill.tsx
```diff
@@ -21,6 +21,12 @@ export const UserPill = (props: UserPillProps) => {
     navigate(`${props.locationPrefix ?? ""}/settings/profile`);
   };
 
+  const executeClick = () => {
```
Schwehn42: weird function name imo

also, `onClick` doesn't fire anyway a native disabled button, so the is disabled check is redundant

src/constants/_styles.scss
```diff
@@ -113,3 +113,7 @@ $container__note: "note (max-width: 300px)";
 // shadows
 $box-shadow--light: 0px 5px 15px 0px rgba(0, 0, 0, 0.2);
 $box-shadow--dark: 0px 5px 15px 0px rgba(0, 0, 0, 0.35);
+
+// button heights
+$button-height--default: 52px;
+$button-height--small: 46px;
```
Schwehn42: putting these in `_styles.scss` only makes sense if the values are used in different places, which I'm not sure they are

src/routes/Boards/Templates/Templates.tsx
```diff
@@ -6,8 +6,8 @@ import {useTranslation} from "react-i18next";
 import {useEffect, useRef, useState} from "react";
 import {CreateTemplateCard, TemplateCard} from "components/Templates";
 // using a png instead of svg for now. reason being problems with layering
```
Schwehn42: obsolete comment

src/components/HeaderBar/HeaderBar.scss
```diff
@@ -0,0 +1,72 @@
+@use "sass:color";
+@use "constants/styles";
+
+$side-padding--desktop: 14vw;
+
+.headerBar {
```
Schwehn42: should be `header-bar`

src/components/LoginProviders/LoginProviders.tsx
```diff
@@ -1,53 +1,134 @@
...
 
+  const providerConfig = {
+    GOOGLE: {label: t("LoginProviders.signInWithGoogle"), icon: <Google className="login-providers__icon" />, signInKey: "google"},
+    MICROSOFT: {label: t("LoginProviders.signInWithMicrosoft"), icon: <Microsoft className="login-providers__icon" />, signInKey: "microsoft"},
+    AZURE_AD: {label: t("LoginProviders.signInWithAzureAd"), icon: <Azure className="login-providers__icon" />, signInKey: "azure_ad"},
+    APPLE: {label: t("LoginProviders.signInWithApple"), icon: <Apple className="login-providers__icon" />, signInKey: "apple"},
+    OIDC: {label: t("LoginProviders.signInWithOIDC"), icon: <OpenID className="login-providers__icon" />, signInKey: "oidc"},
+  } as const;
+
+  type ProviderKey = keyof typeof providerConfig;
```
Schwehn42, line 40: missing GitHub provider

src/assets/stan/Stan_Hanging_With_Coffee_Cropped_Light_Big.svg
jmoehler: fixed it, I initially had that in, because the template page needed a different size but I made some adjustments there and now it should work with just one size

src/components/Button/Button.scss
jmoehler: removed it, works without.

src/components/Button/Button.tsx
jmoehler: it is a white button that only has the hover in the given color. I use it for the non primary login providers.

src/components/Button/Button.tsx
jmoehler: adjusted

src/components/Input/Input.scss
jmoehler: I adjusted it to use a conditional class.

Selina wanted the input of the username in the anonymous setup to be active it the refresh button (the one to generate names) is hit. 
image

src/components/Input/Input.scss
jmoehler: see above

src/components/Input/Input.tsx
jmoehler: removed the onKeyDown and replaced it with on submit and a function inside input that now listens to enter and then fires the onSubmit

src/components/LoginProviders/LoginProviders.tsx
jmoehler: done

src/components/LoginProviders/LoginProviders.tsx
jmoehler: added

src/components/LoginProviders/LoginProviders.tsx
jmoehler: Now uses the useSize.

The thing is hacky. It calculates the space needed for the text of the primary button and then, according to the text width, calculates the breaking point between the full and the compact mode.

I first tried it with container queries. The problem is that depending on the language, we need more or less space for the text.

I could either use container queries with a fixed space for text or stay with the current approach.

src/components/UserPill/UserPill.tsx
jmoehler: since check is not needed, I removed the executeClick entirely.

src/constants/_styles.scss
jmoehler: they are being used in the Button.scss but I also use it to calculate the size of the overlayed button for the button of the primary LoginProvider in LoginProviders.scss

src/routes/Boards/Templates/Templates.tsx
jmoehler: removed

src/components/HeaderBar/HeaderBar.scss
jmoehler: adjusted

src/components/Button/Button.tsx
Schwehn42: there's still no need for the tertiary condition 

`props.ariaHidden ? true : undefined` and `props.ariaHidden` are the same, as you can see by the type definition

src/components/LoginProviders/LoginProviders.tsx
Schwehn42: its fine, looks better now with the use of our hook

src/components/LoginProviders/LoginProviders.tsx
Schwehn42: there's still an old comment saying " // Using Google icon as placeholder" there

jmoehler: done

src/components/Button/__tests__/Button.test.tsx
```diff
@@ -12,10 +12,10 @@ describe("Button", () => {
   });
 
   it("should render with different button types", () => {
-    const {rerender} = render(<Button type="secondary">Secondary</Button>);
+    const {rerender} = render(<Button variant="secondary">Secondary</Button>);
```
Schwehn42, line 15: the button changes affected other buttons, e.g. on the template page which don't look correct now

src/components/Button/Button.tsx
```diff
@@ -3,47 +3,97 @@ import {Color, getColorClassName} from "constants/colors";
...
 
-  dataCy?: string;
+  // Icon Props
+  icon?: ReactNode;
+  iconStyle?: IconStyle; // Defaults to 'default'
+  iconPosition?: IconPosition; // Defaults to 'right'
+  iconAlignment?: IconAlignment; // Defaults to 'compact' (close)
 
-  hideLabel?: boolean;
+  /** TODO remove tabIndex and ariaHidden as it will no longer be needed */
```
Schwehn42: will the TODO be done or no?

src/components/Button/Button.tsx
```diff
@@ -3,47 +3,97 @@ import {Color, getColorClassName} from "constants/colors";
...
+
+    if (isIconOnly) {
+      return iconMarkup;
+    }
+
+    // 4. Return them in the correct order
+    return (
+      // 1. Fixed spelling (alignment)
+      // 2. Added default value (iconAlign)
+      // 3. Added position class (iconPos) -> Critical for CSS Flex direction/margins
```
Schwehn42: remove AI comments

src/components/LoginProviders/LoginProviders.scss
```diff
@@ -1,5 +1,11 @@
 @use "constants/styles";
 
+//todo remove these later
```
Schwehn42, line 3: TODO yes or no?

src/components/Button/Button.tsx
jmoehler: is no longer needed, as this was only used when I stacked 2 buttons on top of each other

src/components/Button/__tests__/Button.test.tsx
jmoehler: yes, the changes were not done yet. now everything should work with the new implementation

src/components/Button/Button.tsx
jmoehler: should be gone now

src/components/Button/Button.tsx
jmoehler: gone

src/components/LoginProviders/LoginProviders.scss
jmoehler: I now removed them

Schwehn42 (approved):
lgtm 🧾

PR #5760, opened 2026-01-22
refactor: migrate to vite

### LINKED ISSUE #5619: Vite migration
CRA deprecated: [#5353](https://github.com/inovex/scrumlr.io/issues/5353)

Schwehn42:
## Description

**CRA is dead, long live Vite**

Closes #5619

We now use the modern and non-deprecated vite.
Loads of files had to be updated, but I'm sure its worth the increased compatibility and build speed.

## Changelog

- replace CRA with Vite
  - env var prefix
  - fix issues, specifically with different handling of *.svg files
- replace jest with vitest
  - update / fix all tests
- fix tsc errors
- adjust pipeline
- adjust deployment

## Checklist

### DIFF
--- a/server/src/realtime/board_sessions_requests_test.go
+++ b/server/src/realtime/board_sessions_requests_test.go
@@ -3,6 +3,7 @@ package realtime
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
@@ -164,4 +165,49 @@ func (suite *RealtimeBoardSessionRequestTestSuite) Test_Redis_BoardSessionReques
 	event := <-eventChannel
 	assert.NotNil(t, event)
 	assert.Equal(t, &eventType, event)
+}
+
+// Regression test for the same missing-return bug, applied to session request subscriptions.
+// Scenario: user and admin are both waiting on a board session request channel;
+// the user's context is cancelled, then a new request event is broadcast.
+func (suite *RealtimeBoardSessionRequestTestSuite) Test_Redis_BoardSessionRequest_FirstSubscriberContextCancelled_SecondSubscriberReceivesEvent() {
+	t := suite.T()
+	boardId := uuid.New()
+	userId1 := uuid.New()
+	userId2 := uuid.New()
+	eventType := RequestAccepted
+
+	broker, err := NewRedis(RedisServer{Addr: suite.redisConnectionString})
+	assert.Nil(t, err)
+
+	// Subscriber 1: user who will close the tab
+	ctx1, cancel1 := context.WithCancel(context.Background())
+	ch1 := broker.GetBoardSessionRequestChannel(ctx1, boardId, userId1)
+
+	// Subscriber 2: another user still waiting
+	ch2 := broker.GetBoardSessionRequestChannel(context.Background(), boardId, userId2)
+
+	// Subscriber 1 closes browser tab → context cancelled
+	cancel1()
+
+	// Wait until ch1 is closed to confirm the goroutine processed ctx.Done()
+	select {
+	case _, ok := <-ch1:
+		assert.False(t, ok, "ch1 should be closed after context cancellation")
+	case <-time.After(5 * time.Second):
+		t.Fatal("timeout: ch1 was not closed after context cancellation")
+	}
+
+	// A new session request event is broadcast for subscriber 2
+	// Without the fix: goroutine 1 panics with "send on closed channel"
+	err = broker.BroadcastUpdateOnBoardSessionRequest(context.Background(), boardId, userId2, eventType)
+	assert.Nil(t, err)
+
+	select {
+	case event := <-ch2:
+		assert.NotNil(t, event)
+		assert.Equal(t, &eventType, event)
+	case <-time.After(5 * time.Second):
+		t.Fatal("timeout: ch2 did not receive the expected event")
+	}
 }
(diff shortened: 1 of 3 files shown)


k8s/deployment.yaml
```diff
@@ -214,7 +214,7 @@ spec:
       services:
         - name: scrumlr-backend-PR_NUMBER
           port: 8080
-    - match: Host(`PR_NUMBER.development.scrumlr.fra.ics.inovex.io`) && (PathPrefix(`/static`) || PathPrefix(`/locales`) || PathPrefix(`/login`) || PathPrefix(`/board`) || PathPrefix(`/new`)  || PathPrefix(`/timer_finished.mp3`) || PathPrefix(`/hotkeys.pdf`) || PathPrefix(`/legal/`) || PathPrefix(`/emoji-data/`))
+    - match: Host(`PR_NUMBER.development.scrumlr.fra.ics.inovex.io`) && (PathPrefix(`/assets`, `/static`, `/locales`, `/login`, `/board`, `/new`, `/legal/`, `/emoji-data/`) || Path(`/timer_finished.mp3`, `/hotkeys.pdf`))
```
Planlos5000: Does this work. The [documentation](https://doc.traefik.io/traefik/reference/routing-configuration/http/routing/rules-and-priority/#path-pathprefix-and-pathregexp) does not say anything about the multiple routes in one `PathPrefix`

Schwehn42: I do believe it does, though there is no mention of it in the most recent docs.  
If you look at [an older documentation](https://doc.traefik.io/traefik/v2.2/routing/routers/#rule), there are in fact examples for using sequential paths.

So it seems to still be compatible, although the usage of regex seems to be favoured.
I can change it, but would need to know whether we're on API version v2.x or v3.x since the syntax slightly differs between the version (I think we're still on the old one? All the recent docs use `apiVersion: traefik.io/v1alpha1` instead of ours)

Schwehn42: so I just ended up using our old, but working path matching as regex matching in v2 does not work as expected. maybe we can try again after upgrading https://github.com/inovex/scrumlr.io/issues/5830

k8s/deployment.yaml
```diff
@@ -214,7 +214,7 @@ spec:
       services:
         - name: scrumlr-backend-PR_NUMBER
           port: 8080
-    - match: Host(`PR_NUMBER.development.scrumlr.fra.ics.inovex.io`) && (PathPrefix(`/static`) || PathPrefix(`/locales`) || PathPrefix(`/login`) || PathPrefix(`/board`) || PathPrefix(`/new`)  || PathPrefix(`/timer_finished.mp3`) || PathPrefix(`/hotkeys.pdf`) || PathPrefix(`/legal/`) || PathPrefix(`/emoji-data/`))
+    - match: Host(`PR_NUMBER.development.scrumlr.fra.ics.inovex.io`) && (PathPrefix(`/static`) || PathPrefix(`/locales`) || PathPrefix(`/login`) || PathPrefix(`/board`) || PathPrefix(`/new`) || PathPrefix(`/legal`) || PathPrefix(`/emoji-data`) || Path(`/timer_finished.mp3`) || Path(`/hotkeys.pdf`))
```
ffischbach, line 217: Shouldn't we configure the /assets directory here as well?

vite.config.ts
```diff
@@ -0,0 +1,40 @@
...
+      scss: {
+        loadPaths: [path.resolve(__dirname, 'src'),
+        ]
+
+      },
+    },
+  },
+  build: {
+    outDir: 'build',
+    assetsDir: 'static',
```
ffischbach, line 28: We configure here, that assets will move into 'static' directory after build. But in the Caddyfile both directories (/assets* and /static*) are referenced.

.husky/pre-commit
```diff
@@ -1,2 +1,2 @@
 echo "pre-commit: formatting code ..."
-yarn lint-staged
+lint-staged
```
ffischbach, line 2: can we really get rid of the 'yarn' prefix here?

Schwehn42: yep it works all the same, I think I read it in some documentation 😅 it runs directly from the local `.bin` instead of taking the extra route using any package manager

k8s/deployment.yaml
Schwehn42: no, explicitly not -- it can be misleading, but `/assets` is routed through the landing page. this is why I changed the vite assetsDir to `/static`, so there's no conflict there. you could also add a rule down at the landingpage matching for `/assets`, but currently it just takes whatever isn't matched by both front -and backend.

vite.config.ts
Schwehn42: good catch, you're right this was an oversight 👍

package.json
malterusko-ino: I would suggest removing it, since we now use vite.

```suggestion
```

package.json
```diff
@@ -89,27 +91,30 @@
...
     "prettier": "^3.8.1",
     "react-dnd-test-backend": "^15.1.2",
     "react-dnd-test-utils": "^15.1.2",
-    "react-scripts": "^5.0.1",
     "redux-mock-store": "^1.5.4",
     "resize-observer-polyfill": "^1.5.1",
     "sass": "^1.97.3",
     "stylelint": "^16.26.1",
     "stylelint-config-standard": "^39.0.1",
     "ts-jest": "^29.4.6",
```
malterusko-ino: I would suggest removing it, since we now use vite.


```suggestion
```

package.json
malterusko-ino: I would suggest removing it, since we now use vite and it is not used in `.eslintrc.json`.


```suggestion
```

malterusko-ino: I would suggest removing it, since we now use vite.


```suggestion
```

PR #5763, opened 2026-01-23
feat: integrate new healthcheck

kersten:
## Description

This PR replaces the shell-based Docker Compose healthcheck (`wget` via `CMD-SHELL`) with a small Go-based healthcheck binary.

The healthcheck is now built alongside the server and executed directly in the container, improving portability and ensuring consistent behavior across environments.

In addition, a dedicated healthcheck implementation and test suite was added to validate correct handling of base paths, ports, and non-OK responses.

## Changelog

- Replaced `wget`-based Docker Compose healthcheck with `CMD ["/healthcheck"]`
- Added new Go healthcheck binary (`server/src/cmd/healthcheck`) that calls the server `/health` endpoint
- Added unit tests for healthcheck behavior (default path, base path, invalid envs, non-OK response)
- Updated Dockerfile build step to compile and include the `healthcheck` binary in the server image

## Checklist

### DIFF
--- a/server/src/cmd/healthcheck/main.go
+++ b/server/src/cmd/healthcheck/main.go
@@ -42,7 +42,9 @@ func run() error {
 	if err != nil {
 		return fmt.Errorf("healthcheck failed: %w", err)
 	}
-	defer response.Body.Close()
+	defer func() {
+		_ = response.Body.Close()
+	}()
 
 	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusMultipleChoices {
 		return fmt.Errorf("healthcheck failed: %s", response.Status)


server/src/Dockerfile
```diff
@@ -6,14 +6,16 @@ ARG TARGETOS
...
+RUN GOOS=$TARGETOS GOARCH=$TARGETARCH CGO_ENABLED=0 go build -a -o main \
+    && CGO_ENABLED=0 GOOS=$TARGETOS GOARCH=$TARGETARCH go build -a -o healthcheck ./cmd/healthcheck
 
 # Start application
 FROM alpine:3.23
 
 RUN mkdir /app
 WORKDIR /app
 COPY --from=0 /go/src/github.com/inovex/scrumlr.io/main ./
+COPY --from=build /go/src/github.com/inovex/scrumlr.io/healthcheck .
```
Planlos5000: The `from=build` does not exist when I try to build the image. This should be `from=0`. Could you also add the `/` to the copy path?

deployment/docker/docker-compose.yml
```diff
@@ -46,7 +46,7 @@ services:
       nats:
         condition: service_started
     healthcheck:
-      test: ["CMD-SHELL", "wget --no-verbose --tries=1 --spider http://localhost:${SCRUMLR_SERVER_PORT:-8080}/api/health || exit 1"]
+      test: ["CMD", "/healthcheck"]
```
Planlos5000: This should be `/app/healthcheck` bcause the workdir is `/app`

server/src/cmd/healthcheck/main.go
```diff
@@ -0,0 +1,65 @@
...
+		healthPath = basePath + "/health"
+	}
+
+	url := fmt.Sprintf("http://%s:%d%s", address, port, healthPath)
+	client := http.Client{Timeout: 3 * time.Second}
+	response, err := client.Get(url)
+	if err != nil {
+		return fmt.Errorf("healthcheck failed: %w", err)
+	}
+	defer func() {
```
Planlos5000: Here a `defer reponse.Body.Close()` should be enough

server/src/cmd/healthcheck/main.go
```diff
@@ -0,0 +1,65 @@
...
+
+	address := strings.TrimSpace(os.Getenv("SCRUMLR_SERVER_LISTEN_ADDRESS"))
+	if address == "" {
+		address = "127.0.0.1"
+	}
+
+	basePath := strings.TrimSpace(os.Getenv("SCRUMLR_BASE_PATH"))
+	if basePath == "" {
+		basePath = "/"
+	}
```
Planlos5000: Would it be possible to write a helper function to get the env vars, that also accepts a default value like

```
port := GetEnv("SCRUMLR_SERVER_PORT", 8080)
address := GetEnv("SCRUMLR_SERVER_LISTEN_ADDRESS", "127.0.0.1")
```

server/src/cmd/healthcheck/main.go
kersten: I've did that because of your linter settings :)

```shell
cmd/healthcheck/main.go:56:27: Error return value of `response.Body.Close` is not checked (errcheck)
        defer response.Body.Close()
                                 ^
1 issues:
* errcheck: 1
```

I changed it anyway.

Planlos5000: If the linter wants it to be checked then put it in a defer func

PR #5784, opened 2026-02-02
chore: Refactored other integration tests and db test to use the temp…

SimonRocker:
Using templating in all the integration tests, including the db tests

## Description
Use the spiked db test templating everywhere.

## Changelog
Every integration and database test has been changed to use the db templating. Plus now there is no base seed for tests that don't use them.

## Checklist

### DIFF
--- a/server/src/initialize/testDbTemplates/pgtestdb_helper.go
+++ b/server/src/initialize/testDbTemplates/pgtestdb_helper.go
@@ -1,83 +1,83 @@
 package testDbTemplates
 
 import (
-	"context"
-	"log"
-	"sync"
-	"testing"
+  "context"
+  "log"
+  "sync"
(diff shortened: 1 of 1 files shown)


server/src/initialize/testDbTemplates/pgtestdb_helper.go
```diff
@@ -14,6 +14,11 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+const POSTGRES_IMAGE = "postgres:17.5-alpine"
```
Planlos5000: Maybe upgrade the docker image to `18.1-alpine`

SimonRocker: Sure, done!

PR #5792, opened 2026-02-05
Feat(4350): Add last modified at to board

SimonRocker:
## Description
Added a lastModifiedAt field to the board type. Also added it in the db. The field can now be requested by the frontend.

## Changelog
Added field and migration for lastModifiedAt field at the board. Added interface so that this field can be updated (set to now). Modified all tests to have this field and also test this functionality. 

## Checklist

### DIFF
--- a/server/e2e-tests/board_workflow_test.go
+++ b/server/e2e-tests/board_workflow_test.go
@@ -1,20 +1,21 @@
 package e2e
 
 import (
-  "errors"
-  "testing"
-  "time"
+	"errors"
+	"testing"
+	"time"
 
-  "github.com/google/uuid"
-  "github.com/stretchr/testify/assert"
-  "github.com/stretchr/testify/require"
-  "github.com/stretchr/testify/suite"
-  "scrumlr.io/server/boards"
-  "scrumlr.io/server/columns"
-  "scrumlr.io/server/notes"
-  "scrumlr.io/server/reactions"
-  "scrumlr.io/server/users"
-  "scrumlr.io/server/votings"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	"scrumlr.io/server/boards"
+	"scrumlr.io/server/columns"
+	"scrumlr.io/server/common"
+	"scrumlr.io/server/notes"
+	"scrumlr.io/server/reactions"
+	"scrumlr.io/server/users"
+	"scrumlr.io/server/votings"
 )
 
 // Helper functions for pointers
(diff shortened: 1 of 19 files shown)


server/src/boards/database.go
```diff
@@ -46,9 +47,10 @@ func (d *DB) UpdateBoardTimer(ctx context.Context, update DatabaseBoardTimerUpda
 }
 
 func (d *DB) UpdateBoard(ctx context.Context, update DatabaseBoardUpdate) (DatabaseBoard, error) {
+	update.LastModifiedAt = time.Now()
```
Planlos5000: Use the `timeprovider` so that `time.Now()` can be mocked

server/src/boards/database_test.go
```diff
@@ -1,477 +1,503 @@
...
+  t := suite.T()
+  database := NewBoardDatabase(suite.db)
+
+  boardId := suite.boards["UpdatePassphrase"].ID
+
+  boardBefore, err := database.GetBoard(context.Background(), boardId)
+  assert.Nil(t, err)
+
+  time.Sleep(10 * time.Millisecond)
+  timeBeforeUpdate := time.Now()
```
Planlos5000: If the `timeprovider` is used, this can be mocked and the test does not need to sleep

server/src/columns/service_test.go
```diff
@@ -7,427 +7,323 @@ import (
...
+func (suite *ColumnServiceTestSuite) expectColumnsUpdated(result DatabaseColumn, err error) {
+	suite.mockDB.EXPECT().Update(mock.Anything, DatabaseColumnUpdate{
+		ID:          suite.columnID,
+		Board:       suite.boardID,
+		Name:        suite.columnName,
+		Description: suite.columnDescription,
+	}).Return(result, err)
+	if err == nil {
+		suite.expectBroadcast()
+	}
```
Planlos5000, line 296: I would prefere if the call to the `suite.expectBroadcast()` must be called in the test function and is not hidden in the helper function

server/src/columns/service_test.go
```diff
@@ -7,427 +7,323 @@ import (
...
+}
 
-	mockColumnDatabase := NewMockColumnDatabase(t)
-	mockColumnDatabase.EXPECT().Count(mock.Anything, boardId).
-		Return(0, dbError)
+func (suite *ColumnServiceTestSuite) expectColumnCreated(insert DatabaseColumnInsert, result DatabaseColumn, err error) {
+	suite.mockDB.EXPECT().Create(mock.Anything, insert).Return(result, err)
+	if err == nil {
+		suite.expectBroadcast()
+	}
```
Planlos5000, line 303: I would prefere if the call to the `suite.expectBroadcast()` must be called in the test function and is not hidden in the helper function

server/src/boards/database.go
SimonRocker: Done :heavy_check_mark:

server/src/boards/database_test.go
SimonRocker: Done :heavy_check_mark:

server/src/columns/service_test.go
SimonRocker: I changed the name of the helper function to reflect this

server/src/columns/service_test.go
SimonRocker: I changed the name of the helper function to reflect this

server/src/boards/database.go
```diff
@@ -46,9 +49,10 @@ func (d *DB) UpdateBoardTimer(ctx context.Context, update DatabaseBoardTimerUpda
 }
 
 func (d *DB) UpdateBoard(ctx context.Context, update DatabaseBoardUpdate) (DatabaseBoard, error) {
+	update.LastModifiedAt = d.clock.Now()
 	query := d.db.NewUpdate().
 		Model(&update).
-		Column("timer_start", "timer_end", "shared_note")
+		Column("timer_start", "timer_end", "shared_note", "last_modified_at")
```
jmoehler: when you now call this function with only a subset of  
`timer_start", "timer_end", "shared_note", "last_modified_at"`, the others are being set to `NULL`.

This causes potentially running timers to be removed when anything on the board changes (due to the call for `"last_modified_at"`. Perhaps this even more side effects.

to make this save I suggest adjusting UpdateBoard to only update given columns plus `"last_modified_at"`.

server/src/boards/service.go
jmoehler, line 1: It would perhaps be cleaner to move this functionality in a database method (e.g. `UpdateLastModifiedOnly`), instead of wrapping the UpdateBoard here in the service?

server/src/notes/service.go
jmoehler, line 92: the import function currently doesn't call last modified at, meaning that if notes are imported, the board last modified date wouldn't change. I suggest also calling last modified at here

server/src/boards/last_modified_updater.go
jmoehler, line 1: if every service uses this updater, we probably do not even need the UpdateLastModified from the boards/service.go . I suggest removing the UpdateLastModified to avoid code drift and confusion. maybe adding a comment that explains where to find the updateLastModified

server/src/notes/service_test.go
```diff
@@ -2,993 +2,797 @@ package notes
...
 }
 
-func TestImport(t *testing.T) {
-	authorId := uuid.New()
-	boardId := uuid.New()
-	columnId := uuid.New()
-	rank := 0
-	noteId := uuid.New()
-	edited := false
+func (suite *NoteServiceTestSuite) TestImport() {
```
jmoehler: import tests currently do not check for last modified. this should be adjusted for all import tests.

server/src/common/simple_board_last_modified_updater.go
jmoehler, line 1: Makes sense as a quick solution, but this adds duplicate behavior and extra indirection, which may increase debugging complexity and code clutter over time.

server/src/boards/database.go
Planlos5000: You can also only use the `Set()` method to just set specific properties without changing the rest

jmoehler: I fixed it by making "timer_start", "timer_end", "shared_note" also conditional, which makes more sense logically.

server/src/boards/service.go
jmoehler: removed it, as it is never used. All references use the `last_modified_updater.go`

server/src/boards/last_modified_updater.go
jmoehler: done, see above

server/src/common/simple_board_last_modified_updater.go
jmoehler: ok for now

server/src/notes/service.go
jmoehler: fixed that.

server/src/notes/service_test.go
jmoehler: fixed

server/src/boards/database.go
```diff
@@ -46,10 +49,20 @@ func (d *DB) UpdateBoardTimer(ctx context.Context, update DatabaseBoardTimerUpda
 }
 
 func (d *DB) UpdateBoard(ctx context.Context, update DatabaseBoardUpdate) (DatabaseBoard, error) {
+	update.LastModifiedAt = d.clock.Now()
```
Planlos5000: This should not be in the database layer. This should happen in the service layer.

server/src/boards/database.go
malterusko-ino: lgtm

server/src/notes/service.go
malterusko-ino: lgtm

server/src/boards/last_modified_updater.go
malterusko-ino: lgtm

server/src/notes/service_test.go
malterusko-ino: lgtm

server/src/common/simple_board_last_modified_updater.go
malterusko-ino: lgtm

server/src/columns/service_test.go
malterusko-ino: lgtm

server/src/columns/service_test.go
malterusko-ino: lgtm

server/src/boards/database.go
malterusko-ino: lgtm

PR #5874, opened 2026-03-12
fix: preserve manually changed user name on OAuth re-login

malterusko-ino:
## Description

When an OAuth user already existed, the login flow updated the name in the users table with the provider's name, discarding any manually changed name. Now only the provider-specific table is updated on re-login, preserving the user's chosen name.

## Changelog

- `server/src/users/database.go`: When an OAuth user logs in again, their name in the users table is no longer overwritten with the provider's name. Instead, the existing user record is fetched as-is, preserving any manually changed name. Only the provider-specific table (e.g. google_users) is still updated with the latest provider data.
- `server/src/users/database_test.go`: Updated the test for re-login of an existing OAuth user to assert that the name in the users table is preserved rather than overwritten by the provider's name. 

## Checklist


Planlos5000:
Maybe an edge case that we have not defined and should talk about:

What happens if you don't rename your account that is linked with e.g. google and you change your name in the google account. 
I would expect that the updated google name would then be now my scrumlr user name. But now the name is not updated.

Issue #371, opened 2021-04-26
Prepare a Kubernetes instance


bitionaire:
Install & setup a Kubernetes instance on our own cloud as the deployment target for our application from #369.

AC:
- Kubernetes setup is implemented in code (e.g. Terraform)
- Deployment on a dev instance is setup in our GitHub Actions (CI/CD)

brandstetterm:
Is this task related to #430?

PR #5989, opened 2026-04-15
fix(i18n): detect language from URL query to support landing page language handoff

jmoehler:
## Description
Refactoring language detection of i18n to enable language pass through by landing page via URL

‼️**Note**: This merge request is the basis for the landing page adjustments in merge request [#371](https://github.com/inovex/scrumlr.io-landing-page/pull/371), where the selected language is now added to the request URL.

## Changelog
- Added URL-based language detection (lng) to frontend i18n and prioritized it over localStorage/browser language.

## Checklist


Schwehn42 (commented):
can you also do it the other way around? meaning changing the language within the frontend and clicking on the home logo should also return to the respective landing page language

jmoehler:
> can you also do it the other way around? meaning changing the language within the frontend and clicking on the home logo should also return to the respective landing page language

good point, I didn't have that in, should be fixed now.

src/components/HeaderBar/HeaderBar.tsx
```diff
@@ -2,6 +2,7 @@ import {useTranslation} from "react-i18next";
 import {ScrumlrLogo} from "components/ScrumlrLogo/ScrumlrLogo";
 import {UserPill} from "components/UserPill/UserPill";
 import "./HeaderBar.scss";
+import i18n from "i18n";
```
Schwehn42, line 5: instead of the extra import you can import `getI18n` in line 1 instead

src/components/HeaderBar/HeaderBar.tsx
```diff
@@ -19,7 +20,7 @@ export const HeaderBar = (props: HeaderBarProps) => {
         {/* logo - - - profile */}
         <div className="header-bar__scrumlr-logo-container">
           {/* this still needs fixing bc that uses other styles */}
-          <a className="new-board__scrumlr-logo-href" href="/" aria-label={t("BoardHeader.returnToHomepage")}>
+          <a className="new-board__scrumlr-logo-href" href={`/${i18n.language}/`} aria-label={t("BoardHeader.returnToHomepage")}>
```
Schwehn42, line 23: dynamic vars such as this should be visible at once, so I'd recommend extracting it to something like `const activeLanguage = getI18n().language` or maybe even `redirectHomeURL` below `const {t} ...` and then using the var below

Schwehn42 (commented):
1. note this does not work for deployments not employing the landing page, which is not only the case when running dev, but also non-official custom deployments 
2. there are possibly other redirects to the home page as well, like when deleting a board to give an example. at some point it might be even worth considering to make a util function to avoid duplicate code

src/components/HeaderBar/HeaderBar.tsx
jmoehler: fixed

src/components/HeaderBar/HeaderBar.tsx
jmoehler: adjusted

jmoehler:
> 1. note this does not work for deployments not employing the landing page, which is not only the case when running dev, but also non-official custom deployments
> 2. there are possibly other redirects to the home page as well, like when deleting a board to give an example. at some point it might be even worth considering to make a util function to avoid duplicate code



to 1: true I now fixed that by passing the language on as a query item rather than a subpage

2. I now created a central url handler. I might have still missed redirects. However, adding the language is mainly important when leaving the app not within the app. So for now that should be fine.

Schwehn42 (approved):
lgtm 🌍

PR #6030, opened 2026-04-28
build: upgrade cypress es5 -> esnext

Schwehn42:
## Description

with our recent TS upgrade, cypress no longer functions together as it still used es5 as the build target:

```
Error: Webpack Compilation Error
[tsl] ERROR
      TS5107: Option 'target=ES5' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
```

this PR fixes the issue by upgrading to the modern esnext build target.

## Changelog

`cypress/tsconfig.json`: upgrade build target and lib from `es5` to `esnext`

## Checklist

### DIFF
--- a/cypress/tsconfig.json
+++ b/cypress/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es5",
-    "lib": ["es5", "dom"],
+    "target": "esnext",
+    "lib": ["esnext", "dom"],
     "types": ["cypress", "node"],
     "resolveJsonModule": true,
     "esModuleInterop": true,


cypress/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es5",
-    "lib": ["es5", "dom"],
+    "target": "esnext",
```
Planlos5000, line 3: Why do we use here `esnext` and not `es6`?
The docs for [typescript](https://www.typescriptlang.org/tsconfig/#target) say that the `esnext` value can change between versions.

Schwehn42: because `esnext` is more modern than `es6`.

`esnext` has the latest features for us in development and allows for greatest speed, whereas the actual vite build target comes out to `['chrome111','edge111','firefox114','safari16.4','ios16.4']`, which is the so-called baseline (features that have been supported across all major engines for 2.5 years). all features that are "too modern" are transpiled/polyfilled, which guarantees for compatibility.

as for the cypress config, the same is the case (the runner currently uses chrome 147)

Schwehn42: I agree you're right though than since we're currently on node 20, something like `es202x`, which is also the default setting in the link you posted, would be safer.

but we'll implicitly fix that by upgrading to node 24, see https://github.com/inovex/scrumlr.io/issues/6036

Planlos5000 (approved):
lgtm

PR #6032, opened 2026-04-28
build: yarn berry

### LINKED ISSUE #5626: Migrate to Yarn Modern
It would make sense to migrate from Yarn Classic to Yarn Modern for a [number of benefits](https://yarnpkg.com/migration/overview).
Not listed there is better configurability via the .yarnrc.yml, for example disabling running user scripts on yarn install for the whole project (instead of having to pass a flag every time).

Schwehn42:
## Description

resolves https://github.com/inovex/scrumlr.io/issues/5626

## Changelog

- `yarn install` for both src and docs -> new package.json
- ci: 
  - enable `corepack` to activate yarn berry
  - `yarn install` with extra flags
  - remove caching for now

## Checklist


.yarnrc.yml
```diff
@@ -0,0 +1,6 @@
+approvedGitRepositories:
+  - "**"
+
+enableScripts: true
```
Planlos5000: Do we need this enabled? That's the dangerous part where e.g. sha1-hulud were executed

Schwehn42: 👍 you're right this was the autogenerated default, set to false now.

it's possible that some packages which do use post-install scripts (e.g. husky, sharp) might be affected by this. but if they're not its fine, otherwise a whitelist is still possible

Planlos5000 (approved):

PR #6069, opened 2026-05-06
fix: prevent host header injection and harden CI workflows, added tests to api boards

### LINKED ISSUE #6156: fix: prevent host header injection and harden CI workflows, added tests to api boards
## Description

This pull request introduces security hardening for our GitHub Actions workflows and patches a potential vulnerability in the Go backend API.

## Changelog

- CI: Mapped GitHub Actions inputs to environment variables and added -- to curl commands to prevent script and flag injection.
- CI: Updated the SonarQube Dependabot condition to safely check github.event.pull_request.user.login.
- API: Changed Location headers to use relative paths, preventing Host Header / Open Redirect vulnerabilities caused by trusting r.Host.

jmoehler:
## Description
This pull request introduces security hardening for our GitHub Actions workflows and patches a potential vulnerability in the Go backend API.

## Changelog
- CI: Mapped GitHub Actions inputs to environment variables and added `--` to `curl` commands to prevent script and flag injection.
- CI: Updated the SonarQube Dependabot condition to safely check `github.event.pull_request.user.login`.
- API: Changed `Location` headers to use relative paths, preventing Host Header / Open Redirect vulnerabilities caused by trusting `r.Host`.

## Checklist


server/src/api/boards.go
```diff
@@ -215,21 +215,31 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
...
 		if s.basePath == "/" {
-			w.Header().Set("Location", fmt.Sprintf("%s://%s/boards/%s/participants/%s", common.GetProtocol(r), r.Host, board, user))
+			w.Header().Set("Location", fmt.Sprintf("/boards/%s/participants/%s", boardID, userID))
 		} else {
-			w.Header().Set("Location", fmt.Sprintf("%s://%s%s/boards/%s/participants/%s", common.GetProtocol(r), r.Host, s.basePath, board, user))
+			w.Header().Set("Location", fmt.Sprintf("%s/boards/%s/participants/%s", s.basePath, boardID, userID))
 		}
 		w.WriteHeader(http.StatusCreated)
+		return nil
+	}
```
Planlos5000: Why is this function declared inside a function and not just a private function for the service?

sonar-project.properties
```diff
@@ -14,4 +14,4 @@ sonar.organization=scrumlr
 
 sonar.typescript.tsconfigPath=tsconfig.json
 sonar.go.coverage.reportPaths=server/src/coverage.txt
-sonar.coverage.exclusions=src/**/*,**/*_test.go,**/mock_*.go,mock_*.go,cypress/**/*
+sonar.coverage.exclusions=src/**/*,**/*_test.go,**/mock_*.go,mock_*.go,cypress/**/*,.github/**/*,**/*.yml,**/*.yaml
```
Planlos5000: Why should we exclude the whole `.github/` folder? SonarQube checks the workflows for security issues

jmoehler: I fixed that, that was still in there from my testing.

server/src/api/boards.go
jmoehler: I made that a function insight because its tightly coupled to joinBoard and the http state.

Planlos5000: Maybe we should go another route here and change the `basePath`. Currently the base path is `/` if no base path was provided and e.g. `/api/something` if a base path was provided.
I think we could get rid of that `if ... else ...` check for the base path, if we make sure the base path ends with a `/`.
I you could check the implications of that?

server/src/api/boards_test.go
```diff
@@ -305,6 +335,96 @@ func (suite *BoardTestSuite) TestJoinBoard() {
 	}
 }
 
+func (suite *BoardTestSuite) TestJoinBoard_RootBasePath_ExistingSessionRedirectsToParticipant() {
+	s, boardMock, sessionMock, sessionRequestMock, boardID, userID, req := suite.setupRootJoinBoardRequest()
```
Planlos5000, line 347: I would prefere to have the setup inside the method instead of a method call

server/src/api/boards_test.go
```diff
@@ -305,6 +335,96 @@ func (suite *BoardTestSuite) TestJoinBoard() {
...
+
+	suite.Equal(http.StatusSeeOther, rr.Result().StatusCode)
+	suite.Equal(fmt.Sprintf("/boards/%s/participants/%s", boardID, userID), rr.Result().Header.Get("Location"))
+	boardMock.AssertExpectations(suite.T())
+	sessionMock.AssertExpectations(suite.T())
+	sessionRequestMock.AssertExpectations(suite.T())
+}
+
+func (suite *BoardTestSuite) TestJoinBoard_RootBasePath_PublicCreateSessionErrorReturnsInternalServerError() {
+	s, boardMock, sessionMock, sessionRequestMock, boardID, userID, req := suite.setupRootJoinBoardRequest()
```
Planlos5000, line 363: I would prefere to have the setup inside the method instead of a method call

server/src/api/boards_test.go
```diff
@@ -305,6 +335,96 @@ func (suite *BoardTestSuite) TestJoinBoard() {
...
+	s.joinBoard(rr, req)
+
+	suite.Equal(http.StatusInternalServerError, rr.Result().StatusCode)
+	boardMock.AssertExpectations(suite.T())
+	sessionMock.AssertExpectations(suite.T())
+	sessionRequestMock.AssertExpectations(suite.T())
+}
+
+func (suite *BoardTestSuite) TestJoinBoard_RootBasePath_PublicCreatesParticipantLocation() {
+	s, boardMock, sessionMock, sessionRequestMock, boardID, userID, req := suite.setupRootJoinBoardRequest()
```
Planlos5000, line 381: I would prefere to have the setup inside the method instead of a method call

server/src/api/boards_test.go
```diff
@@ -305,6 +335,96 @@ func (suite *BoardTestSuite) TestJoinBoard() {
...
+
+	suite.Equal(http.StatusCreated, rr.Result().StatusCode)
+	suite.Equal(fmt.Sprintf("/boards/%s/participants/%s", boardID, userID), rr.Result().Header.Get("Location"))
+	boardMock.AssertExpectations(suite.T())
+	sessionMock.AssertExpectations(suite.T())
+	sessionRequestMock.AssertExpectations(suite.T())
+}
+
+func (suite *BoardTestSuite) TestJoinBoard_RootBasePath_ByInviteExistingRequestLocation() {
+	s, boardMock, sessionMock, sessionRequestMock, boardID, userID, req := suite.setupRootJoinBoardRequest()
```
Planlos5000, line 400: I would prefere to have the setup inside the method instead of a method call

server/src/api/boards_test.go
```diff
@@ -305,6 +335,96 @@ func (suite *BoardTestSuite) TestJoinBoard() {
...
+
+	suite.Equal(http.StatusSeeOther, rr.Result().StatusCode)
+	suite.Equal(fmt.Sprintf("/boards/%s/requests/%s", boardID, userID), rr.Result().Header.Get("Location"))
+	boardMock.AssertExpectations(suite.T())
+	sessionMock.AssertExpectations(suite.T())
+	sessionRequestMock.AssertExpectations(suite.T())
+}
+
+func (suite *BoardTestSuite) TestJoinBoard_RootBasePath_ByInviteCreatesRequestLocation() {
+	s, boardMock, sessionMock, sessionRequestMock, boardID, userID, req := suite.setupRootJoinBoardRequest()
```
Planlos5000, line 418: I would prefere to have the setup inside the method instead of a method call

server/src/api/boards_test.go
```diff
@@ -305,6 +335,96 @@ func (suite *BoardTestSuite) TestJoinBoard() {
 	}
 }
 
+func (suite *BoardTestSuite) TestJoinBoard_RootBasePath_ExistingSessionRedirectsToParticipant() {
```
Planlos5000: Could the name of the test be shorter? Why is there the `RootBasePath` in the name?

server/src/api/boards_test.go
```diff
@@ -305,6 +335,96 @@ func (suite *BoardTestSuite) TestJoinBoard() {
...
+	s.joinBoard(rr, req)
+
+	suite.Equal(http.StatusSeeOther, rr.Result().StatusCode)
+	suite.Equal(fmt.Sprintf("/boards/%s/participants/%s", boardID, userID), rr.Result().Header.Get("Location"))
+	boardMock.AssertExpectations(suite.T())
+	sessionMock.AssertExpectations(suite.T())
+	sessionRequestMock.AssertExpectations(suite.T())
+}
+
+func (suite *BoardTestSuite) TestJoinBoard_RootBasePath_PublicCreateSessionErrorReturnsInternalServerError() {
```
Planlos5000: Could the name of the test be shorter? Why is there the `RootBasePath` in the name?

server/src/api/boards_test.go
```diff
@@ -305,6 +335,96 @@ func (suite *BoardTestSuite) TestJoinBoard() {
...
+	rr := httptest.NewRecorder()
+	s.joinBoard(rr, req)
+
+	suite.Equal(http.StatusInternalServerError, rr.Result().StatusCode)
+	boardMock.AssertExpectations(suite.T())
+	sessionMock.AssertExpectations(suite.T())
+	sessionRequestMock.AssertExpectations(suite.T())
+}
+
+func (suite *BoardTestSuite) TestJoinBoard_RootBasePath_PublicCreatesParticipantLocation() {
```
Planlos5000: Could the name of the test be shorter? Why is there the `RootBasePath` in the name?

server/src/api/boards_test.go
```diff
@@ -305,6 +335,96 @@ func (suite *BoardTestSuite) TestJoinBoard() {
...
+	s.joinBoard(rr, req)
+
+	suite.Equal(http.StatusCreated, rr.Result().StatusCode)
+	suite.Equal(fmt.Sprintf("/boards/%s/participants/%s", boardID, userID), rr.Result().Header.Get("Location"))
+	boardMock.AssertExpectations(suite.T())
+	sessionMock.AssertExpectations(suite.T())
+	sessionRequestMock.AssertExpectations(suite.T())
+}
+
+func (suite *BoardTestSuite) TestJoinBoard_RootBasePath_ByInviteExistingRequestLocation() {
```
Planlos5000: Could the name of the test be shorter? Why is there the `RootBasePath` in the name?

server/src/api/boards_test.go
```diff
@@ -305,6 +335,96 @@ func (suite *BoardTestSuite) TestJoinBoard() {
...
+	s.joinBoard(rr, req)
+
+	suite.Equal(http.StatusSeeOther, rr.Result().StatusCode)
+	suite.Equal(fmt.Sprintf("/boards/%s/requests/%s", boardID, userID), rr.Result().Header.Get("Location"))
+	boardMock.AssertExpectations(suite.T())
+	sessionMock.AssertExpectations(suite.T())
+	sessionRequestMock.AssertExpectations(suite.T())
+}
+
+func (suite *BoardTestSuite) TestJoinBoard_RootBasePath_ByInviteCreatesRequestLocation() {
```
Planlos5000: Could the name of the test be shorter? Why is there the `RootBasePath` in the name?

server/src/api/boards.go
jmoehler: Good point. Normalizing base path once is probably the cleaner direction, but it has quite some impact across all redirects and path handling. I would say we put that in a separate follow-up ticket.

For this ticket, the main focus is reducing risk rather than changing behavior

server/src/api/boards_test.go
jmoehler: Since the same setup is used five times in a row, I kept it in a helper to reduce repetition and keep the tests cleaner. I now moved the helper closer to the tests that use it, so the context stays local and easier to follow.

server/src/api/boards_test.go
jmoehler: I adjusted the test names to be shorter and clearer.

server/src/api/boards.go
mateo-ivc, line 292: Line 176 of this method already checks whether the session exists. If it does, you are redirected there. If I understand the code correctly, this part is never executed.

server/src/api/boards.go
mateo-ivc: I don't think that the usage of closure is the right choice here. 

Placing repetitive code inside a anonymous function doesn't help in terms of readability, especially if the code of joinBoard is already cluttered enough. 

I would definitely follow the suggestion of Florian (fine if its in a new PR)

jmoehler: I now started fixing that here:
https://github.com/inovex/scrumlr.io/tree/refactor/redirects

Also @SerseusWasTaken adjusted that as well. He will now take over and merge our two approaches.

server/src/api/boards.go
jmoehler: all redirects will be adjusted with the above mentioned changes

jmoehler:
@Planlos5000 or @mateo-ivc could we please merge this as soon as possible? @SerseusWasTaken is already working on some of the Sonarqube issues that this mr fixes. I want to avoid that we do the work twice. Thanks :)

server/src/api/boards.go
Planlos5000: If the location header is changed with an MR from @SerseusWasTaken the declared function in the function should be removed.

jmoehler: I reverted that now.

server/src/api/boards.go
Planlos5000: This is not about the redirects. Line 176 checks if a session for the board and user exists. In line 280 the check is performed again although we already know that the session does not exist

jmoehler: that may be the case but as I said again and again, this is not the topic of that merge request!

Planlos5000: Then we should open a new issue/merge request for this

jmoehler: will be handled in a separate issue: https://github.com/inovex/scrumlr.io/issues/6235

Planlos5000 (approved):

PR #6072, opened 2026-05-07
build: node 26

### LINKED ISSUE #6036: Upgrade to node 26
> also we need to make sure all actions are using node 24 instead of the current node 20, which has reached EOL, as those will no longer work from June 2nd: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/ 

 _Originally posted by @Schwehn42 in [#6005](https://github.com/inovex/scrumlr.io/issues/6005#issuecomment-4335975377)_

### Node 26
Instead of Node 24, we'll make the jump to Node 26, releasing soon: https://github.com/nodejs/node/pull/62526#issuecomment-4337667357

It will mark the latest LTS version of Node.

one ramification is the fact `corepack`, which is required for yarn 2+, doesn't ship with node anymore.
Thus, we'll have to manually install it in order to yarn berry to work (see https://stackoverflow.com/a/79928399)

### AC
- upgrade `Dockerfile` node
- manually install `corepack` where needed
- upgrade node usages (implicit and explicit) of GitHub runner actions

Schwehn42:
## Description

resolves https://github.com/inovex/scrumlr.io/issues/6036

## Changelog

- Dockerfile: upgrade image to node 26. manually install and activate Corepack.
- CI/CD: for both frontend and cypress jobs, use node actions node 26 and manually install and activate Corepack.
- test: newer node versions require a mocked lokalstorage. for this we use `node-localstorage` and add it to `setupTests.ts`. the emitted file is git ignored.

## Checklist


.github/workflows/continuous-integration.yml
```diff
@@ -31,16 +31,20 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Enable Corepack # to enable yarn berry
-        run: corepack enable
+      - name: Install Corepack # for yarn berry
+        run: |
+          npm install -g corepack@0.34.7 --force --ignore-scripts
```
Planlos5000: Why do we need the `--force` flag? Because of the `--ignore-scripts`?

Schwehn42: no, I thought it can't hurt in case of conflicts, but I guess on a a fresh runner we can assume that won't happen. I've now removed it.

Planlos5000 (approved):

PR #6127, opened 2026-05-18
fix: keep banned participants visible after reload, auto-rejoin on unban

### LINKED ISSUE #4738: banned participants cant be unbanned after session reload
Banned participants can only be unbanned while the owner/moderator is in the current session. Upon reloading, the banned participants are not listed in the participants (not even grayed out!) which makes it impossible to unban them.

Depending on implementation, banned users should still be part in the participants list, but of course still be unable to do anything.

Phai2o:
Closes #4738.

- Participants.tsx: include banned participants in the list regardless of the connected/online filter, so a moderator can still reach the unban button after reloading the page (banned sessions are persisted as connected=false, which previously caused them to disappear).
- RejectionPage.tsx: when a user lands here with status="banned" and a boardId in the URL, silently probe the join endpoint every 30 seconds and surface a "Try to rejoin" button. On a successful re-admission the page dispatches permittedBoardAccess and the board mounts without a manual refresh. Probing goes through API.joinBoard directly instead of the joinBoard thunk so a still-banned response does not flip the redux status to pending and unmount the rejection page.
- i18n (en/de/fr): added bannedRetryHint, retry, retrying.
- styles: button group + retry button styling, dark mode respected.

## Checklist

### DIFF
--- a/src/components/RejectionPage/RejectionPage.scss
+++ b/src/components/RejectionPage/RejectionPage.scss
@@ -82,6 +82,30 @@
   box-shadow: 0 5px 4px color.change(styles.$pink--500, $alpha: 0.25);
 
   cursor: pointer;
+
+  &:disabled {
+    opacity: 0.6;
+    cursor: progress;
+  }
+}
+
+.rejection-page__return-button--retry {
+  background: styles.$blue--500;
+  box-shadow: 0 5px 4px color.change(styles.$blue--500, $alpha: 0.25);
+}
+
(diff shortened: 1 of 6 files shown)


Schwehn42 (commented):
hey @Phai2o, thanks for the PR! on the first look, this looks pretty great. please make sure to update the snapshots by running
```bash
yarn run test --watch=false --update
```

so that the pipeline is happy:)

PR #6130, opened 2026-05-18
fix: handle missing users during board import

jmoehler:
## Description
Board import was refactored to run through the board service and now returns a structured import result with optional warnings.

During note import, author IDs are validated against existing users. Notes with missing/deleted authors are removed. The import logic also repairs stack references and ranks after filtering so imported stacks remain consistent.

## Changelog
- Moved import orchestration into `BoardService` via Import and switched the API endpoint to use it.
- Added import response warnings (`importWarnings.removedNotesMissingAuthorCount`) for skipped notes.
- Added validation of imported note authors via `GetExistingUserIDs` in user `database/service/api` interfaces.
- Renamed `GetUsers` to `GetUsersByBoardID` for clearer naming.
- Added stack repair and rank normalization after filtering missing-author notes.
- Included column description in imported column creation and added column mapping/count validation.
- Added and updated tests for import success/failure, warning behavior, user ID lookup, stack repair, and related API/service behavior; regenerated mocks.

## Checklist


server/src/api/boards.go
```diff
@@ -567,73 +591,178 @@ func (s *Server) importBoard(w http.ResponseWriter, r *http.Request) {
 		Columns:      importColumns,
 		Owner:        owner,
 	})
+	return b, err
+}
```
mating02: nicely split up the big monolithic importBoard function to distinct parts

server/src/api/boards.go
```diff
@@ -567,73 +591,178 @@ func (s *Server) importBoard(w http.ResponseWriter, r *http.Request) {
...
+		if err := s.users.Delete(ctx, userID); err != nil {
+			log.Errorw("failed to cleanup imported placeholder user", "user", userID, "err", err)
+		}
+	}
+}
+
+func (s *Server) replaceDeletedParticipants(ctx context.Context, body boards.ImportBoardRequest) (boards.ImportBoardRequest, []uuid.UUID, error) {
+	if len(body.Notes) == 0 {
+		return body, nil, nil
+	}
```
mating02: good: checked for edge case empty board

server/src/api/boards.go
```diff
@@ -567,73 +591,178 @@ func (s *Server) importBoard(w http.ResponseWriter, r *http.Request) {
...
+
+	participants := uniqueImportedNoteAuthors(body.Notes)
+
+	existingParticipants, err := s.users.GetExistingUserIDs(ctx, participants)
 	if err != nil {
-		span.SetStatus(codes.Error, "failed to get columns from imported board")
-		span.RecordError(err)
-		_ = s.boards.Delete(ctx, b.ID)
+		return body, nil, err
```
mating02: since this is called in the api layer and an error here would mean a database request error, it would make sense to use common.InternalServerError instead of err OR I could also see the functions you added be transferred over to the service layer, since none of these functions take a http request or responsewriter as an argument

server/src/api/boards.go
```diff
@@ -567,73 +591,178 @@ func (s *Server) importBoard(w http.ResponseWriter, r *http.Request) {
...
-				parentNote = *note
-			}
+		newColID, exists := colMap[parentNote.Position.Column]
+		if !exists {
+			continue
 		}
-		organizedNotes = append(organizedNotes, ParentChildNotes{
-			Parent:   parentNote,
+
+		note, err := s.notes.Import(ctx, notes.NoteImportRequest{
```
mating02: little suggestion: importing inside the loop may degrade data validity -> for example if there are connectivity issues to the database mid loop then only half the notes would get imported -> see https://go.dev/doc/database/execute-transactions for reference

server/src/api/boards_test.go
```diff
@@ -452,3 +456,274 @@ func (suite *BoardTestSuite) TestIncrementTimer() {
...
+
+	columnsMock.EXPECT().GetAll(mock.Anything, boardID).Return([]*columns.Column{{ID: createdColumnID}}, nil)
+	usersMock.EXPECT().GetExistingUserIDs(mock.Anything, []uuid.UUID{deletedAuthorID}).Return([]uuid.UUID{}, nil)
+	usersMock.EXPECT().CreateAnonymous(mock.Anything, "deleted user "+deletedAuthorID.String()[:5]).Return(&users.User{ID: replacementAuthorID}, nil)
+	notesMock.EXPECT().Import(mock.Anything, mock.Anything).Return(nil, errors.New("import failed"))
+	usersMock.EXPECT().Delete(mock.Anything, replacementAuthorID).Return(nil)
+
+	err := s.processImportedNotes(ctx, boardID, body)
+
+	suite.Error(err)
```
mating02: missing assertexpectations for validity

server/src/api/boards_test.go
```diff
@@ -452,3 +456,274 @@ func (suite *BoardTestSuite) TestIncrementTimer() {
...
+	}
+
+	usersMock.EXPECT().GetExistingUserIDs(mock.Anything, []uuid.UUID{firstDeletedAuthorID, secondDeletedAuthorID}).Return([]uuid.UUID{}, nil)
+	usersMock.EXPECT().CreateAnonymous(mock.Anything, "deleted user "+firstDeletedAuthorID.String()[:5]).Return(&users.User{ID: firstReplacementAuthorID}, nil)
+	usersMock.EXPECT().CreateAnonymous(mock.Anything, "deleted user "+secondDeletedAuthorID.String()[:5]).Return(nil, errors.New("create failed"))
+	usersMock.EXPECT().Delete(mock.Anything, firstReplacementAuthorID).Return(nil)
+
+	_, _, err := s.replaceDeletedParticipants(ctx, body)
+
+	suite.Error(err)
```
mating02: missing assertexpectations

server/src/api/boards_test.go
```diff
@@ -452,3 +456,274 @@ func (suite *BoardTestSuite) TestIncrementTimer() {
...
+	columnIndex      int
+	columnColor      common.Color
+
+	ownerID         uuid.UUID
+	importColumnID  uuid.UUID
+	createdBoardID  uuid.UUID
+	createdColumnID uuid.UUID
+	authorID        uuid.UUID
+	importedNoteID  uuid.UUID
+}
```
mating02: i would suggest to put this struct on top so it can also be used by other functions like TestCreateBoard (for this purpose it could simply be renamed to boardTestFixture)

server/src/api/boards_test.go
```diff
@@ -452,3 +456,274 @@ func (suite *BoardTestSuite) TestIncrementTimer() {
...
+	boardMock.EXPECT().Delete(mock.Anything, fixture.createdBoardID).Return(nil)
+
+	rr := httptest.NewRecorder()
+
+	s.importBoard(rr, req)
+
+	suite.Equal(http.StatusInternalServerError, rr.Result().StatusCode)
+	boardMock.AssertExpectations(suite.T())
+	columnMock.AssertExpectations(suite.T())
+}
```
mating02, line 740: Overall well done! The tests are a good addition and provide good coverage of different edge cases!

server/src/users/database_test.go
```diff
@@ -90,7 +90,7 @@ func (suite *DatabaseUserTestSuite) TesDatabaseCreatAzureAdUser() {
 	assert.Nil(t, dbUser.Avatar)
 }
 
-func (suite *DatabaseUserTestSuite) TesDatabaseCreatGitHubUser() {
+func (suite *DatabaseUserTestSuite) TestDatabaseCreatGitHubUser() {
```
mating02: you corrected the spelling mistake (tes -> test) which is good but you left creat uncorrected

server/src/users/database_test.go
```diff
@@ -295,11 +295,84 @@ func (suite *DatabaseUserTestSuite) TestDatabaseGetBoardUsersEmpty() {
...
+	dbUsers, err := database.GetExistingUserIDs(context.Background(), userIDs)
+
+	assert.NoError(t, err)
+
+	expected := []uuid.UUID{
+		suite.users["Stan"].ID,
+		suite.users["Friend"].ID,
+	}
+	assert.ElementsMatch(t, expected, dbUsers)
+}
```
mating02, line 375: nice coverage for the tests! well done

server/src/users/database_test.go
```diff
@@ -362,7 +435,7 @@ func (suite *DatabaseUserTestSuite) TesDatabasIsAvailableForKeyMigratioFalse() {
 	assert.False(t, isAvailable)
 }
 
-func (suite *DatabaseUserTestSuite) TesDatabasSetKeyMigration() {
+func (suite *DatabaseUserTestSuite) TestDatabasSetKeyMigration() {
```
mating02: same issue as one before here, spelling mistake was corrected for test but not for database

server/src/users/service_test.go
```diff
@@ -77,10 +77,38 @@ func (suite *UserServiceTestSuite) TestGetUser_DatabaseError() {
...
+	mockNotesService := notes.NewMockNotesService(suite.T())
+	userService := NewUserService(suite.mockUserDatabase, suite.broker, mockSessionService, mockNotesService)
+
+	user, err := userService.GetExistingUserIDs(context.Background(), userIDs)
+
+	suite.Nil(user)
+	suite.NotNil(err)
+	suite.Equal(common.InternalServerError, err)
+}
+
```
mating02, line 107: if not existent in a hook anywhere else, it would probably be a good idea to include assertexpectations() here

server/src/users/database_test.go
jmoehler: fixed that

server/src/users/database_test.go
jmoehler: fixed that

server/src/api/boards_test.go
jmoehler: true I fixed that

server/src/api/boards_test.go
jmoehler: true, I fixed that

server/src/api/boards_test.go
jmoehler: I moved it to the top, but I am not sure if we will ever need to use  this in the testCreateBoard, as that is only used in the import board tests

server/src/users/service_test.go
jmoehler: how do you mean? I am already comparing the thrown error type to the expected type.

mating02: yeah my bad, assertexpectations is not necessary here

jmoehler:
apart from the change requests, I need to adjust according to the changed requirements by Selina. So for now I set the mr to be a draft

server/src/boards/service.go
```diff
@@ -683,3 +688,333 @@ func (service *Service) BoardEditableContext(next http.Handler) http.Handler {
...
+	}
+
+	if len(newColumns) != len(request.Columns) {
+		return nil, nil, fmt.Errorf("column count mismatch during import mapping: imported=%d created=%d", len(request.Columns), len(newColumns))
+	}
+
+	columnMap := make(map[uuid.UUID]uuid.UUID, len(request.Columns))
+	for i, sourceColumn := range request.Columns {
+		columnMap[sourceColumn.ID] = newColumns[i].ID
+	}
```
Planlos5000: The columns from the `GetAll` call are sorted with the index. Can we be sure, that the columns from the request are also sorted by the index?

jmoehler: This wouldn't be a problem, because the index was overwritten for all columns at import. however that also meant that importing columns in orders like 0 2 1 would switch the columns. I now adjusted the column import as well as the old <-> new column connection to be more stable.

server/src/api/boards.go
mating02, line 1: good job refactoring this since the function was too overloaded

server/src/boards/service.go
```diff
@@ -180,37 +217,69 @@ func (service *Service) Create(ctx context.Context, body CreateBoardRequest) (*B
...
-		_, err = service.columnService.Create(ctx, column)
+		createdColumn, err := service.columnService.Create(ctx, column)
 		if err != nil {
-			span.SetStatus(codes.Error, "failed to create column")
-			span.RecordError(err)
 			return nil, err
 		}
+
+		if value.SourceColumnID == nil {
+			continue
```
mating02, line 250: Why continue here? If every column needs a source ID, then we can throw an error here immediately.

server/src/boards/service_test.go
```diff
@@ -20,742 +22,1710 @@ import (
...
-	service := NewBoardService(mockBoardDatabase, broker, sessionRequestMock, sessionsMock, columnMock, noteMock, reactionMock, votingMock, mockClock, mockHash)
-	result, err := service.Get(context.Background(), boardID)
-
-	require.Error(t, err)
-	assert.Nil(t, result)
-	assert.Equal(t, dbError, err)
+	suite.Error(err)
+	suite.Nil(result)
+	suite.Equal(dbError, err)
 }
```
mating02, line 111: very clean and helpful refactoring!

server/src/boards/service_integration_test.go
mating02, line 1: Despite already being rigorously tested in the service_test.go file, the new Import, ImportCreatedBoard functions should definitely be featured in the service_integration_test.go as well to ensure full functionality in final integration

mating02 (changes requested):
service integration test should also test the new import function to ensure full functionality, , merge conflicts need to be resolved. Other than that, everything looks fine. Functionalities have successfully been tested in local development.

server/src/boards/service.go
jmoehler: This is intentional. The helper is used by both normal board creation (no source column IDs) and import creation (source column IDs present). When source IDs are present, it validates import mapping integrity by rejecting duplicate source IDs and incomplete mappings.

server/src/boards/service_test.go
jmoehler: thanks :)

server/src/boards/service_integration_test.go
jmoehler: added tests now

server/src/boards/service_integration_test.go
```diff
@@ -663,6 +667,280 @@ func (suite *BoardServiceIntegrationTestSuite) Test_IncrementTimer() {
...
+	assert.Equal(t, rootColumnID, rootNote.Position.Column)
+	assert.False(t, rootNote.Position.Stack.Valid)
+	assert.Equal(t, 0, rootNote.Position.Rank)
+
+	childNote, found := notesByText["Child note"]
+	require.True(t, found)
+	assert.Equal(t, rootColumnID, childNote.Position.Column)
+	assert.True(t, childNote.Position.Stack.Valid)
+	assert.Equal(t, rootNote.ID, childNote.Position.Stack.UUID)
+	assert.Equal(t, 0, childNote.Position.Rank)
```
mating02, line 870: why should the rank be equal to root rank? Isn't the rank supposed to order the importance of notes chronologically?

jmoehler: The ranks are a bit weird. They are ordered from 0 up to the highest one that exists. And within a stack, it starts again at 0 and goes up sorted. The higher the number the higher its position.

In this case:
The root note is the first root in its column.
The child note is the first child in its stack.

mating02 (approved):
all functionalities successfully tested and all conflicts resolved

PR #6141, opened 2026-05-18
feat: history cards design implementation 

Schwehn42:
> [!WARNING]  
> Before merging, make sure to set `.env` feature flag to false to avoid prod deployment (DONE)

## Description

This PR implements the history page and cards as seen in the design specs.

No actual data is pulled from the backend; instead, test data is used. 

The more important part is that the data is displayed and formatted correctly.

## Changelog

- `.env`: add `VITE_SHOW_HISTORY_PAGE` to toggle between teaser and cards 
- `History`
  - renamed from `Sessions`
  - add search filtering
  - auto wrapping of cards in different screen sizes
- `HistoryCards`
  - add as per design, including:
    - board title and description
    - access policy icon
    - user role chip
    - favorite state
    - locked/read-only indicator
    - column, participant, note, created, and updated metadata
    - action menu for delete, copy link, create template, edit, and close  
  - localized and formatted texts and date times 

## Checklist


.env
jmoehler, line 1: why do we have envs in git? shouldn't they be excluded?

.env.development
jmoehler, line 1: why do we have envs in git? shouldn't they be excluded?

src/utils/datetime.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+/**
+ * compares two dates and returns duration components alongside calendar-specific context.
+ */
+export const getTimeDifference = (date1: Date, date2: Date): DurationComponents => {
+  // calculate the raw difference and take absolute value just in case they're flipped
+  const diffMs = Math.abs(date1.getTime() - date2.getTime());
+
+  return getDurationComponents(new Date(diffMs));
```
jmoehler: modeling a time difference as Date looks a bit odd to me, why not use a number?

src/utils/datetime.ts
```diff
@@ -0,0 +1,42 @@
+export interface DurationComponents {
+  milliseconds: number;
+  seconds: number;
+  minutes: number;
+  hours: number;
+  days: number;
+}
+
+export const getDurationComponents = (date: Date): DurationComponents => {
```
jmoehler: this doesn't really return the duration components but rather its totals. I suggest renaming it to `getDurationTotals` or similar

jmoehler:
image


in responsive mode, once can currently scroll though the description. in figma it shows three dots.

also the overview looks very very full and crowded in responsive mode. could we perhaps do everything a bit smaller or less cluttered? e.g. showing less information?

Also there is a lot of space left and right of the cards that is not used

src/components/Templates/TemplateCard/TemplateCard.tsx
```diff
@@ -57,16 +57,14 @@ export const TemplateCard = (props: TemplateCardProps) => {
           {label: "Edit", element: <EditIcon />, onClick: () => props.onNavigateToEdit(template.id)},
           {label: "Close", element: <CloseIcon />, onClick: closeMenu},
         ]}
-        focusBehaviour="moveFocus"
+        focusBehaviour="trap"
```
jmoehler, line 60: that doesn't seem to have any effect, at least I couldn't find any difference.

src/components/Templates/TemplateCard/TemplateCard.scss
```diff
@@ -6,25 +6,28 @@ $card-description-height: 4 * 20px;
...
     ".          .  description   description   ."
     ".          .  .             .             ."
-    "col-icon   .  columns       start         start";
+    "col-icon   .  columns       button        button";
 
   width: 100%;
   height: $card-height;
   border-radius: styles.$rounded--default;
-  padding: styles.$spacing--base;
+  padding: styles.$spacing--md;
```
jmoehler, line 20: this reduces usable space, I would suggest we go back to base at least in mobile. or is there any reason for md?

src/components/Templates/TemplateCard/TemplateCard.scss
```diff
@@ -61,9 +64,9 @@ $card-description-height: 4 * 20px;
 
   width: 100%;
 
-  font-size: styles.$text--md;
+  font-size: styles.$text--lg;
```
jmoehler, line 67: this reduces usable space, I would suggest we go back to md at least in mobile. Or is there any reason for lg?

src/routes/Boards/Sessions/HistoryCard/HistoryCard.tsx
```diff
@@ -0,0 +1,238 @@
...
+import {useTranslation} from "react-i18next";
+import {useAppSelector} from "store";
+import {getTimeDifference, isDateYesterday} from "utils/datetime";
+import "./HistoryCard.scss";
+
+type HistoryCardProps = {
+  board: HistoryBoard;
+};
+
+export const HistoryCard = (props: HistoryCardProps) => {
```
jmoehler, line 38: the differentiation on class name level e.g. `history-card__icon--public` is not used by any css. pharaohs that can be tightened by giving them all the same className?

jmoehler: same with columns, calendar, participans, notes in 159, 169, 181, and 188

src/i18n/fr/translation.json
```diff
@@ -94,11 +94,31 @@
       }
     }
   },
-  "Sessions": {
+  "History": {
     "switchTitle": "Histoire",
-    "title": "Vos Boards",
-    "savedBoards": "Vos sessions",
-    "teaserText": "Les utilisateurs inscrits pourront prochainement consulter et gérer l'ensemble de leurs propres tableaux ici. Les tableaux créés en mode invité ne peuvent être ni rattachés à un compte, ni sauvegardés."
+    "title": "Gérer tes tableux",
```
jmoehler: should be tableaux :)

jmoehler (dismissed):
some smaller change requests but looks good overall.

One general Point: The design looks very dense (especially on mobile version). Maybe we can make it a bit less dense by making stuff smaller and reducing margins?

src/i18n/fr/translation.json
Schwehn42: fixed

src/routes/Boards/Sessions/HistoryCard/HistoryCard.tsx
Schwehn42: its just something I like to do clarify the difference of similar elements, especially icons, in case we want to apply styling to a specific element

src/components/Templates/TemplateCard/TemplateCard.scss
Schwehn42: its for design parity between history and template cards, so they don't look off when switching

src/components/Templates/TemplateCard/TemplateCard.scss
Schwehn42: its for design parity between history and template cards, so they don't look off when switching.

space wise, it shouldn't be too much of a loss

src/components/Templates/TemplateCard/TemplateCard.tsx
Schwehn42: the only difference is how tabbing is handled; `trap` keeps the focus rotating inside the mini menu while `moveFocus` like the name implies only initially moves focus, but tabbing through will go outside again

src/utils/datetime.ts
Schwehn42: I can agree with that, should be clarified now with more fitting names and comments 👍

src/utils/datetime.ts
Schwehn42: I thought a date is more versatile, but agreed and changed to numeric milliseconds value

.env
Schwehn42: I agree it looks weird, and we certainly must never put any sensitive information there

we use our .env for frontend feature flags, which is handy in this case because we want to enable the feature in our PR deployment pipeline, but not have it active in prod

as precaution, I added a comment to the env and also a TODO in the PR description to remove it before merging

.env.development
tohega: These aren't secrets, we use it for the Vite build time config / feature flags. There is nothing sensitive to hide in our case. In this PR we use it specifically for the `VITE_SHOW_HISTORY_PAGE` feature flag

tohega (approved):

PR #6142, opened 2026-05-18
fix: sync language state with redux and i18n integration

Schwehn42:
## Description

previously, the redux state only got initialized, but not updated on language changes. this PR actually dispatches language changes from the settings to the store, which in turn updates the i18n config, keeping them in sync

## Changelog
- `LanguageSettingsDropdown`: instead of setting i18n itself, dispatch to store
- store: `setLanguage` action sets i18n language

## Checklist


src/store/features/view/reducer.ts
```diff
@@ -50,6 +51,9 @@ export const viewReducer = createReducer(initialState, (builder) =>
       state.moderating = action.payload;
     })
     .addCase(setLanguage, (state, action) => {
+      i18n.changeLanguage(action.payload).then(() => {
```
jmoehler: This could create race conditions and side effects, since it makes the reducer asynchronous. 
Perhaps move the async behavior in a thunk/listener.

jmoehler (changes requested):
two more thinks:

- Homepage.tsx line 27-31 still uses a direct language update bypassing redux
- We have the reducer.test.ts, adding tests for the language change would nice.

src/store/features/view/reducer.ts
Schwehn42: that's a very good point, side effects should always be done within a thunk 👍 changed this

Schwehn42:
@jmoehler thanks for the review 👍 

concerning the tests, we generally don't really test redux itself as its pretty tedious and out of the scope of a unit test. the tests you're referring to are to make sure that the server parameters when initially opening scrumlr are set correctly, which I guess is a rather specific use case.

jmoehler (approved):
lgtm!

PR #6155, opened 2026-05-20
ci: Add dependency groups for frontend dependencies

Planlos5000:
## Description

Add dependency groups in the dependabot configuration for the frontend.
This reduces the merge requests from dependabot and helps that the merge requests are valid.
At the moment it could happen that some merge requests from dependabot don't make it through the pipeline because it also needs other dependencies to be updated.
## Changelog

- Add groups to the dependabot configuration for the frontend
## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,6 +18,14 @@ updates:
     commit-message:
       prefix: "yarn"
       include: "scope"
+    groups:
+      typescript-eslint:
+        patterns:
+          - "@typescript-eslint/*"
+      react:
+        patterns:
+          - "react"
+          - "react-dom"
 
   # Server: src module
   - package-ecosystem: "gomod"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -18,6 +18,14 @@ updates:
     commit-message:
       prefix: "yarn"
       include: "scope"
+    groups:
+      typescript-eslint:
+        patterns:
+          - "@typescript-eslint/*"
```
Schwehn42, line 24: could also include eslint* once we have the upgrade refactor going

.github/dependabot.yml
```diff
@@ -18,6 +18,14 @@ updates:
...
+      react:
+        patterns:
+          - "react"
+          - "react-dom"
```
Schwehn42, line 29: could also include `react-router` (maintained by another team but thematically fitting), and their respective`@types/`. though we have a bunch of deps starting with `react`, so the question is how tight the grouping should be

Schwehn42 (approved):
lgtm

PR #6157, opened 2026-05-20
refactor errors, seperating service domain errors from API layer errors with status codes

### LINKED ISSUE #5812: Refactor errors
We should check the returned errors for the api and the service and should discuss the errors returned and where we set them.

e.g. Currently we return an error is we ask for all notes of a board. It could also return an ok respones with an empty list

mating02:
closes #5812 

## Description
This PR refactors the error handling and service structure to enforce a stricter separation of concerns. Previously, our service layer was "leaky," as it directly handled HTTP-specific errors (`common.APIError`).

I have moved all HTTP-related logic into the api package. Services now return pure domain errors or wrapped infrastructure errors. This makes the service layer transport-agnostic and much easier to unit test.

## Changelog
- Domain Error Mapping: Introduced domain errors (e.g., `ErrBoardNotFound`, `ErrNameEmpty`) in a seperate `errors.go` file for the `boards`, `boardtemplates`, `columns`, `columntemplates`, `notes`, `reactions`, `sessionrequests`, `sessions`, `users`, and `votings` packages in `server/src/`, adapted the according `service.go` files and implemented a centralized `mapError` helper in the api package (`server/src/api/`) to translate these into HTTP status codes.
- Editing `service_test` and `service_integration` test files in the packages where it was necessary, so they do not assert the throw of a `common.APIError` error type at specific places in the code
- replacing `assert.Equal` and `suite.Equal` comparisons for domain errors in the respective `service_test.go` and `service_integration_test.go` files

## Checklist

## (Optional) Visual Changes
No visual changes, since it is a pure backend issue

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ deployment/**/.env
 
 # script output
 scripts/out
+
+#coverage files
+*coverage.txt
--- a/.gitignore
+++ b/.gitignore
@@ -46,5 +46,3 @@ deployment/**/.env
 # script output
 scripts/out
 
-#coverage files
-*coverage.txt
--- a/server/src/api/boards.go
+++ b/server/src/api/boards.go
@@ -136,7 +136,7 @@ func (s *Server) getBoards(w http.ResponseWriter, r *http.Request) {
 		span.SetStatus(codes.Error, "failed to get boards")
 		span.RecordError(err)
 		log.Errorw("failed to get boards", "err", err)
-		common.Throw(w, r, common.InternalServerError)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
@@ -145,7 +145,7 @@ func (s *Server) getBoards(w http.ResponseWriter, r *http.Request) {
 		span.SetStatus(codes.Error, "failed to get board overview")
 		span.RecordError(err)
 		log.Errorw("failed to get board overview", "err", err)
-		common.Throw(w, r, common.InternalServerError)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 	render.Status(r, http.StatusOK)
@@ -236,7 +236,7 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		span.SetStatus(codes.Error, "failed to check session")
 		span.RecordError(err)
-		common.Throw(w, r, common.InternalServerError)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
@@ -245,7 +245,7 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			span.SetStatus(codes.Error, "failed to check if participant is banned")
 			span.RecordError(err)
-			common.Throw(w, r, common.InternalServerError)
+			common.Throw(w, r, mapError(err))
 			return
 		}
 
@@ -265,7 +265,7 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		span.SetStatus(codes.Error, "failed to get board")
 		span.RecordError(err)
-		common.Throw(w, r, common.NotFoundError)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
@@ -274,7 +274,7 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			span.SetStatus(codes.Error, "failed to create session")
 			span.RecordError(err)
-			common.Throw(w, r, common.InternalServerError)
+			common.Throw(w, r, mapError(err))
 			return
 		}
 
@@ -308,7 +308,7 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
 			if err != nil {
 				span.SetStatus(codes.Error, "failed to create session")
 				span.RecordError(err)
-				common.Throw(w, r, common.InternalServerError)
+				common.Throw(w, r, mapError(err))
 				return
 			}
 
@@ -394,7 +394,7 @@ func (s *Server) updateBoard(w http.ResponseWriter, r *http.Request) {
 		span.SetStatus(codes.Error, "failed to update board")
 		span.RecordError(err)
 		log.Errorw("Unable to update board", "err", err)
-		common.Throw(w, r, err)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
@@ -439,7 +439,7 @@ func (s *Server) setTimer(w http.ResponseWriter, r *http.Request) {
 		span.SetStatus(codes.Error, "failed to set board timer")
 		span.RecordError(err)
 		log.Errorw("Unable to set board timer", "err", err)
-		common.Throw(w, r, err)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
@@ -474,7 +474,7 @@ func (s *Server) deleteTimer(w http.ResponseWriter, r *http.Request) {
 		span.SetStatus(codes.Error, "failed to delete board timer")
 		span.RecordError(err)
 		log.Errorw("Unable to delete board timer", "err", err)
-		common.Throw(w, r, err)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
@@ -602,7 +602,7 @@ func (s *Server) exportBoard(w http.ResponseWriter, r *http.Request) {
 					if err != nil {
 						span.SetStatus(codes.Error, "failed to get note author user")
 						span.RecordError(err)
-						common.Throw(w, r, err)
+						common.Throw(w, r, mapError(err))
 						return
 					}
 					author = user.Name
--- a/server/src/api/boards.go
+++ b/server/src/api/boards.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"database/sql"
 	"encoding/csv"
 	"errors"
 	"fmt"
@@ -69,7 +68,7 @@ func (s *Server) createBoard(w http.ResponseWriter, r *http.Request) {
 		span.SetStatus(codes.Error, "failed to create board")
 		span.RecordError(err)
 		log.Errorw("failed to create board", "err", err)
-		common.Throw(w, r, err)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
@@ -182,17 +181,10 @@ func (s *Server) getBoard(w http.ResponseWriter, r *http.Request) {
 
 	board, err := s.boards.Get(ctx, boardId)
 	if err != nil {
-		if err == sql.ErrNoRows {
-			span.SetStatus(codes.Error, "no board found")
-			span.RecordError(err)
-			common.Throw(w, r, common.NotFoundError)
-			return
-		}
-
 		span.SetStatus(codes.Error, "failed to get board")
 		span.RecordError(err)
 		log.Errorw("unable to access board", "err", err)
-		common.Throw(w, r, common.InternalServerError)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
@@ -512,7 +504,7 @@ func (s *Server) incrementTimer(w http.ResponseWriter, r *http.Request) {
 		span.SetStatus(codes.Error, "failed to increment board timer")
 		span.RecordError(err)
 		log.Errorw("Unable to increment board timer", "err", err)
-		common.Throw(w, r, err)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
@@ -547,7 +539,7 @@ func (s *Server) exportBoard(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		span.SetStatus(codes.Error, "failed to get full board")
 		span.RecordError(err)
-		common.Throw(w, r, err)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
@@ -696,7 +688,7 @@ func (s *Server) importBoard(w http.ResponseWriter, r *http.Request) {
 		span.SetStatus(codes.Error, "failed to import board")
 		span.RecordError(err)
 		log.Errorw("Could not import board", "err", err)
-		common.Throw(w, r, err)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
--- a/server/src/api/boards.go
+++ b/server/src/api/boards.go
@@ -182,7 +182,11 @@ func (s *Server) getBoard(w http.ResponseWriter, r *http.Request) {
 	board, err := s.boards.Get(ctx, boardId)
 	if err != nil {
 		mappedErr := mapError(err)
-		span.SetStatus(codes.Error, mappedErr.Error())
+		if errors.Is(mappedErr, common.NotFoundError) {
+			span.SetStatus(codes.Error, "board not found")
+		} else {
+			span.SetStatus(codes.Error, "failed to get board")
+		}
 		span.RecordError(err)
 		log.Errorw("unable to access board", "err", err)
 		common.Throw(w, r, mappedErr)
--- a/server/src/api/boards.go
+++ b/server/src/api/boards.go
@@ -181,10 +181,11 @@ func (s *Server) getBoard(w http.ResponseWriter, r *http.Request) {
 
 	board, err := s.boards.Get(ctx, boardId)
 	if err != nil {
-		span.SetStatus(codes.Error, "failed to get board")
+		mappedErr := mapError(err)
+		span.SetStatus(codes.Error, mappedErr.Error())
 		span.RecordError(err)
 		log.Errorw("unable to access board", "err", err)
-		common.Throw(w, r, mapError(err))
+		common.Throw(w, r, mappedErr)
 		return
 	}
 
--- a/server/src/api/errors.go
+++ b/server/src/api/errors.go
@@ -1,0 +1,37 @@
+package api
+
+import (
+	"database/sql"
+	"errors"
+
+	"scrumlr.io/server/boards"
+	"scrumlr.io/server/common"
+)
+
+// mapError translates domain and database errors to HTTP API Errors.
+func mapError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	// Check for specific domain errors
+	switch {
+	case errors.Is(err, boards.ErrPassphraseForbidden),
+		errors.Is(err, boards.ErrPassphraseRequired),
+		errors.Is(err, boards.ErrNameEmpty),
+		errors.Is(err, boards.ErrInvalidPassphrase):
+		return common.BadRequestError(err)
+
+	case errors.Is(err, boards.ErrNotAuthorized):
+		return common.ForbiddenError(err)
+
+	case errors.Is(err, boards.ErrBoardNotFound),
+		errors.Is(err, sql.ErrNoRows): // Catch database "not found"
+		return common.NotFoundError
+
+	default:
+		// If it's an unmapped error (like a DB connection failure),
+		// return a generic 500 without leaking DB internals to the client.
+		return common.InternalServerError
+	}
+}
--- a/server/src/api/errors.go
+++ b/server/src/api/errors.go
@@ -4,15 +4,7 @@ import (
 	"database/sql"
 	"errors"
 
-	"scrumlr.io/server/boards"
-	"scrumlr.io/server/columns"
 	"scrumlr.io/server/common"
-	"scrumlr.io/server/notes"
-	"scrumlr.io/server/reactions"
-	"scrumlr.io/server/sessionrequests"
-	"scrumlr.io/server/sessions"
-	"scrumlr.io/server/users"
-	"scrumlr.io/server/votings"
 )
 
 // mapError translates domain and database errors to HTTP API Errors.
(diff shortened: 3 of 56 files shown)


Planlos5000:
Why not have a specific error for each package like `NoteError` that implements the `error` interface (https://go.dev/blog/error-handling-and-go#the-error-type)?

server/src/notes/errors.go
```diff
@@ -0,0 +1,47 @@
...
+import "fmt"
+
+// NoteErrorCategory allows us to classify errors for easier handling at the API layer
+type NoteErrorCategory string
+
+const (
+	CategoryBadRequest NoteErrorCategory = "BAD_REQUEST"
+	CategoryConflict   NoteErrorCategory = "CONFLICT"
+	CategoryForbidden  NoteErrorCategory = "FORBIDDEN"
+	CategoryNotFound   NoteErrorCategory = "NOT_FOUND"
```
Planlos5000: For the other packages the categories have e.g. just the name `BadRequest` or `NotFound`. Why do these have here the prefix `Category`?

Maybe also add a `InternalServer` category. This would help to get rid of the `fmt.Errorf` code

server/src/boards/errors.go
```diff
@@ -0,0 +1,37 @@
...
+	ErrPassphraseForbidden  = BoardError{Category: BadRequest, Message: "passphrase should not be set for policies except 'BY_PASSPHRASE'"}
+	ErrPassphraseRequired   = BoardError{Category: BadRequest, Message: "passphrase must be set on access policy 'BY_PASSPHRASE'"}
+	ErrInvalidBoardSettings = BoardError{Category: BadRequest, Message: "invalid board settings"}
+	ErrInvalidPassphrase    = BoardError{Category: BadRequest, Message: "wrong passphrase"}
+	ErrNameEmpty            = BoardError{Category: BadRequest, Message: "name cannot be empty"}
+	// Forbidden Errors
+	ErrNotAuthorized = BoardError{Category: Forbidden, Message: "not authorized to change board"}
+	// Not Found Errors
+	ErrBoardNotFound = BoardError{Category: NotFound, Message: "board not found"}
+)
```
Planlos5000: Wouldn't it be more maintainable to have a function that creates e.g. a `BoardError` with the category `BadRequest` and takes a custom message. Then the errors don't have to be defined first.

server/src/notes/errors.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// NoteError is our custom domain error that implements the built-in error interface
+type NoteError struct {
+	Category NoteErrorCategory
+	Message  string
+}
```
Planlos5000, line 21: What do you think of the idea to also include the original error e.g. for logging purposes

server/src/api/errors.go
```diff
@@ -0,0 +1,38 @@
...
+// mapError translates domain and database errors to HTTP API Errors.
+func mapError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	// DB "no rows" error is a common case for "not found"
+	if errors.Is(err, sql.ErrNoRows) {
+		return common.NotFoundError
+	}
```
Planlos5000, line 19: Is this correct with the criteria that a get all endpoint should return a empty list with a status of `200`?

mating02: > Is this correct with the criteria that a get all endpoint should return a empty list with a status of `200`?

Yes , because sql.ErrNoRows is not returned when an empty row / entry is found. It is only returned when NO matching row / entry is found at all  (see https://go.dev/doc/database/querying )

server/src/notes/errors.go
mating02: good observation, will do!

server/src/notes/errors.go
mating02: > What do you think of the idea to also include the original error e.g. for logging purposes

Only makes sense for internal server errors. For the domain errors (like ErrNameEmpty), that does not really make sense since the error is defined by the domain error itself and there is no "original error" in that sense

server/src/boards/errors.go
mating02: I'd rather not do that as the tests would then lose the ability to use errors.Is() for precise testing of specific domain errors. Testing would then have to be then via string message matching and suite.Equal(), which is rather tedious and ill maintainable

server/src/notes/errors.go
```diff
@@ -0,0 +1,57 @@
...
+	NotFound   NoteErrorCategory = "NOT_FOUND"
+	Internal   NoteErrorCategory = "INTERNAL"
+)
+
+type NoteError struct {
+	Category NoteErrorCategory
+	ErrType  NoteErrorType
+	Message  string
+	Err      error
+}
```
Planlos5000, line 21: Do we still need the error type. I would argue that the category and a descriptive message is sufficient and the type does not have an additional value.

mating02: I came up with the idea of the error type because it makes testing much safer and more easily comprehensible. If you leave out the error type and just go with the descriptive message, how would you test for the specific domain error? You would need to match the string and that is not a good practice in testing I would argue

server/src/api/errors.go
```diff
@@ -0,0 +1,38 @@
...
+		case "BAD_REQUEST":
+			return common.BadRequestError(err)
+		case "FORBIDDEN":
+			return common.ForbiddenError(err)
+		case "NOT_FOUND":
+			return common.NotFoundError
+		case "CONFLICT":
+			return common.ConflictError(err)
+		default:
+			return common.InternalServerError
```
Planlos5000: The default case is not needed here because is no case matches we can use the return of the function in line 37

server/src/api/boards.go
```diff
@@ -130,17 +129,10 @@ func (s *Server) getBoard(w http.ResponseWriter, r *http.Request) {
...
-			return
-		}
-
 		span.SetStatus(codes.Error, "failed to get board")
 		span.RecordError(err)
 		log.Errorw("unable to access board", "err", err)
-		common.Throw(w, r, common.InternalServerError)
+		common.Throw(w, r, mapError(err))
 		return
 	}
```
Planlos5000, line 194: Is it possible to set the message of the error as the status of the span. This would allow to differentiate more easily in the spans if the board was not found or another error occurred

server/src/boards/service.go
Planlos5000: Missing the new board error

Planlos5000: Missing new board error

Planlos5000: Missing new board error

Planlos5000: Missing new board error

Planlos5000: Missing new board error

Planlos5000: Missing new board error

Planlos5000: Missing new board error

server/src/boardtemplates/service.go
Planlos5000, line 1: Where are the boar template errors?

server/src/columns/service.go
Planlos5000: Missing new column error

Planlos5000: Missing new column error

server/src/columntemplates/service.go
Planlos5000, line 1: Where are the column template errors?

server/src/common/error.go
```diff
@@ -54,9 +54,11 @@ var NotFoundError = &APIError{StatusCode: http.StatusNotFound, StatusText: "Reso
 var InternalServerError = &APIError{StatusCode: http.StatusInternalServerError, StatusText: "Internal server error."}
 
 func Throw(w http.ResponseWriter, r *http.Request, err error) {
+	//If it is an APIError, it uses render.Render to send that specific error to the user. An APIError contains a specific HTTP status code and a user-friendly message.
 	if apiErr, ok := errors.AsType[*APIError](err); ok {
 		_ = render.Render(w, r, apiErr)
 		return
 	}
+	// Branch B (Unknown Error): If the error is just a standard string error (like a database connection failure or a null pointer), it returns a generic InternalServerError.
 	_ = render.Render(w, r, InternalServerError)
```
Planlos5000: Remove the AI generated comments

server/src/notes/service.go
Planlos5000: Missing new note error

Planlos5000: Missing new note error

Planlos5000: Missing new note error

server/src/sessionrequests/service.go
Planlos5000: Missing new session request error

Planlos5000: Missing new session request error

Planlos5000: We should assign the values from the database call to variables so that we could map the error to the new session request error and return this error then

server/src/sessions/service.go
Planlos5000: Missing new session error

Planlos5000: Missing new session error

Planlos5000: Missing new session error

Planlos5000: Missing new session error

Planlos5000: Missing new session error

Planlos5000: Missing new session error

Planlos5000: Missing new session error

Planlos5000: Assign the call to variables so that the error could be mapped to the new session error

Planlos5000: Assign the call to variables so that the error can be mapped to the new session error

Planlos5000: Assign the call to variables so that the error can be mapped to the new session error

server/src/users/service.go
Planlos5000: Missing new user error

Planlos5000: Assign the call to variables so that the error could be mapped to the new user error

Planlos5000: Missing new user error

server/src/users/service.go
```diff
@@ -93,9 +94,9 @@ func (service *Service) CreateAppleUser(ctx context.Context, id, name, avatarUrl
 
 	err := validateUsername(name)
 	if err != nil {
-		span.SetStatus(codes.Error, "failed to validate user name")
+		span.SetStatus(codes.Error, "failed to validate username")
 		span.RecordError(err)
-		return nil, common.BadRequestError(err)
+		return nil, CreateUserError(BadRequest, "failed to validate username", err)
```
Planlos5000: Validate already returns the user error. Just return this error without mapping

server/src/users/service.go
```diff
@@ -123,7 +124,7 @@ func (service *Service) CreateAzureAdUser(ctx context.Context, id, name, avatarU
 	if err != nil {
 		span.SetStatus(codes.Error, "failed to validate user name")
 		span.RecordError(err)
-		return nil, common.BadRequestError(err)
+		return nil, CreateUserError(BadRequest, "failed to validate username", err)
```
Planlos5000: Validate already returns the user error. Just return this error without mapping

server/src/users/service.go
```diff
@@ -151,7 +152,7 @@ func (service *Service) CreateGitHubUser(ctx context.Context, id, name, avatarUr
...
```
Planlos5000: Validate already returns the user error. Just return this error without mapping

server/src/users/service.go
```diff
@@ -179,7 +180,7 @@ func (service *Service) CreateGoogleUser(ctx context.Context, id, name, avatarUr
...
```
Planlos5000: Validate already returns the user error. Just return this error without mapping

server/src/users/service.go
```diff
@@ -207,7 +208,7 @@ func (service *Service) CreateMicrosoftUser(ctx context.Context, id, name, avata
...
```
Planlos5000: Validate already returns the user error. Just return this error without mapping

server/src/users/service.go
```diff
@@ -235,7 +236,7 @@ func (service *Service) CreateOIDCUser(ctx context.Context, id, name, avatarUrl
...
```
Planlos5000: Validate already returns the user error. Just return this error without mapping

server/src/users/service.go
```diff
@@ -264,7 +265,7 @@ func (service *Service) Update(ctx context.Context, body UserUpdateRequest) (*Us
...
```
Planlos5000: Validate already returns the user error. Just return this error without mapping

server/src/votings/service.go
Planlos5000: Missing new voting error

Planlos5000: Missing the new voting error

Planlos5000: Missing the new voting error

Planlos5000: Missing the new voting error

Planlos5000: Missing the new voting error

Planlos5000: Missing the new voting error

Planlos5000: Missing the new voting error

Planlos5000 (changes requested):
Missing feed back error type

nit picky maybe, but could you check the structure of the error.go files and make sure they are structured all the same? This would make it easier to read them.

server/src/api/errors.go
mating02: oh my bad you're absolutely right haha

server/src/api/boards.go
mating02: That is a good idea, it will take only a bit to ensure this is done for all span status lines

server/src/api/boards.go
```diff
@@ -130,17 +129,11 @@ func (s *Server) getBoard(w http.ResponseWriter, r *http.Request) {
...
-		if err == sql.ErrNoRows {
-			span.SetStatus(codes.Error, "no board found")
-			span.RecordError(err)
-			common.Throw(w, r, common.NotFoundError)
-			return
-		}
-
-		span.SetStatus(codes.Error, "failed to get board")
+		mappedErr := mapError(err)
+		span.SetStatus(codes.Error, mappedErr.Error())
```
Planlos5000: For me the string from `mappedErr.Error()` is empty when i tested it with killing the database.
To reproduce:
1. Start the server in debug mode and place a breakpoint before the request to the service
2. Create a new user
3. Create a board
4. access the board
5. Kill the database if the breakpoint is hit
6. step through in debug mode

If it is not easy to fix then just take a good string that says that it could either be a internal server error or the board is not found

server/src/boards/service.go
```diff
@@ -663,15 +663,15 @@ func (service *Service) BoardEditableContext(next http.Handler) http.Handler {
 			span.SetStatus(codes.Error, "failed to get board settings")
 			span.RecordError(err)
 			log.Errorw("unable to verify board settings", "err", err)
-			common.Throw(w, r, common.BadRequestError(errors.New("unable to verify board settings")))
+			common.Throw(w, r, common.BadRequestError(errors.New("unable to verify board settings"))) //replace this with domain error
```
Planlos5000: Why can this error not be replaced right now?

server/src/sessionrequests/service.go
```diff
@@ -228,6 +234,8 @@ func (service *BoardSessionRequestService) updatedSessionRequest(ctx context.Con
 
 }
 
+// this needs to be moved to the API layer
+
```
Planlos5000: We want to get rid of the api package in the long run. This should be either moved to a middleware package or to a middleware file for the package.

server/src/sessions/service.go
Planlos5000: Could you make here in the new function the same adjustments like in the `ModeratorSessionExists` method

.gitignore
```diff
@@ -45,3 +45,6 @@ deployment/**/.env
 
 # script output
 scripts/out
+
+#coverage files
+*coverage.txt
```
Planlos5000: I think we don't need this anymore with the change from #6199

server/src/boards/service.go
Planlos5000: Could you change the error return here to nil. This should not change any logic, but all other error cases are handled before

server/src/boardtemplates/service.go
```diff
@@ -70,7 +70,7 @@ func (service *Service) Create(ctx context.Context, body CreateBoardTemplateRequ
 		if err != nil {
 			span.SetStatus(codes.Error, "failed to create column template")
 			span.RecordError(err)
-			return nil, err
+			return nil, CreateBoardTemplateError(Internal, "failed to create column template", err)
```
Planlos5000: Here we can use the error that we receive from the column template service since this is already a new mapped error

server/src/notes/service.go
```diff
@@ -191,14 +192,14 @@ func (service *Service) Update(ctx context.Context, user uuid.UUID, body NoteUpd
 			err := errors.New("not allowed to stack notes")
 			span.SetStatus(codes.Error, "not allowed to stack notes")
 			span.RecordError(err)
-			return nil, common.ForbiddenError(err)
+			return nil, CreateNoteError(Forbidden, "not allowed to stack notes", err)
```
Planlos5000: Create the error on top of the if check and use it then in the span and as return value, like in board [Create](https://github.com/inovex/scrumlr.io/pull/6157/changes#diff-2402bceccae9fd2ab6fd2b985ef4143c0fff1d0cc6cc542723031ff6f7b6dcfcL156)
Also with the check before (181-184) and the next (199-202) check and line 254-257, 261-264, 279-282,

server/src/notes/service.go
```diff
@@ -288,7 +289,7 @@ func (service *Service) Delete(ctx context.Context, user uuid.UUID, body NoteDel
 		if err != nil {
 			span.SetStatus(codes.Error, "failed to get note stack")
 			span.RecordError(err)
-			return common.InternalServerError
+			return CreateNoteError(Internal, fmt.Sprintf("failed to get note stack: %v", err), err)
```
Planlos5000: Use the error that gets returned from the method since it is already returns the mapped error

server/src/reactions/service.go
```diff
@@ -103,15 +103,15 @@ func (service *Service) Create(ctx context.Context, body ReactionCreateRequest)
...
+		return nil, CreateReactionError(Internal, fmt.Sprintf("failed to get current reactions: %v", err), err)
 	}
 
 	for _, currentReaction := range currentReactions {
 		if currentReaction.User == body.User {
 			span.SetStatus(codes.Error, "multiple reactions not allowed")
 			span.RecordError(err)
 			log.Errorw("Cannot make multiple reactions on the same note by the same user", "user", body.User, "note", body.Note)
-			return nil, common.ConflictError(errors.New("cannot make multiple reactions on the same note by the same user"))
+			return nil, CreateReactionError(Conflict, "cannot make multiple reactions on the same note by the same user", err)
```
Planlos5000: There seams to be a bug here. The error that is used for the span is the error that is thrown by the database before which should be nil here all the time. Create here a new error and use it in the span and as a return value

server/src/reactions/service.go
```diff
@@ -155,15 +155,15 @@ func (service *Service) Delete(ctx context.Context, board, user, id uuid.UUID) e
 		span.SetStatus(codes.Error, "cannot remove reaction from other user")
 		span.RecordError(err)
 		log.Errorw("Unable to remove reaction from other users", "reactionUserId", reaction.User, "user", user)
-		return common.ForbiddenError(errors.New("forbidden"))
+		return CreateReactionError(Forbidden, "forbidden to delete other user's reaction", err)
```
Planlos5000: Same bug as in the `Create` function

server/src/reactions/service.go
```diff
@@ -194,15 +194,15 @@ func (service *Service) Update(ctx context.Context, board, user, id uuid.UUID, b
 		span.SetStatus(codes.Error, "cannot update reaction from other user")
 		span.RecordError(err)
 		log.Errorw("Unable to update reaction from other users", "reactionUserId", currentReaction.User, "user", user)
-		return nil, common.ForbiddenError(errors.New("forbidden"))
+		return nil, CreateReactionError(Forbidden, "forbidden to update other user's reaction", err)
```
Planlos5000: The bug. You know it (see `Create` and `Update`)

server/src/sessionrequests/service.go
Planlos5000, line 273: Here the `err` should be nil. Could you create here an error?

server/src/sessions/service.go
```diff
@@ -102,36 +102,36 @@ func (service *BoardSessionService) Update(ctx context.Context, body BoardSessio
...
 		log.Errorw("unable to get board session", "board", body.Board, "calling user", body.Caller, "error", err)
-		return nil, fmt.Errorf("unable to get session for board: %w", err)
+		return nil, CreateSessionError(Internal, fmt.Sprintf("unable to get session for board: %v", err), err)
 	}
 
 	if sessionOfCaller.Role == common.ParticipantRole && body.User != body.Caller {
 		span.SetStatus(codes.Error, "not allowed to change user session")
 		span.RecordError(err)
-		return nil, common.ForbiddenError(errors.New("not allowed to change other users session"))
+		return nil, CreateSessionError(Forbidden, "not allowed to change other users session", err)
```
Planlos5000: Create first the error and set it also in the span

server/src/sessions/service.go
```diff
@@ -102,36 +102,36 @@ func (service *BoardSessionService) Update(ctx context.Context, body BoardSessio
...
 		span.RecordError(err)
 		log.Errorw("unable to get board session", "board", body.Board, "target user", body.User, "error", err)
-		return nil, fmt.Errorf("unable to get session for board: %w", err)
+		return nil, CreateSessionError(Internal, fmt.Sprintf("unable to get session for board: %v", err), err)
 	}
 
 	if body.Role != nil {
 		if sessionOfCaller.Role == common.ParticipantRole && *body.Role != common.ParticipantRole {
-			err := common.ForbiddenError(errors.New("cannot promote role"))
+			err := CreateSessionError(Forbidden, "cannot promote role", err)
```
Planlos5000: The `err` should be nil here

server/src/users/service.go
Planlos5000: Map this error also to the new error

server/src/votings/service.go
```diff
@@ -128,30 +128,30 @@ func (service *Service) Create(ctx context.Context, body VotingCreateRequest) (*
...
 			span.RecordError(err)
-			return nil, common.BadRequestError(errors.New("only one open voting per session is allowed"))
+			return nil, CreateVotingError(BadRequest, "only one open voting per session is allowed", err)
 		}
 
 		span.SetStatus(codes.Error, "failed to get open votings")
 		span.RecordError(err)
-		return nil, common.InternalServerError
+		return nil, CreateVotingError(Internal, fmt.Sprintf("failed to get open votings: %v", err), err)
 	}
```
Planlos5000, line 153: For the case that there is already an open voting create first the error and then use it in the span and return value. For the second case you should be able to use the already mapped error from the return

Planlos5000 (commented):
The api methods for e.g. a board session request currently throw the new `SessionRequestError`. This error should be mapped or am i missing something? The new errors are currently only mapped for the board api methods.

For the error message: Sometimes the thrown error is now in the message of the error. I think if we would like to be not that verbose, then we should not add the error message e.g. from the database to the message.

server/src/boards/service.go
mating02: because this function is essentially an API function which uses the method common.Throw() to propagate errors back to the client. Or do you mean I should replace the common.BadRequestError(errors.New(...)) with the domain error? even in that case, since the domain error is not of type APIError, it would be ignored and a common.InternalServerError would be thrown in this case

server/src/sessionrequests/service.go
mating02: the error is sort of created with the common.NotFoundError, the creation of a domain error would be of limited value here since only the common.NotFoundError will be propagated anyway (and also the whole function is aimed to be transferred to the middleware anyway)

server/src/sessions/service.go
mating02: why?

server/src/votings/service.go
mating02: yeah i have just taken a look and I decided to also remove the span lines here since the span is already recording the error in the GetOpen method

server/src/api/notes_test.go
```diff
@@ -95,12 +89,6 @@ func (suite *NotesTestSuite) TestGetNote() {
 
 	testParameterBundles := *TestParameterBundles{}.
 		Append("all ok", http.StatusOK, nil, false, false, nil).
```
Planlos5000, line 92: Could you add a test for the not found case? Then both error cases from the get method are covered

server/src/api/notes_test.go
```diff
@@ -141,11 +129,11 @@ func (suite *NotesTestSuite) TestDeleteNote() {
 		},
 		{
 			name:         "Delete Note when board is locked",
-			expectedCode: http.StatusBadRequest,
+			expectedCode: http.StatusInternalServerError,
```
Planlos5000, line 134: Could you here add a two more tests that all error cases are covered? (Forbidden and Conflict)

server/src/api/notes_test.go
```diff
@@ -215,9 +203,9 @@ func (suite *NotesTestSuite) TestDeleteNote() {
 func (suite *NotesTestSuite) TestEditNote() {
 	testParameterBundles := *TestParameterBundles{}.
 		Append("all ok", http.StatusOK, nil, false, false, nil).
-		Append("api err", http.StatusForbidden, &common.APIError{
+		Append("api err", http.StatusInternalServerError, &common.APIError{
```
Planlos5000, line 238: Also here: Add tests for the missing errors (Forbidden and Conflict)

server/src/api/votes_test.go
```diff
@@ -32,12 +31,6 @@ func (suite *VoteTestSuite) TestAddVote() {
 
 	testParameterBundles := *TestParameterBundles{}.
 		Append("all ok", http.StatusCreated, nil, false, false, nil).
-		Append("specific error", http.StatusTeapot, &common.APIError{
```
Planlos5000, line 35: Could you add here a test case for the not found error?

server/src/api/votings_test.go
```diff
@@ -34,7 +33,6 @@ func (suite *VotingTestSuite) TestCreateVoting() {
 
 	testParameterBundles := *TestParameterBundles{}.
 		Append("all ok", http.StatusCreated, nil, false, false, nil).
-		Append("api error", http.StatusBadRequest, common.BadRequestError(errors.New("foo")), false, false, nil).
```
Planlos5000, line 37: Why is this case removed. The create method can throw an bad request error?

Planlos5000 (approved):

PR #6160, opened 2026-05-20
refactor/lint: eslint 9

Schwehn42:
## Description

upgrades to eslint (eslint 10 doesn't work just yet), but this is still a (or two) major steps 
- move from eslint.json to aslant.config.mjs
- fix about 250 new linting errors, including (not all are listed)
  - unused import
  - semantic intractable elements
  -  casts
  -  and more

## Changelog

- upgrade deps
- transfer config
- fix new issues, any in some cases, ignore if not relevant enough
- update tests 

## Checklist


src/api/auth.ts
```diff
@@ -15,7 +15,7 @@ export const AuthAPI = {
         credentials: "include",
       });
     } catch (error) {
-      throw new Error(`unable to sign out with error: ${error}`);
+      throw new Error(`unable to sign out with error: ${error}`, {cause: error});
```
Planlos5000: Is the message form the error here now duplicated?

Dockerfile
```diff
@@ -18,8 +18,7 @@ COPY vite-env-override.d.ts .
 COPY vite.config.ts .
 COPY index.html .
 COPY .prettierrc .
-COPY .eslintignore .
-COPY .eslintrc.json .
+COPY eslint.config.mjs
```
Planlos5000: Missing `.` for the location

src/api/auth.ts
Schwehn42: the thrown error is divided into `message`, containing above string, and `cause`, which contains the stack trace

so 
```ts
try {
  throw new Error("This is a simulated error");
} catch (error) {
     const finalError = new Error(`custom error: ${error}`, {cause: error});

      console.log("Message:", finalError.message);
      console.log("Cause:", finalError.cause);

      throw finalError;
```

will result in
```
Message: custom error: This is a simulated error
Cause: This is a simulated error
     at customError.ts:42
     at someFile:1283
```

nevertheless, I agree its enough to just have the caught error as the cause object

Schwehn42: error messages are now updated

Dockerfile
Schwehn42: oops you're right 🙈

src/components/ScrumlrLogo/ScrumlrLogo.tsx
```diff
@@ -45,21 +45,19 @@ export const ScrumlrLogo = ({className}: ScrumlrLogoProps) => {
...
-            // eslint-disable-next-line max-len
             d="M.884 47.523S48.86 32.596 109.857 30.136c1.44-.057 2.655.815 2.655 1.903 0 1.008-1.038 1.84-2.371 1.896-11.103.472-59.25 3.044-103.032 13.588 0 .162-6.225.973-6.225 0"
           />
         </g>
       </svg>
 
       <svg version="1.1" viewBox="0 0 44 48" className={classNames("scrumlr-logo", "scrumlr-logo--mobile", className)} xmlns="http://www.w3.org/2000/svg">
         <g className="scrumlr-logo__paths">
-          {/* eslint-disable-next-line max-len */}
+          {}
```
Planlos5000: Why are these empty brackets still there? Can they be removed?

src/utils/auth.ts
```diff
@@ -19,7 +19,7 @@ const signInAnonymously = async (displayName: string) => {
       store.dispatch(signIn({id: user.id, name: user.name, isAnonymous: true}));
     }
     return true;
-  } catch (err) {
+  } catch (_err) {
```
Planlos5000: In all other catch guards the error is called `error` or `_error`. Shouldn't that here also the case?

src/components/ScrumlrLogo/ScrumlrLogo.tsx
Schwehn42: yep, oversight from changing so many files^^

src/utils/auth.ts
Schwehn42: `_` prefix means its not being used anyway, so the naming might be off from other files but you can still identify it as the error imo. but if you like consistency throughout the files I can also change it

Planlos5000: i would prefere consistency. Or what would you prefere here?

Planlos5000 (approved):

PR #6185, opened 2026-06-01
feat: board history backend

SerseusWasTaken:
## Description

Backend part of #6141 
## Changelog

- updates board endpoint to include information required by the frontend
- DB: migrate DB, alter table sessions to add `favourite`
## Checklist


server/src/boards/service.go
Schwehn42, line 1: I noticed two things:
1. `createdAt` doesn't have the current date, but rather by the look of it, the epoch start date. If I remember correctly this was already the case earlier, but I guess it wasn't relevant enough yet to be fixed
2. Board users with the role `PARTICIPANT` should not be able to see hidden columns (possibly see `event_filter.go`)

SerseusWasTaken: Good catch with the column visibility, i will fix that.
Regarding `createdAt` im not sure if thats the case - the initial db migration defines table boards as: 
```sql
create table boards
(
    id                        uuid                   default gen_random_uuid() not null primary key,
    created_at                timestamptz            default now(),
    "name"                    varchar(128),
    access_policy             access_policy not null DEFAULT 'PUBLIC',
    passphrase                varchar(128),
    salt                      varchar(32),
    show_authors              boolean       not null DEFAULT true,
    show_notes_of_other_users boolean       not null DEFAULT true,
    allow_stacking            boolean       not null DEFAULT true,
    timer_end                 timestamptz
);
```

So i would expect created_at to be the expected timestamp, since now() is defined as "current transaction's start time" which sounds correct to me.

Schwehn42: I just checked, in the db, the `createdAt` value is indeed the expected `now()` date (e.g. `2026-06-01 11:19:59.081611 +00:00`), but the returned value in the endpoint is always `"createdAt": "0001-01-01T00:00:00Z"`

SerseusWasTaken: I'll go investigate

server/src/boards/dto.go
```diff
@@ -144,10 +147,13 @@ type BoardUpdateRequest struct {
 }
 
 type BoardOverview struct {
-	Board        *Board    `json:"board"`
-	Columns      int       `json:"columnsNumber"`
-	CreatedAt    time.Time `json:"createdAt"`
-	Participants int       `json:"participants"`
+	Board        *Board             `json:"board"`
+	Columns      []*columns.Column  `json:"columns"`
```
Planlos5000, line 151: Why do we now need all columns here and not just the number of columns? The history only shows the number of columns or am i wrong?

Planlos5000: Okay my bad. Since we want to display the column names we cannot just send the number of columns

Schwehn42: still it's not necessary to send the whole entity :D well I guess it depends on whether we do want to use the template editor for some of the features after all

Planlos5000 (approved):

PR #6197, opened 2026-06-03
fix: fix parallel map access of global state maps

SerseusWasTaken:
## Description

This PR introduces some thread safety to the Server functions. 
It also introduces two regression tests that can find two different race conditions.

I provide no gurantee that this covers all unsafe locations - these are just the problematic locations i found. 
## Changelog

- add sync.RWMutex to several levels of the Server struct
- cover map access with locks
- add regression tests to confirm thread safety
## Checklist


server/src/api/boards_listen_on_board.go
```diff
@@ -107,37 +119,60 @@ func (s *Server) openBoardSocket(w http.ResponseWriter, r *http.Request) {
...
+	s.boardSubscriptionsMu.Unlock()
+
+	b.mu.RLock()
+	// checking first and assigning the channel later is a ckeck-then-act race, but the alternative is to lock while performing I/O...
+	startListener := b.subscription == nil
+	b.mu.RUnlock()
+	var boardChannel chan *realtime.BoardEvent
+	if startListener {
+		boardChannel = s.realtime.GetBoardChannel(ctx, boardID)
+	}
```
SerseusWasTaken, line 139: This here needs a arch/design decision. My initial implementation for this had a lock for the complete duration of mutating `b` but this also means calling `s.realtime.GetBoardChannel(ctx, boardID)` while claiming a write lock, which sucks. This implementation, which first performs a check and releases the lock, has the trade-off of having a check-then-act race while avoiding the write lock during I/O. I am sticking with the check-then-act implementation here, because this does not change the behavior of prod (although the behavior is unsafe). The same issue can be found in websocket.go

PR #6199, opened 2026-06-03
refactor: refactored createUser functions

### LINKED ISSUE #6070: Refactor createUser function in service.go of users folder
### 📖 Background & Context
The `Service` struct currently has multiple methods for creating users based on their authentication provider (`CreateAnonymous`, `CreateAppleUser`, `CreateGitHubUser`, etc.). 

Every single one of these functions (found in [users/service.go](../blob/main/server/src/users/service.go) ) repeats the exact same ~20 lines of boilerplate:
1. Initializing an OpenTelemetry tracer span
2. Validating the username
3. Setting span attributes
4. Calling the database
5. Incrementing the global user counter
6. Incrementing the provider-specific counter

This introduces a high degree of redundancy.

### 🎯 Goal
Consolidate the boilerplate into a single, private helper function using a higher-order function approach. The public `Create*User` methods should be reduced to 3-4 line wrappers that delegate to this helper.

### 🛠 Proposed Implementation
Create a private helper method `createProviderUser` that accepts the dynamic parts (provider name, specific metric counter, and a closure for the database call).

mating02:
closes #6070 

- refactored create user functions for the individual platforms (apple, azure, ...) using a higher order CreatePlatformUser function effectively reducing the overall code size by a great margin

- changed tests for createUserAnonymous in the service_test.go file to expect InternalServerError and BadRequest (will be changed later when resolving merge conflict with issue 6157 to expect domain errors)

- corrected spelling mistakes for oidc

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ yarn-error.log*
 /coverage
 server/src/coverage.*
 src/utils/test/localstorage
+*coverage.txt
 
 # builds
 build
--- a/.gitignore
+++ b/.gitignore
@@ -13,9 +13,8 @@ yarn-error.log*
 
 # testing
 /coverage
-server/src/coverage.*
+**/coverage.*
 src/utils/test/localstorage
-*coverage.txt
 
 # builds
 build
--- a/server/src/users/otel_counter.go
+++ b/server/src/users/otel_counter.go
@@ -8,48 +8,6 @@ var userCreatedCounter, _ = meter.Int64Counter(
 	metric.WithUnit("users"),
 )
 
-var anonymousUserCreatedCounter, _ = meter.Int64Counter(
-	"scrumlr.users.anonymous.created.counter",
-	metric.WithDescription("Number of anonymous users created"),
-	metric.WithUnit("users"),
-)
-
-var appleUserCreatedCounter, _ = meter.Int64Counter(
-	"scrumlr.users.apple.created.counter",
-	metric.WithDescription("Number of apple users created"),
-	metric.WithUnit("users"),
-)
-
-var azureAdUserCreatedCounter, _ = meter.Int64Counter(
-	"scrumlr.users.azuread.created.counter",
-	metric.WithDescription("Number of azuread users created"),
-	metric.WithUnit("users"),
-)
-
-var githubUserCreatedCounter, _ = meter.Int64Counter(
-	"scrumlr.users.github.created.counter",
-	metric.WithDescription("Number of github users created"),
-	metric.WithUnit("users"),
-)
-
-var googleUserCreatedCounter, _ = meter.Int64Counter(
-	"scrumlr.users.google.created.counter",
-	metric.WithDescription("Number of google users created"),
-	metric.WithUnit("users"),
-)
-
-var microsoftUserCreatedCounter, _ = meter.Int64Counter(
-	"scrumlr.users.microsoft.created.counter",
-	metric.WithDescription("Number of microsoft users created"),
-	metric.WithUnit("users"),
-)
-
-var oidcUserCreatedCounter, _ = meter.Int64Counter(
-	"scrumlr.users.oidc.created.counter",
-	metric.WithDescription("Number of OIDC users created"),
-	metric.WithUnit("users"),
-)
-
 var deletedUserCounter, _ = meter.Int64Counter(
 	"scrumlr.users.deleted.counter",
 	metric.WithDescription("Number of deleted users"),
--- a/server/src/users/otel_counter.go
+++ b/server/src/users/otel_counter.go
@@ -15,7 +15,7 @@ var anonymousUserCreatedCounter, _ = meter.Int64Counter(
 )
 
 var appleUserCreatedCounter, _ = meter.Int64Counter(
-	"scrumlr.users.aplle.created.counter",
+	"scrumlr.users.apple.created.counter",
 	metric.WithDescription("Number of apple users created"),
 	metric.WithUnit("users"),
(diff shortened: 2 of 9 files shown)


Planlos5000:
Why not just simply create a function that takes the context, ident, name, avatar and the type in the service that then creates the user

```
func (service *Service) CreateUser(ctx context.Context, id, name, avatarUrl string, accountType common.AccountType)
```

Let the function then switch handle the different types by calling the specific database function.
Then we could delete the specific functions in the service. This would then also move some logic from the api layer to the service layer.

mating02:
> Why not just simply create a function that takes the context, ident, name, avatar and the type in the service that then creates the user
> 
> ```
> func (service *Service) CreateUser(ctx context.Context, id, name, avatarUrl string, accountType common.AccountType)
> ```
> 
> Let the function then switch handle the different types by calling the specific database function. Then we could delete the specific functions in the service. This would then also move some logic from the api layer to the service layer.

Good idea, the thing is I purposely did it that way because the signature of the individual functions would still be present, so a bigger refactoring is not needed. If I just use one CreateUser function, I would have to remove all calls of the individual functions and a greater refactoring would be necessary. I could do that though, I think it's a good idea if you are still in favor of your proposal. Are you?

server/src/users/service.go
```diff
@@ -59,200 +60,61 @@ func NewUserService(db UserDatabase, rt *realtime.Broker, sessionService session
...
-}
-
-func (service *Service) CreateAppleUser(ctx context.Context, id, name, avatarUrl string) (*User, error) {
-	ctx, span := tracer.Start(ctx, "scrumlr.users.service.create.apple")
+	ctx, span := tracer.Start(ctx, traceName)
 	defer span.End()
 
-	err := validateUsername(name)
-	if err != nil {
+	// Validation
```
Planlos5000: Remove the comments

server/src/users/service.go
```diff
@@ -59,200 +60,61 @@ func NewUserService(db UserDatabase, rt *realtime.Broker, sessionService session
...
-		attribute.String("scrumlr.users.service.create.anonymous.name", name),
-	)
-
-	user, err := service.database.CreateAnonymousUser(ctx, name)
-	if err != nil {
-		span.SetStatus(codes.Error, "failed to create user")
-		span.RecordError(err)
-		return nil, err
-	}
+func (service *Service) CreateUser(ctx context.Context, id, name, avatarUrl string, accountType common.AccountType, specificCounter metric.Int64Counter) (*User, error) {
```
Planlos5000: Could we remove the counter from the parameter list. This should be a implementation detail and nothing a caller should worry about.
Maybe select the counter that should be called in the switch statement?

server/src/users/otel_counter.go
```diff
@@ -8,48 +8,6 @@ var userCreatedCounter, _ = meter.Int64Counter(
...
-	metric.WithDescription("Number of anonymous users created"),
-	metric.WithUnit("users"),
-)
-
-var oicdUserCreatedCounter, _ = meter.Int64Counter(
-	"scrumlr.users.oicd.created.counter",
-	metric.WithDescription("Number of anonymous users created"),
-	metric.WithUnit("users"),
-)
-
```
Planlos5000, line 52: Why are these deleted?

server/src/users/service.go
```diff
@@ -59,200 +60,61 @@ func NewUserService(db UserDatabase, rt *realtime.Broker, sessionService session
...
+	platformName := strings.ToLower(string(accountType))
+	traceName := fmt.Sprintf("scrumlr.users.service.create.%s", platformName)
```
Planlos5000: I think we can go here only with `scrumlr.users.service.create` since we then only have one create function

server/src/users/otel_counter.go
mating02: because due to the unification, they are not used anymore and caused an issue in the linter

server/src/users/service.go
mating02: good catch! will do

server/src/users/otel_counter.go
Planlos5000: But we could use them after we created the user in the database, or am i missing something here?

mating02: No you are absolutely correct, I made an unintelligent decision giving the responsibility for the specificCounter to the caller. So that makes these variables relevant again! I will restore this file and remove the specificCounter from the parameter list.

.gitignore
```diff
@@ -15,6 +15,7 @@ yarn-error.log*
 /coverage
 server/src/coverage.*
 src/utils/test/localstorage
+*coverage.txt
```
Planlos5000: Why do we need to include this in the ignore? The coverage files from the backend tests should be included in line 16

server/src/users/service.go
```diff
@@ -59,200 +59,63 @@ func NewUserService(db UserDatabase, rt *realtime.Broker, sessionService session
...
-	)
-
-	user, err := service.database.CreateAnonymousUser(ctx, name)
-	if err != nil {
-		span.SetStatus(codes.Error, "failed to create user")
-		span.RecordError(err)
-		return nil, err
-	}
+func (service *Service) CreateUser(ctx context.Context, id, name, avatarUrl string, accountType common.AccountType) (*User, error) {
+	traceName := "scrumlr.users.service.create"
```
Planlos5000: nit picky: in every other file in the project we write the trace name directly into the start function

Planlos5000 (approved):
LGTM

PR #6317, opened 2026-07-09
feat: disallow moderators from deleting boards

### LINKED ISSUE #6117: Moderator ist not allowed to delete boards
### The problem

Currently, a person designated as a moderator is also permitted to delete sessions/boards.

### Browser

Chrome

1. Open a board
2. Make sure to have at least 2 user on the board
3. Appoint one of them as moderator
4. Go to Settings (Board settings) & scroll to the end
5. As you can see, the button "Delete session" is also available for the moderators  

### Screenshots

_No response_

### Additional context

Acceptance Criteria:
- With the "Moderator" role, it is not possible to delete a board
- Only the board owner has this permission

Schwehn42:
## Description

fixes https://github.com/inovex/scrumlr.io/issues/6117

Currently, moderators are privileged to do actions such as deleting whole boards. this should be limited to the owner only.

This PR adds a new context, `BoardOwnerContext` which passes the board owner only. this is applied to the delete board route. 

In the Frontend, only owners will see the delete board button.

## Changelog

- add `BoardOwnerContext`
- use `BoardOwnerContext` for DELETE board route
- FE: only show delete board button for board owner

## Checklist

### DIFF
--- a/server/src/sessions/service.go
+++ b/server/src/sessions/service.go
@@ -32,6 +32,7 @@ type SessionDatabase interface {
 	UpdateAll(ctx context.Context, update DatabaseBoardSessionUpdate) ([]DatabaseBoardSession, error)
 	Exists(ctx context.Context, board, user uuid.UUID) (bool, error)
 	ModeratorExists(ctx context.Context, board, user uuid.UUID) (bool, error)
+	OwnerExists(ctx context.Context, board, user uuid.UUID) (bool, error)
 	IsParticipantBanned(ctx context.Context, board, user uuid.UUID) (bool, error)
 	Get(ctx context.Context, board, user uuid.UUID) (DatabaseBoardSession, error)
 	GetAll(ctx context.Context, board uuid.UUID, filter ...BoardSessionFilter) ([]DatabaseBoardSession, error)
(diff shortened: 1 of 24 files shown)


server/src/sessions/service.go
```diff
@@ -335,6 +336,18 @@ func (service *BoardSessionService) ModeratorSessionExists(ctx context.Context,
 	return service.database.ModeratorExists(ctx, boardID, userID)
 }
 
+func (service *BoardSessionService) OwnerSessionExists(ctx context.Context, boardID, userID uuid.UUID) (bool, error) {
+	ctx, span := tracer.Start(ctx, "scrumlr.sessions.service.exists.owner")
+	defer span.End()
+
+	span.SetAttributes(
+		attribute.String("scrumlr.sessions.service.exists.owner.board", boardID.String()),
```
Schwehn42, line 344: btw can I just define it like this or does the tracer work on predefined attribute names?

Planlos5000 (changes requested):
Could you add some tests for the new functions you implemented?

server/src/sessions/service.go
Planlos5000: You can define it like this. The trace has no predefined attributes. There some conventions and we should be consistent in our naming

mating02 (approved):
New tests cover the functionality checks of boardowner sessions and moderator sessions, deployment functionality was tested locally, all good!

Planlos5000 (approved):

PR #6352, opened 2026-07-17
feat: scrumlr configuration

### LINKED ISSUE #6319: Database configuration
Currently the database is configured using the database url through the `-d` flag or the env variable `SCRUMLR_SERVER_DATABASE_URL`.
We want to allow to configure the database also through four env variables that supply the host, username, password and database name. The database url is the build on startup of the server.

If the server is configured with both database configurations (url and host, user, password, db name), the server should either not start and throw an error or select one configuration.

This feature should also be documented in the scrumlr docs

- [x] Add database environment variables
- [x] Handle multi configuration
- [ ] Document the variables

Planlos5000:
## Description

This merge request updates the configuration for the scrumlr backend and adds a new way to configure the database via three environment variables (`SCRUMLR_SERVER_DATABASE_HOST`, `SCRUMLR_SERVER_DATABASE_USER`, `SCRUMLR_SERVER_DATABASE_PASSWORD`). It also adds tests for the initialization of the cache, database and message broker.

## Changelog

- Update database initialization
- Add `SCRUMLR_SERVER_DATABASE_HOST`, `SCRUMLR_SERVER_DATABASE_USER`, `SCRUMLR_SERVER_DATABASE_PASSWORD` env vars
- Add tests for database initialization
- Update cache initialization
- Add tests for cache initialization
- Update realtime initialization
- Add tests for realtime initialization
- Update auth provider configuration
- Update documentation (curently missing)

## Checklist

### DIFF
--- a/docs/src/content/docs/self-hosting/env-vars.md
+++ b/docs/src/content/docs/self-hosting/env-vars.md
@@ -109,6 +109,8 @@ Only configure one of the two.
 SCRUMLR_SERVER_NATS_URL=''
 ```
 
+**Note**: If a redis and nats address are defined, the redis address takes precedence and will be used.
+
 ### Redis Variables
 
 You can either use a NATS server or a Redis server for the backend communication.
@@ -119,6 +121,8 @@ SCRUMLR_SERVER_REDIS_HOST=''
 SCRUMLR_SERVER_REDIS_USERNAME=''
 SCRUMLR_SERVER_REDIS_PASSWORD=''
 ```
+
+**Note**: If a redis and nats address are defined, the redis address takes precedence and will be used.
 
 ### Insecure Mode
 
@@ -138,15 +142,30 @@ Please generate a new key for production use and keep it secure.
 SCRUMLR_PRIVATE_KEY=''
 ```
 
-### Database URL
+### Database
 
-The URL of the PostgreSQL database.
+To configure the database you can choose between two options.
+The first option is to configure the URL of the PostgreSQL database.
 Credentials are passed in the URL.
 If you havent configured postgres for TLS, you can use the `?sslmode=disable` parameter.
 
 ```ini
 SCRUMLR_SERVER_DATABASE_URL='postgres://user:password@host:port/database'
 ```
+
+The second option is to configure the database with the following three variables
+
+```ini
+SCRUMLR_SERVER_DATABASE_HOST=host:port/database
+SCRUMLR_SERVER_DATABASE_USER=user
+SCRUMLR_SERVER_DATABASE_PASSWORD=password
+```
+
+From these variables the database url is created. All three variables must be set that this option to be used.
+If you haven't configured postgres for TLS, you can use the `?sslmode=disable` parameter for the database host.
+
+**Note:** If both options are configured, the database url takes precedence and will be used.
+
 
 ### Base Path
 
--- a/server/src/cache/cache.go
+++ b/server/src/cache/cache.go
@@ -2,10 +2,13 @@ package cache
 
 import (
 	"context"
+	"errors"
 	"time"
 
+	"github.com/urfave/cli/v2"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/trace"
+	"scrumlr.io/server/logger"
 )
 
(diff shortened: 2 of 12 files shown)


docs/src/content/docs/self-hosting/env-vars.md
```diff
@@ -138,16 +142,31 @@ Please generate a new key for production use and keep it secure.
...
 
+The second option is to configure the database with the following three variables
+
+```ini
+SCRUMLR_SERVER_DATABASE_HOST=host:port/database
+SCRUMLR_SERVER_DATABASE_USER=user
+SCRUMLR_SERVER_DATABASE_PASSWORD=password
+```
+
+From these variables the database url is created. All three variables must be set that this option is used.
```
mating02: All three variables must be set so that this option can be used.*

docs/src/content/docs/self-hosting/env-vars.md
```diff
@@ -138,16 +142,31 @@ Please generate a new key for production use and keep it secure.
...
+If you havent configured postgres for TLS, you can use the `?sslmode=disable` parameter for the database host.
```
mating02: haven't

server/src/cache/cache.go
```diff
@@ -20,3 +23,33 @@ type Client interface {
...
+	if ctx.String("nats") != "" {
+		address := ctx.String("nats")
+
+		log.Infof("Connecting to nats at %v as cache", address)
+
+		cache, err := NewNats(address, "scrumlr")
+		return cache, err
+	}
+
+	return cache, errors.New("no valid cache configuration found")
```
mating02, line 54: probably nit picky, but wouldn't it make more sense to explicitly return nil instead of the cache variable (which of course is nil anyway) for better clarity?

Planlos5000: The `cache` variable should be used here as the return value or this variable is marked as unused

server/src/initialize/database_test.go
mating02, line 1: tests are successful, but I'd suggest refactoring the beginning of every individual function. The beginning of every test checks if the container has started successfully and if the host and mapped port can be obtained. Maybe a helper function like `func startTestContainer(t *testing.T, ctx context.Context) (host string, port string)` could get rid of this redundancy. For the third case you could then just do _, _ = startTestContainer(t, ctx), since you don't care about the port and host here. Maybe another helper like `func setupCLIContext(t *testing.T) *cli.Context` would also help get rid of the redundant flagset lines. These are just suggestions in terms of coding style though, the tests are covering everything and are working perfectly.

server/src/cache/cache_test.go
mating02, line 1: I'd also suggest a helper in a shape like `func setupCLIContext(t *testing.T) *cli.Context` here. Also, I just realized that the names of the tests are a bit confusing. You are writing TestInitializeRealtime_[...] but it should be TestInitializeCache_[...] since that is the function you are testing here. The tests are working perfectly fine though

server/src/realtime/broker_test.go
mating02, line 1: I'd also suggest a helper in a shape like `func setupCLIContext(t *testing.T) *cli.Context` here.

mating02 (changes requested):

server/src/cache/cache_test.go
Planlos5000: I have fixed the naming of the tests. Copy paste error

server/src/initialize/database_test.go
Planlos5000: I am not a big fan of helper functions for tests. For the database we already have a setup method for the testcontainer, but this method also sets up the test template stuff for the integration tests.
For the cli context setup i think it is good to see for each test what is setup and what not and that you don't have to jump in the file

mating02: okay I see your point for the container setup, but the flagset lines are exactly the same for each test, so I think we could at least remove this redundancy:

`flagset := flag.NewFlagSet("scrumlr-tests", flag.ExitOnError)
	flagset.String("database", "", "")
	flagset.String("database-host", "", "")
	flagset.String("database-username", "", "")
	flagset.String("database-password", "", "")
	init := cli.NewContext(nil, flagset, nil)`

mating02 (approved):
locally tested, code readability fine, good to go

PR #6367, opened 2026-07-21
feat: Add swagger page

### LINKED ISSUE #6339: Documentation REST API
We want to document our REST API for the scrumlr backend. 
For that we want to create a OpenAPI file. Here we have two options
1. Let the OpenAPI file be generated with e.g. [https://github.com/swaggo/swag](https://github.com/swaggo/swag). For this approach we need to annotate all API endpoints.
2. Write a OpenAPI file and let the REST API be generated and implement the generated interfaces with e.g. [https://github.com/oapi-codegen/oapi-codegen](https://github.com/oapi-codegen/oapi-codegen). This would require to reimplement the API. 

This file could then be used to host a swagger page in development mode.

- [ ] Choose a tool
- [ ] Setup an OpenAPI file
- [ ] (Would be nice) Setup a Swagger page for development mode

Planlos5000:
## Description

Adds a swagger page to document the scrumlr backend api.
## Changelog

- Configure swagger page
- Add make command to generate swagger
## Checklist


tohega (approved):
Nice this will really help for development from now on.

I have also added a commit to fix typos and because it didnt work out of the box for me, but the cause was two undocumented environment prerequisites for macOS, not the PR itsself.

Basically macOS needs GNU Make ≥ 3.82 otherwise it silently ignores `.ONESHELL` because it ships with GNU Make 3.81, which led to `cd src` commands not persisting and then `make swagger` fails with `go: cannot find main module`.

My follow up commit documents it now and upon regenerating with the pinned `swag` version from `go.mod (v1.16.6)` it also edited the sligthly stale spec. (It now added missing `contact.email` and a different generated-header format).

Issue #6338, opened 2026-07-13
Edit board: template edit mask


SelinaBuff:
### Description:
Currently, there is no functionality to edit existing boards from the History page. To allow board owners to modify their boards without having to recreate them, we need to implement a new edit workflow.
This feature requires building an integration between the History page and the Template Editor mask: when a board owner triggers the edit action, the editor needs to open pre-populated with the board's data, adapt its UI elements to an "edit/save" state instead of "create", and handle the update and redirect flow back to the History page.

### User Story:
**As a** board wwner,
**I want** to edit my boards directly from the History page using the template editor,
**so that** I can easily update board details (like title, description, and columns) without having to recreate the board from scratch.

### Acceptance Criteria:
- [ ] **Accessing the Editor:**
- A board owner can click on the 3-dot menu of a board on the History page
- Clicking the pencil icon (Edit) within this menu opens the Template Editor mask

- [ ] **Permissions:**
The different roles have the following functions:
- Owner: Every function is available
- Moderator: "Delete" is not available
- Participant: "Delete" and "edit" is not available
_(Note: These options must either be completely hidden or removed from the 3-dot menu for that specific board)_

<img width="560" height="519" alt="Image" src="https://github.com/user-attachments/assets/c9d1ce4d-734b-4d8a-81a2-e4f521ce924e" />

- [ ] **Pre-Populated Fields:**
When the Template Editor opens for an existing board, all fields must be pre-filled with the board's current data:
- Board Title
- Description
- Columns (including column titles, colors, and current order)

- [ ] **UI Adjustments (Primary Button):**
Since the user is editing an existing board rather than creating a new one, the primary button on the right side must change as follows:
- Text changes from "Create template" to "Save"
- Icon changes from plus-icon (+) to checkmark-icon (✓)

- [ ] **Save and Redirect:**
- Clicking the updated "Save" button saves the changes
- Upon saving, the user is automatically redirected back to the History page
- The updated details of the edited board must be immediately visible on the History page

PR #6372, opened 2026-07-21
feat: edit boards from the History Page

tohega:
Issue: #6338 

## Description

Adds a board editor reachable from the History page card menu, letting owners and moderators edit a boards title, description and columns without recreating it. Extracts a shared EditorShell out of the TemplateEditor so templates and boards reuse the same editing UI and introduces a history Redux slice to drive the list. 

Wired to the backend (GET/PUT/DELETE /boards plus the participant endpoint for favourite), with a delete confirmation dialog, copy-link, and role-gated actions (owner: edit + delete, moderator: edit only, participant: neither).
## Checklist


src/routes/Router.tsx
```diff
@@ -101,6 +102,8 @@ const Router = () => {
             </Route>
           </Route>
 
+          <Route path="edit-board/:boardId" element={<BoardEditor />} />
```
Schwehn42: I noticed that template editing is `boards/edit/:templateId` whereas board editing is `boards/edit-board/:boardId`. maybe it would be more fitting to have either `boards/edit-template` or `templates/edit` for the url? but to be fair that might overcomplicate it

src/routes/Boards/Sessions/History.tsx
```diff
@@ -1,74 +1,36 @@
...
+
+  useEffect(() => {
+    if (historyBoards.length === 0) dispatch(getBoards());
+  }, [dispatch, historyBoards.length]);
+
   const matchSearchInput = (historyBoard: HistoryBoard) => historyBoard.name.toLowerCase().includes(searchBarInput.toLowerCase());
 
   // for development, only show history boards if the env variable is set to true, otherwise show the teaser
   const renderHistoryContent = () =>
     import.meta.env.VITE_SHOW_HISTORY_PAGE === "true" ? (
```
Schwehn42: I reckon we don't require the env toggle anymore

src/routes/Boards/Sessions/HistoryCard/HistoryCard.tsx
```diff
@@ -89,34 +101,41 @@ export const HistoryCard = (props: HistoryCardProps) => {
...
+                },
+              ]
+            : []),
           {
             label: t("History.HistoryCard.Menu.copyLink"),
             element: <LinkIcon />,
             onClick: () => {
-              throw new Error("Not implemented yet");
+              navigator.clipboard.writeText(`${window.location.origin}/board/${props.board.id}`);
+              Toast.success({title: t("History.HistoryCard.linkCopied")});
```
Schwehn42: wrap the success toast in the clipboard `.then` callback, in case it fails for whatever reason

src/store/features/history/thunks.ts
```diff
@@ -0,0 +1,91 @@
...
+    const userId = getState().auth.user?.id;
+    if (userId) {
+      const updateFavouriteOnServer = () => API.editParticipant(boardId, userId, {favourite});
+      const retrySetBoardFavourite = () => setBoardFavourite({boardId, favourite});
+
+      await retryable(
+        updateFavouriteOnServer,
+        dispatch,
+        retrySetBoardFavourite,
+        "editSelf"
```
Schwehn42: is `"editBoard"` a more fitting error key?

src/routes/Boards/History/History.tsx
```diff
@@ -1,74 +1,36 @@
...
   const matchSearchInput = (historyBoard: HistoryBoard) => historyBoard.name.toLowerCase().includes(searchBarInput.toLowerCase());
 
   // for development, only show history boards if the env variable is set to true, otherwise show the teaser
   const renderHistoryContent = () =>
     import.meta.env.VITE_SHOW_HISTORY_PAGE === "true" ? (
-      TEST_HISTORY_BOARDS.filter(matchSearchInput)
+      historyBoards
+        .filter(matchSearchInput)
         .sort((a, b) => Number(b.favourite) - Number(a.favourite)) // move favourites to the top using the fact that true is 1 and false is 0
         .map((hb) => <HistoryCard key={hb.id} board={hb} />)
```
Schwehn42: we might want to reconsider our sorting, as new sessions are displayed at the very bottom. 

maybe we should sort by matchSearchInput -> favourites -> lastModifiedDate

src/routes/Boards/Sessions/History.tsx
```diff
@@ -1,74 +1,36 @@
...
 export const History = () => {
   const {t} = useTranslation();
 
   const {searchBarInput} = useOutletContext<{searchBarInput: string}>();
 
+  const dispatch = useAppDispatch();
+  const historyBoards = useAppSelector((state) => state.history);
+
+  useEffect(() => {
+    if (historyBoards.length === 0) dispatch(getBoards());
```
Schwehn42: that means clicking on templates and then history again doesn't update changes which were possibly made during that time

src/routes/Boards/Editor/EditorShell.tsx
Schwehn42, line 1: - the component location should be `Boards/EditorShell`
- types and css classes should reflect the new component name

Schwehn42 (commented):
looks pretty good all in all!

src/routes/Router.tsx
Schwehn42: changed template edit route to `template-edit/:id`

jmoehler:
This is how the register yourself is suppose to look:

image


and this is how it currently looks:

image


apart from the registration not being the same (which I find no problem) the text and the i is almost not readable in dark mode

jmoehler:
image

when one clicks on that button on the same one twice, it goes to the other. so if I hit history page while not active it goes to history page, but if I hit history page while active it goes to templates

jmoehler:
same goes for 
image

src/routes/Boards/Editor/EditorShell.tsx
```diff
@@ -0,0 +1,184 @@
...
+import "./EditorShell.scss";
+
+export type EditorSubmitPayload = {
+  name: string;
+  description: string;
+  columns: EditableTemplateColumn[];
+  deletedColumns: EditableTemplateColumn[];
+};
+
+// shared editor used to create/edit a template or a board (from history).
```
jmoehler, line 21: quite a lot of comments in here. is that really needed?

src/routes/Boards/History/History.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const matchSearchInput = (historyBoard: HistoryBoard) => historyBoard.name.toLowerCase().includes(searchBarInput.toLowerCase());
+
+  const renderRequireRegisteredUser = () => (
+    <div className={"history__info"}>
+      <div className="history__info-container">
+        <InfoIcon className="history__info-icon" />
+        <p className="history__info-text">{t("History.requireRegisteredAccount")}</p>
+      </div>
+      <div className="history__require-registered-user-register-area">
+        {/*redirect to the login screen and after and then back to here after completion.
```
jmoehler, line 45: not really needed that comment

src/routes/Boards/History/History.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const renderEmptyHistory = () => (
+    <div className="history__info">
+      <div className="history__info-container">
+        <InfoIcon className="history__info-icon" />
+        <p className="history__info-text">{t("History.empty")}</p>
+      </div>
+    </div>
+  );
+
+  // show history for registered users or if the respective env flag is set and if there are history boards, otherwise an info section
```
jmoehler, line 63: again

src/store/features/history/__tests__/thunks.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        columns: [
+          col({id: "new-col", name: "Added", color: "goal-green", index: 2, persisted: false, mode: "create"}),
+          col({id: "c2", name: "Renamed", color: "planning-pink", index: 1, mode: "edit"}),
+          col({id: "c3", name: "Untouched", index: 0, mode: undefined}),
+        ],
+        deletedColumns: [col({id: "c4", name: "Removed", mode: "delete"})],
+      })
+    );
+
+    // board metadata goes through editBoard
```
jmoehler: lots of comments here

src/routes/Boards/History/HistoryCard/HistoryCard.tsx
```diff
@@ -233,6 +244,19 @@ export const HistoryCard = (props: HistoryCardProps) => {
           </Tooltip>
         )}
       </div>
+      {showDeleteConfirmation && (
+        <ConfirmationDialog
+          title={t("History.HistoryCard.deleteConfirmationTitle")}
+          text="History.HistoryCard.deleteConfirmationText"
```
jmoehler: should not be plain text

Image

src/routes/Boards/History/History.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const renderRequireRegisteredUser = () => (
+    <div className={"history__info"}>
+      <div className="history__info-container">
+        <InfoIcon className="history__info-icon" />
+        <p className="history__info-text">{t("History.requireRegisteredAccount")}</p>
+      </div>
+      <div className="history__require-registered-user-register-area">
+        {/*redirect to the login screen and after and then back to here after completion.
+          NOTE: possible subject to change, e.g., by replacing it with an 'upgrade account' functionality*/}
+        <Button className="history__require-registered-user-register-button" onClick={() => navigate("/login", {state: {from: {pathname: "boards/history"}}})}>
```
jmoehler: according to AI, `"boards/history"` is missing a leading slash and would therefore break the redirect. I can not test this as on the test deployment anonymous history page is activated and on local dev I can not be not anonymous.

src/routes/Boards/Boards.tsx
jmoehler: this is still on `edit` instead of `edit-board`. results in:

Image
then click on user pill:

Image

src/store/features/history/thunks.ts
```diff
@@ -0,0 +1,85 @@
...
+        description: column.description,
+        color: column.color,
+        visible: column.visible,
+        index: column.index,
+      })
+    ),
+    ...columnsToDelete.map((column) => API.deleteColumn(boardId, column.id)),
+  ]);
+
+  await dispatch(getBoards());
```
jmoehler: doesn't unwrap the thing, a rejected refresh doesn't reject the thunk.

src/routes/Boards/Editor/TemplateEditor/__tests__/TemplateEditor.test.tsx
jmoehler: still uses edit instead of edit-template

src/routes/Boards/History/History.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const dispatch = useAppDispatch();
+  const historyBoards = useAppSelector((state) => state.history);
+  const isAnonymous = useAppSelector((state) => state.auth.user?.isAnonymous);
+  const allowAnonymousHistory = useAppSelector((state) => state.view.allowAnonymousHistory);
+
+  const canViewHistory = !isAnonymous || allowAnonymousHistory;
+
+  // init history boards
+  useEffect(() => {
```
jmoehler, line 30: runs unconditionally, only checks at render if even accessible.

jmoehler (dismissed):

src/routes/Boards/History/History.tsx
Schwehn42: used to be conditionally (if history.length === 0), but when the user would switch between templates and history, they would not get the most recent state which I thought wasn't ideal. What do you reckon?

jmoehler: pherhaps we add something like:

`if (!canViewHistory) return;`

in it so it wouldnt run for accounts who don't have access anyways.

src/routes/Boards/Editor/EditorShell.tsx
Schwehn42: why not? its a complex component and the comments don't state obvious things imo

src/routes/Boards/History/History.tsx
Schwehn42: is the behavior that after completion of the login the user will get redirected back to the history page apparent to you by glancing at the code? to me it isn't

src/routes/Boards/History/History.tsx
Schwehn42: and again, I don't see what's wrong about having the comment

src/store/features/history/__tests__/thunks.test.ts
Schwehn42: I can agree there is some redundancy here, what would you suggest to do about it?

src/store/features/history/thunks.ts
Schwehn42: in fact the dispatch can be omitted since after saving and returning to the history view, the data is refetched fresh again

Planlos5000 (approved):

Issue #6129, opened 2026-05-18
Define Roles/Permissions


SelinaBuff:
During the sprint (most likely on June 8th, Refinement session), we will discuss the permissions associated with our roles (Moderator, Owner, and Standard Participant).

- [ ] Own package for the roles (Backend) 

https://miro.com/app/board/uXjVIaFE38I=/

SelinaBuff:
We will discuss this in our next workshop (July).

PR #6400, opened 2026-07-24
feat: update roles

Planlos5000:
## Description

Update the roles and add helper methods for checking if a user is allowed to perform an action or to check if a user is allowed to see the requested data. This will help us to change permissions easily for #6129 
## Changelog

- Add role package
- Move roles from common package to role package
- Add helper methods for the roles
- Add tests for the helper methods 
## Checklist


tohega (approved):
Very clean with role test at 100% coverage and it reads much better now with the positive allowlist (CanX()... instead of == ParticipantRole)

PR #6401, opened 2026-07-24
feat: Add delete session

### LINKED ISSUE #6320: Backend history implementation
For the history page we need a `DELETE` endpoint to remove sessions from a user.  This will require the following endpoint to be added: `DELETE /boards/{boardId}/participants/{session}`. For this endpoint we also need the corresponding functions to be added to delete a session. 

- [x] Add delete endpoint
- [x] add delete functions for the service and database
- [x] Test the functions
- [x] Test that no notes, votes or votings are deleted if a session is deleted
- [x] Test that non anonymous users can rejoin a board after the session was deleted

Planlos5000:
## Description

This will add a delete endpoint to remove board sessions. Only a user can remove his own session. If a user is the owner of a board, the user cannot remove this session. After a user removed a session the notes are anonymized (like import).
No notes are deleted when a session is deleted. A user can rejoin the board after the session was deleted.
## Changelog

- Add delete endpoint for board sessions
- Add tests for deleting board sessions
- Update mocks
- Update swagger page
## Checklist

### DIFF
--- a/server/src/sessions/api.go
+++ b/server/src/sessions/api.go
@@ -30,10 +30,17 @@ type SessionService interface {
 	OwnerSessionExists(ctx context.Context, boardID, userID uuid.UUID) (bool, error)
 	IsParticipantBanned(ctx context.Context, boardID, userID uuid.UUID) (bool, error)
 	BoardSessionFilterTypeFromQueryString(query url.Values) BoardSessionFilter
+	Delete(ctx context.Context, callerID, boardID, userID uuid.UUID) error
 }
 
 type API struct {
 	service SessionService
+}
+
+func NewSessionApi(service SessionService) SessionApi {
+	api := new(API)
+	api.service = service
+	return api
 }
 
 // Get all sessions for a board
@@ -182,7 +189,6 @@ func (api *API) UpdateBoardSession(w http.ResponseWriter, r *http.Request) {
 //	@Accept			json
 //	@Param			Cookie	header	string						true	"jwt token to authenticate"
 //	@Param			boardId	path	string						true	"id of the board"
-//	@Param			id		path	string						true	"id of the session"
 //	@Param			session	body	BoardSessionUpdateRequest	true	"values to update the session"
 //	@Produce		json
 //	@Success		200	{object}	sessions.BoardSession	"updated session for the board"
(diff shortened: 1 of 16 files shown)


server/src/sessions/service.go
```diff
@@ -361,6 +362,47 @@ func (service *BoardSessionService) IsParticipantBanned(ctx context.Context, boa
 	return service.database.IsParticipantBanned(ctx, boardID, userID)
 }
 
+func (service *BoardSessionService) Delete(ctx context.Context, callerID, boardID, userID uuid.UUID) error {
```
tohega, line 393: Every other mutation in this service ends with a broadcast (createdSession, updatedSession, updatedSessions) but the Delete here returns after the DB call. I think Delete needs to notify the other clients before it returns like the neighbouring functions do. 

(Otherwise it leads to this unexpected behaviour: A as owner, B as participant in one Board Session. B writes a note. We delete B's session (204 OK). Nothing changes on A's screen: B is still listed as a participant and B's note still shows B as the author. Only after A manually reloads the page does B disappear/anonymize.)

tohega: Another unexpected Behaviour: After we delete B's session, if we then add a new note in A's browser tab it also appears in B's tab still. The board keeps updating for B unless closing the tab/refresh. I think the place to fix this is in file [`server/src/api/boards_listen_ob_board.go` ](https://github.com/inovex/scrumlr.io/blob/main/server/src/api/boards_listen_on_board.go) line 89-107 for loop. This is out of scope but a follow up issue with a "someone left" notification?

server/src/sessions/service.go
```diff
@@ -361,6 +362,47 @@ func (service *BoardSessionService) IsParticipantBanned(ctx context.Context, boa
...
+	}
+
+	session, err := service.Get(ctx, boardID, userID)
+	if err != nil {
+		span.SetStatus(codes.Error, "failed to retrieve session")
+		span.RecordError(err)
+		return err
+	}
+
+	if session.Role == common.OwnerRole {
```
tohega: As the board owner, deleting your own sessions returns `400 Bad Request` right now. I think `403 ForbiddenError `is more suited, also need to update the swagger response list + the tests that assert 400.

server/src/sessions/api.go
```diff
@@ -219,6 +225,52 @@ func (api *API) UpdateBoardSessions(w http.ResponseWriter, r *http.Request) {
 	render.Respond(w, r, updatedSessions)
 }
 
+// Delete a sessions for a board
+//
+//	@Summary		Delete a sessions for a board
```
tohega: This description reads confusing to me on our Swagger UI, maybe something like "Deletes a participant from a board?"

server/src/sessions/service.go
tohega, line 288: Leaving a board always logs errors and the "people connected" number never goes back down.

How to recreate: delete a session, then close that deleted users's browser tab. The backend imediately logs this:

`ERROR  unable to disconnect from board session   error: "sql: no rows in result set"
WARN   failed to disconnected session            err:   "sql: no rows in result set"
ERROR  unable to disconnect from board session   error: "sql: no rows in result set"
WARN   failed to disconnected session            err:   "sql: no rows in result set"`

I think its because  when a tab closes, the server's normal cleanup is to go find that person's row in the participants table and mark them "not connected anymore". But we just deleted that row so there's nothing to find and it fails. never reaching the `connectedSessions.Add(ctx, -1)` step to decrement the connected sessions. 

I suggest that the cleanup step should not treat a missing row as an error. If the session is already gone, there's nothing to disconnect and nothing went wrong (it should quietly move on). Also the delete itsself should hanlde the connected count, since at that point it already knows whether the person was connected.

server/src/sessions/service.go
tohega, line 393: Here is where I suggest editing the connection count instead of line 288

tohega (changes requested):
Nice, it all works as described, however I noticed some unexpected behaviour that I documented here regarding 2 different users during a live session. To be fair they are mostly only edge cases.

server/src/sessions/service.go
Planlos5000: Yes. I was debating if it should be a `403` or a `400`. I will change it to a `403`

tohega (approved):
Nice, looks good now. All thats left is for the frontend to listen to the boradcast.

PR #6416, opened 2026-07-29
feat: Notify importers when notes with unavailable authors are removed

### LINKED ISSUE #6289: Add removed notes toast to board import
When a board is imported and one or more participants deleted their accounts before the import, these notes are not imported again. To show this to the importing person, we need a toast showing this information after the import.

This feature builds upon https://github.com/inovex/scrumlr.io/pull/6130

jmoehler:
## Description
When importing a board, notes whose authors can no longer be resolved are removed by the backend. This change informs the person importing the board about those removed notes with a one-time toast after they are redirected to the newly imported board.
## Changelog
- Added `importWarnings` to the board import response type.
- Updated the import API client to return the complete import response instead of only the board id.
- Persisted import warning data in session storage before redirecting to the imported board.
- Added a one-time toast on the board page when notes were removed because their authors could not be resolved.
- Added English, German, and French translations for the warning, including singular and plural forms.
- Added tests for the import API response and board-page warning behavior, including malformed session-storage data.

## Checklist

## Visual Changes
image

image


src/api/__tests__/board.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+describe("BoardAPI", () => {
+  const payload = {
+    board: {
+      name: "Imported Board",
+      accessPolicy: "PUBLIC",
+    },
+    columns: [],
+    notes: [],
+    participants: [] as unknown as BoardImportData["participants"],
```
Schwehn42: this made me skeptical -- the `participants` property of `BoardImportData` is defined as

```ts
participants: ParticipantWithUser;
```

shouldn't it be
```ts
participants: ParticipantWithUser[];
```

?

src/api/__tests__/board.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    voting: null as unknown as BoardImportData["voting"],
```
Schwehn42: similarly for `voting`, which should actually be called `votings`. and is also not nullable atm but should be

Schwehn42 (commented):
I got an idea how to avoid the session storage shenanigans altogether:

- `board/importBoard` thunk:
  - return `ImportBoardResponse`
  - remove `globalThis.location.assign` redirect, as this is a hard redirect making the site reload
-  `ImportBoard.tsx`
  - ```tsx
    dispatch(importBoard(updatedData))
    .unwrap()
    .then(response=>{
      if (response.containsWarning) {// this is pseudo code
        Toast.info({title: "warning text"});
        navigate(`/boards/${response.boardId}`); // use `useNavigate` hook instead of hard href redirect
      }
    });
```

that way we have a synchronous way to check and notify the user

src/api/__tests__/board.test.ts
jmoehler: adjusted

src/api/__tests__/board.test.ts
jmoehler: adjusted

Schwehn42 (approved):
lgtm 👍

PR #6417, opened 2026-07-29
refactor: typescript sonarcube maintainability

tohega:
## Description

fixes #6091 

Adresses the maintainability high severity issues for typescript files in sonarcube

## Changelog

- Refactor complex if statements to a switch case in the boards thunk file
- remove unnecessary eslint-disable/ts-ignore comments in SettingsCarousel.tsx
- Specificed eslint-disable-next-line for p element behaving like a button for the LoginBoard.tsx
- Sonarcube flagged the hash.ts utils file wrong, kept the same logic just used Math.imul to avoid using trunctuating with `| 0`
- Refactor complex if statement  for multiline detection in CustomToast.tsx

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have written and understand every part of this contribution myself - if AI tools were used, I have thoroughly reviewed and verified all changes
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] The light- and dark-theme are both supported and tested
- [x] The design was implemented and is responsive for all devices and screen sizes
- [x] The application was tested in the most commonly used browsers (e.g. Chrome, Firefox, Safari)

PR #6422, opened 2026-07-30
fix: repair i18next key typing and resulting type errors

### LINKED ISSUE #6421: Fix broken i18next key typing in src/types/i18next.d.ts
### Whats wrong

[src/types/i18next.d.ts](https://github.com/inovex/scrumlr.io/blob/main/src/types/i18next.d.ts) is meant to give us autocomplete and type errors on `t()` keys, but it does not work. Right now `t()` accepts any string, so typos in translation keys compile fine.

Two things are broken in the file:

// line 9-10: JSON imports are values, so they need `typeof` to be used as types
export type TranslationCommon = translationDe & translationEn & translationFr;

// line 15-16: these are type aliases, so `typeof` is invalid here
translation: typeof TranslationCommon;

The errors dont surface becazse `skipLibCheck: true` in `tsconfig.json` skips type checking of the .d.ts files, including this one. Flipping it to false and the file reports 8 errors, with Typescript suggesting:

```
src/types/i18next.d.ts(9,33): error TS2749: 'translationDe' refers to a value, but is being used as a type here. Did you mean 'typeof translationDe'?
src/types/i18next.d.ts(15,27): error TS2693: 'TranslationCommon' only refers to a type, but is being used as a value here.
```

### What happens today

1. Any key is accepted. `i18n.t("this.key.absolutely.does.not.exist")` compiles with zero errors, so a typo ships silently and could show the raw key to users.
2. `TranslationCommon` and `TranslationTemplates` are exported but imported nowhere else, so they are currently dead exports.
3. Weird sonarqube issues

Once the fix is implemented there will be other files with typescript errors with keys and strict typing rejections.

tohega:
## Description

fixes #6421 

wait for #6417 before merging as this branch is based on it

The CustomTypeOptions augmentation had typeof on the wrong lines, so resources never resolved and t() accepted any string and returned any. Hidden by skipLibCheck. Fixing it surfaced 15 errors: typed the statically known keys via as const and satisfies, fixed a typo key in LegacyNewBoard that rendered raw in an aria-label, dropped a double translation in ConfirmationDialog, and added utils/i18n.ts for the keys that are only known at runtime (to silence the type errors).

## Changelog

- 

## Checklist


src/components/SettingsDialog/Components/SettingsCarousel.tsx
```diff
@@ -34,7 +35,10 @@ export const SettingsCarousel: FC<SettingsCarouselProps<string>> = ({carouselIte
       </button>
       <div className="settings-carousel__text">
         {label && <span className="settings-carousel__text-label">{label}</span>}
-        <span className="settings-carousel__text-value">{localizationPath !== undefined ? t(`${localizationPath}${currentValue}`) : currentValue}</span>
+        <span className="settings-carousel__text-value">
+          {/* the key is composed from the caller's path and the current item, so it can only be resolved at runtime */}
```
jmoehler: I think the use of `dynamicTranslationKey` is self explanatory.  Would prefer to remove the comment.

src/routes/LoginBoard/LoginBoard.tsx
```diff
@@ -31,7 +31,7 @@ const FEATURE_KEYS = [
   "LoginBoard.loginFeature1",
   "LoginBoard.loginFeature2",
   // ...
-];
+] as const;
```
jmoehler: maybe also already check here for TranslationKey consistency and not only at usage.

jmoehler (approved):
All in all lgtm! Two small comments, but mostly optional.

src/components/SettingsDialog/Components/SettingsCarousel.tsx
jmoehler: resolved

src/routes/LoginBoard/LoginBoard.tsx
jmoehler: adjusted

PR #6443, opened 2026-08-03
feat: update frontend scrumlr api

### LINKED ISSUE #6462: Implement scrumlr API for frontend
Currently the frontend does not implement the complete scrumlr api. For the upcoming event changes we need the complete api to be implemented in the frontend.

Planlos5000:
## Description

For the upcoming event refactoring we need the complete api from the backend present in the frontend. This MR adds the missing requests to the frontend api. It also adds types to the requests where they were missing.

## Changelog

- Add missing requests
- Add types to request methods
- Add url builder function
- Add tests for url builder function
- Update swagger page
## Checklist

### DIFF
--- a/src/api/board.ts
+++ b/src/api/board.ts
@@ -1,7 +1,7 @@
 import {Color} from "constants/colors";
-import {BoardImportData, CreateSessionAccessPolicy, EditBoardRequest, ImportBoardResponse} from "store/features/board/types";
+import {Board, BoardImportData, CreateSessionAccessPolicy, EditBoardRequest, ImportBoardResponse} from "store/features/board/types";
 import {BoardOverview} from "store/features/history/types";
-import {SERVER_HTTP_URL} from "../config";
+import {buildUrl} from "./index";
 
 export const BoardAPI = {
   /**
@@ -11,7 +11,8 @@ export const BoardAPI = {
    */
   getBoards: async (): Promise<BoardOverview[]> => {
     try {
-      const response = await fetch(`${SERVER_HTTP_URL}/boards`, {
+      const url = buildUrl(`./boards`);
+      const response = await fetch(url, {
         method: "GET",
         credentials: "include",
       });
@@ -25,6 +26,32 @@ export const BoardAPI = {
       throw new Error(`unable to get boards`, {cause: error});
     }
   },
+
+  /**
+   * Get a board with the given id
+   *
+   * @param id the id of the board
+   *
+   * @returns the board
+   */
+  getBoard: async (id: string): Promise<Board> => {
+    try {
+      const url = buildUrl(`./boards/${id}`);
+      const response = await fetch(url, {
+        method: "GET",
+        credentials: "include",
+      });
+
+      if (response.status === 200) {
+        return (await response.json()) as Board;
+      }
+
+      throw new Error(`request resulted in response status ${response.status}`);
+    } catch (error) {
+      throw new Error(`unable to get board`, {cause: error});
+    }
+  },
+
   /**
    * Creates a board with the specified parameters and returns the board id.
    *
@@ -40,9 +67,10 @@ export const BoardAPI = {
     description: string | undefined,
     accessPolicy: CreateSessionAccessPolicy,
     columns: {name: string; visible: boolean; color: Color}[]
-  ) => {
+  ): Promise<string> => {
     try {
-      const response = await fetch(`${SERVER_HTTP_URL}/boards`, {
+      const url = buildUrl(`./boards`);
+      const response = await fetch(url, {
         method: "POST",
         credentials: "include",
         body: JSON.stringify({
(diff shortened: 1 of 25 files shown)


src/api/board.ts
```diff
@@ -33,18 +80,27 @@ export const BoardAPI = {
...
-  importBoard: async (boardJson: BoardImportData) => {
+
+  /**
+   * Import a josn board
+   *
+   * @param boardJson board to import
+   *
+   * @returns imported board
+   */
+  importBoard: async (boardJson: BoardImportData): Promise<ImportBoardResponse> => {
```
jmoehler, line 103: The BoardImportData contains more data (participants) then the backend consumes. Still works but that is a drift of, we should fix.

src/api/board.ts
```diff
@@ -1,8 +1,54 @@
 import {Color} from "constants/colors";
-import {BoardImportData, CreateSessionAccessPolicy, EditBoardRequest, ImportBoardResponse} from "store/features/board/types";
-import {SERVER_HTTP_URL} from "../config";
+import {Board, BoardImportData, CreateSessionAccessPolicy, EditBoardRequest, ImportBoardResponse} from "store/features/board/types";
+import {buildUrl} from "./index";
 
 export const BoardAPI = {
+  /**
+   * Get all boards
```
jmoehler: should specify: Get all boards for a given user

src/api/board.ts
```diff
@@ -1,8 +1,54 @@
...
+   */
+  getBoards: async () => {
```
jmoehler: has no return type given, is that intentional?

src/api/column.ts
```diff
@@ -1,7 +1,84 @@
...
+   * @param boardId board id
+   * @param column column to create
+   *
+   * @returns created column
+   */
+  createColumn: async (boardId: string, column: {name: string; color: string; visible: boolean; index: number}): Promise<Column> => {
+    const url = buildUrl(`/boards/${boardId}/columns`);
+    try {
+      const response = await fetch(url, {
+        method: "POSt",
```
jmoehler: big T :)

src/api/reaction.ts
```diff
@@ -1,10 +1,71 @@
...
+    }
+  },
+
+  /**
+   * Get a reaction for a board
+   *
+   * @param boardId the board id
+   * @param reactionId the reaction id
+   *
+   * @returnsthe requested reaction
```
jmoehler: space missing :)

jmoehler (changes requested):

src/api/board.ts
Planlos5000: Yes, this was intentional, because the return type from the API is `[]BoardOverview` which is not implemented in the frontend.

src/api/board.ts
Planlos5000: I think there a re a few drifts between the types in the frontend and the types in the backend that we should fix. But those fixes may have some other consequences.

src/api/index.ts
```diff
@@ -29,3 +30,27 @@ export const API = {
...
   ...TemplateColumnsAPI,
 };
+
+interface QueryParameter {
+  key: string;
+  value: string | number | boolean | string[] | number[] | null | undefined;
+}
+
+export function buildUrl(path: string, queryParameter: QueryParameter[] = []): URL {
+  const url = new URL(path, SERVER_HTTP_URL);
```
jmoehler: This will probably cause a bug.

The URL constructor treats any path that starts with / as an absolute path from the domain root, not as a path relative to the base path.

Example:

```
console.log(new URL("/some/path", "https://test.example.org/api/")
// https://test.example.org/some/path
```

So if we pass /boards with a base URL that already contains /api, the /api segment is dropped.

Expected:
https://test.example.org/api/boards

Actual:
https://test.example.org/boards

jmoehler (changes requested):

src/api/index.ts
Planlos5000: I will write a test for this case and fix the function

src/api/index.ts
```diff
@@ -29,3 +31,29 @@ export const API = {
...
 };
+
+interface QueryParameter {
+  key: string;
+  value: string | number | boolean | string[] | number[] | null | undefined;
+}
+
+export function buildUrl(urlPath: string, queryParameter: QueryParameter[] = []): URL {
+  const url = new URL(SERVER_HTTP_URL);
+  url.pathname = path.join(url.pathname, urlPath);
```
jmoehler: using a node module in frontend is not a good idea. Also the path.join is made for file systems, not urls.

I recommend using something like:

```
const normalizedBase = SERVER_HTTP_URL.endsWith("/") ? SERVER_HTTP_URL : `${SERVER_HTTP_URL}/`;
const normalizedPath = urlPath.replace(/^\/+/, ""); 
```

this would first add a / to the base if not there and then remove all / from the path.

you could then put them together like with the new URL

```
const url = new URL(normalizedPath, normalizedBase);
```

jmoehler (changes requested):

jmoehler (approved):
lgtm now :)

Schwehn42 (approved):
also fine for me

PR #6451, opened 2026-08-05
feat: add environment variable to allow anonymous users to view their history

Schwehn42:
## Description

relevant for https://github.com/inovex/scrumlr.io/pull/6372

- adds a backend env configuration to override the default setting that anon users cannot view their history.
- default is `false`.

## Changelog

- add env var / flag `SCRUMLR_ALLOW_ANONYMOUS_HISTORY` 
- send to frontend via `ServerInfo`
- set defaults for server docker composes as well as PR k8 config
- documented in respective env var section

## Checklist

### DIFF
--- a/server/src/main.go
+++ b/server/src/main.go
@@ -410,7 +410,7 @@ func run(ctx *cli.Context) error {
 
 	apiInitializer := serviceinitialize.NewApiInitializer(basePath)
 	sessionApi := apiInitializer.InitializeSessionApi(sessionService)
-	userApi := apiInitializer.InitializeUserApi(userService, sessionService, ctx.Bool("allow-anonymous-board-creation"), ctx.Bool("allow-anonymous-custom-templates"), ctx.Bool("allow-anonymous-history"))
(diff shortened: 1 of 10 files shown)


server/src/users/api.go
```diff
@@ -33,6 +33,7 @@ type API struct {
 	sessions                      sessions.SessionService
 	allowAnonymousBoardCreation   bool
 	allowAnonymousCustomTemplates bool
+	allowAnonymousHistory         bool
```
Planlos5000: We don't need this flag in the user api. It is not used here

server/src/main.go
```diff
@@ -141,6 +141,13 @@ func main() {
 				Required: false,
 				Value:    false,
 			}),
+			altsrc.NewBoolFlag(&cli.BoolFlag{
+				Name:     "allow-anonymous-history",
+				EnvVars:  []string{"SCRUMLR_ALLOW_ANONYMOUS_HISTORY"},
+				Usage:    "allows anonymous clients to view their history",
+				Required: false,
+				Value:    true,
```
Planlos5000: I thought we want the value to be false by default?

Planlos5000 (approved):

PR #6464, opened 2026-08-07
feat: cancel voting (backend)

### LINKED ISSUE #6396: Cancel a voting
### Problem & Kontext:
Currently, it is not possible to cancel an ongoing voting. The user can only click "End voting phase", which closes the voting and automatically evaluates the results. If a voting was started by mistake, or if the context changed and the vote is no longer relevant, hosts need a way to discard/abort the session entirely without generating results or affecting statistics.

### User Story:
**As a** moderator,
**i want** to cancel an ongoing voting session without triggering an evaluation,
**so that** I can abort mistakenly started or invalid votings without saving or calculating results.

### Acceptance Criteria:
**Backend:**
- [x] Add the cancel state to the voting
- [x] Update the voting update endpoint so that a voting can be canceled
- [x] Ensure canceled votings do not trigger result evaluation logic or downstream notification triggers

**Frontend & UX:**
- [x] Update the UI modal to include a "Cancel voting"-button (secondary) below "End voting phase" (primary)
- [ ] Implement frontend layout according to the final design specs (Design: TBD)

mating02:
## Description
I implemented a new canceled state that uses the ABORT db enum to update the voting and set its state to canceled. For this, new functions were created to cancel the voting without triggering the evaluation of the voting.

## Changelog
- new canceled state in votings/voting_status.go
- extended the votings api with the new cancel function
- added cancel function to votings/service.go
- extended close function in votings/database.go to differentiate between the canceled target state and the closed target state
- extended api endpoint in api/votings.go  to also include the canceled state
- added unit tests and integration test to ensure correct functionality 

## Checklist

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/server/src/api/votings.go
+++ b/server/src/api/votings.go
@@ -110,7 +110,26 @@ func (s *Server) updateVoting(w http.ResponseWriter, r *http.Request) {
 		})
 	}
 
-	voting, err := s.votings.Close(ctx, id, board, affectedNotes)
+	var body votings.VotingCloseRequest
+
+	if r.ContentLength != 0 {
+		if err := render.Decode(r, &body); err != nil {
+			span.SetStatus(codes.Error, "unable to decode body")
+			span.RecordError(err)
+			common.Throw(w, r, common.BadRequestError(err))
+			return
+		}
+	}
+	body.ID = id
+	body.Board = board
+
+	var voting *votings.Voting
+	if body.Status == votings.Canceled {
+		voting, err = s.votings.Cancel(ctx, id, board, affectedNotes)
+	} else {
+		voting, err = s.votings.Close(ctx, id, board, affectedNotes)
+	}
+
 	if err != nil {
 		span.SetStatus(codes.Error, "failed to update voting")
 		span.RecordError(err)
--- a/server/src/api/votings.go
+++ b/server/src/api/votings.go
@@ -87,6 +87,16 @@ func (s *Server) updateVoting(w http.ResponseWriter, r *http.Request) {
 	board := ctx.Value(identifiers.BoardIdentifier).(uuid.UUID)
 	id := ctx.Value(identifiers.VotingIdentifier).(uuid.UUID)
 
+	var body votings.VotingCloseRequest
+	if err := render.Decode(r, &body); err != nil {
+		span.SetStatus(codes.Error, "unable to decode body")
+		span.RecordError(err)
+		common.Throw(w, r, common.BadRequestError(err))
+		return
+	}
+	body.ID = id
+	body.Board = board
+
 	notes, err := s.notes.GetAll(ctx, board)
 	if err != nil {
 		span.SetStatus(codes.Error, "failed to get notes")
@@ -110,25 +120,7 @@ func (s *Server) updateVoting(w http.ResponseWriter, r *http.Request) {
 		})
 	}
 
-	var body votings.VotingCloseRequest
-
-	if r.ContentLength != 0 {
-		if err := render.Decode(r, &body); err != nil {
-			span.SetStatus(codes.Error, "unable to decode body")
-			span.RecordError(err)
-			common.Throw(w, r, common.BadRequestError(err))
-			return
-		}
-	}
-	body.ID = id
-	body.Board = board
-
-	var voting *votings.Voting
-	if body.Status == votings.Canceled {
-		voting, err = s.votings.Cancel(ctx, id, board, affectedNotes)
-	} else {
-		voting, err = s.votings.Close(ctx, id, board, affectedNotes)
-	}
+	voting, err := s.votings.Update(ctx, id, board, affectedNotes, body.Status)
 
 	if err != nil {
 		span.SetStatus(codes.Error, "failed to update voting")
--- a/server/src/api/votings.go
+++ b/server/src/api/votings.go
@@ -87,7 +87,7 @@ func (s *Server) updateVoting(w http.ResponseWriter, r *http.Request) {
 	board := ctx.Value(identifiers.BoardIdentifier).(uuid.UUID)
 	id := ctx.Value(identifiers.VotingIdentifier).(uuid.UUID)
 
-	var body votings.VotingCloseRequest
+	var body votings.VotingUpdateRequest
 	if err := render.Decode(r, &body); err != nil {
 		span.SetStatus(codes.Error, "unable to decode body")
 		span.RecordError(err)
@@ -120,7 +120,7 @@ func (s *Server) updateVoting(w http.ResponseWriter, r *http.Request) {
 		})
 	}
 
-	voting, err := s.votings.Update(ctx, id, board, affectedNotes, body.Status)
+	voting, err := s.votings.Update(ctx, id, board, body.Status, affectedNotes)
 
 	if err != nil {
 		span.SetStatus(codes.Error, "failed to update voting")
--- a/server/src/api/votings_test.go
+++ b/server/src/api/votings_test.go
@@ -103,7 +103,7 @@ func (suite *VotingTestSuite) TestCloseVoting() {
 
 			notesMock.EXPECT().GetAll(mock.Anything, boardId).Return([]*notes.Note{}, nil)
 
-			votingMock.EXPECT().Update(mock.Anything, votingId, boardId, []votings.Note(nil), votings.Closed).
+			votingMock.EXPECT().Update(mock.Anything, votingId, boardId, votings.Closed, []votings.Note(nil)).
 				Return(&votings.Voting{Status: votings.Closed}, tt.err)
 
 			s.updateVoting(rr, req.Request())
@@ -140,7 +140,7 @@ func (suite *VotingTestSuite) TestAbortVoting() {
 
 			notesMock.EXPECT().GetAll(mock.Anything, boardId).Return([]*notes.Note{}, nil)
 
-			votingMock.EXPECT().Update(mock.Anything, votingId, boardId, []votings.Note(nil), votings.Aborted).
+			votingMock.EXPECT().Update(mock.Anything, votingId, boardId, votings.Aborted, []votings.Note(nil)).
 				Return(&votings.Voting{Status: votings.Aborted}, tt.err)
 
 			rr := httptest.NewRecorder()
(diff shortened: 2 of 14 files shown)


server/src/api/votings.go
```diff
@@ -110,7 +110,26 @@ func (s *Server) updateVoting(w http.ResponseWriter, r *http.Request) {
...
-	voting, err := s.votings.Close(ctx, id, board, affectedNotes)
+	var body votings.VotingCloseRequest
+
+	if r.ContentLength != 0 {
+		if err := render.Decode(r, &body); err != nil {
+			span.SetStatus(codes.Error, "unable to decode body")
+			span.RecordError(err)
+			common.Throw(w, r, common.BadRequestError(err))
+			return
+		}
```
Planlos5000: The decoding should happen at the top of the method so that we could fail fast. Currently the method fetches notes and transforms them and then we check if there is a decoding error.

server/src/api/votings.go
```diff
@@ -110,7 +110,26 @@ func (s *Server) updateVoting(w http.ResponseWriter, r *http.Request) {
 		})
 	}
 
-	voting, err := s.votings.Close(ctx, id, board, affectedNotes)
+	var body votings.VotingCloseRequest
+
+	if r.ContentLength != 0 {
```
Planlos5000: Why do we need this check?

server/src/api/votings_test.go
```diff
@@ -114,6 +114,43 @@ func (suite *VotingTestSuite) TestCloseVoting() {
...
+			//s.basePath = "/"
+			votingMock := votings.NewMockVotingService(suite.T())
+			notesMock := notes.NewMockNotesService(suite.T())
+
+			boardId, _ := uuid.NewRandom()
+			votingId, _ := uuid.NewRandom()
+			s.votings = votingMock
+			s.notes = notesMock
+
+			req := technical_helper.NewTestRequestBuilder("PUT", "/", strings.NewReader(`{"status": "ABORTED"}`))
```
Planlos5000: can we here use the `VotingStatus`?

server/src/votings/database.go
```diff
@@ -31,25 +31,38 @@ func (d *DB) Create(ctx context.Context, insert DatabaseVotingInsert) (DatabaseV
 
 func (d *DB) Close(ctx context.Context, update DatabaseVotingUpdate) (DatabaseVoting, error) {
```
Planlos5000: I would argue that it would be now better to rename this function to `Update`

server/src/votings/api.go
```diff
@@ -15,6 +15,7 @@ type VotingService interface {
 	AddVote(ctx context.Context, req VoteRequest) (*Vote, error)
 	RemoveVote(ctx context.Context, req VoteRequest) error
 	Close(ctx context.Context, id uuid.UUID, board uuid.UUID, affectedNotes []Note) (*Voting, error)
+	Cancel(ctx context.Context, id uuid.UUID, board uuid.UUID, affectedNotes []Note) (*Voting, error)
```
Planlos5000: Could we combine these functions to one `Update` function? This would have the benefit, that the api does not have to check which method needs to be called and therefor logic is removed from the api to the service layer where the logic should happen

server/src/votings/service.go
```diff
@@ -326,6 +326,40 @@ func (service *Service) createdVoting(ctx context.Context, board uuid.UUID, voti
 	}
 }
 
+// Cancel marks an open voting as canceled without running evaluation
+func (service *Service) Cancel(ctx context.Context, id uuid.UUID, board uuid.UUID, affectedNotes []Note) (*Voting, error) {
```
Planlos5000: Put this function before the private methods

server/src/votings/voting_status.go
```diff
@@ -5,13 +5,16 @@ import (
...
 
-// VotingStatus is the state of a voting session and can be one of open, aborted or closed.
+// VotingStatus is the state of a voting session and can be one of open, canceled or closed.
 type VotingStatus string
 
 const (
 	// Open is the state for an open voting session, meaning that votes are allowed.
 	Open VotingStatus = "OPEN"
 
+	// Canceled represents a canceled voting session (the DB enum uses ABORTED)
```
Planlos5000: If the database uses abort, i think we should also use abort throughout the backend

server/src/api/votings.go
mating02: because otherwise an error is thrown in the test due to the content length of closing requests being 0

server/src/votings/database.go
mating02: good point

server/src/api/votings.go
mating02: good point

server/src/api/votings_test.go
mating02: i'll try

server/src/votings/api.go
mating02: good idea but don't you think the seperation between close and cancel will be less obvious if we'd do that?

server/src/api/votings.go
Planlos5000: I would argue, that the body should never be empty with this change. The user should always specify what action (Abort/cancel, Close) should be applied in the body

wischoepke:
I'd like to bring up a general point that caught my eye again in this PR. The OpenTelemetry stuff tends to clutter up the method quite a bit. Could you check if it makes sense to extract that into separate methods for your PR?

server/src/votings/dto.go
```diff
@@ -49,8 +49,9 @@ type VotingCreateRequest struct {
 
 // VotingCloseRequest represents the request to update a voting session.
 type VotingCloseRequest struct {
```
Planlos5000: I think we should rename this struct now to `VotingUpdateRequest` or something like this

server/src/votings/api.go
```diff
@@ -14,7 +14,7 @@ type VotingService interface {
 	GetVotes(ctx context.Context, board uuid.UUID, f VoteFilter) ([]*Vote, error)
 	AddVote(ctx context.Context, req VoteRequest) (*Vote, error)
 	RemoveVote(ctx context.Context, req VoteRequest) error
-	Close(ctx context.Context, id uuid.UUID, board uuid.UUID, affectedNotes []Note) (*Voting, error)
+	Update(ctx context.Context, id uuid.UUID, board uuid.UUID, affectedNotes []Note, votingStatus VotingStatus) (*Voting, error)
```
Planlos5000: What do you think of changing the signature so that first the status is requested and then the notes?

server/src/votings/service.go
```diff
@@ -263,20 +263,42 @@ func (service *Service) RemoveVote(ctx context.Context, body VoteRequest) error
...
+	})
+
+	if err != nil {
+		span.SetStatus(codes.Error, "failed to send voting created")
+		span.RecordError(err)
+		log.Errorw("unable to send voting created", "err", err)
+	}
+}
+
+func (service *Service) Update(ctx context.Context, id uuid.UUID, board uuid.UUID, affectedNotes []Note, votingStatus VotingStatus) (*Voting, error) {
```
Planlos5000: Public functions should be before private functions

server/src/votings/voting_status.go
```diff
@@ -5,13 +5,16 @@ import (
 	"errors"
 )
 
-// VotingStatus is the state of a voting session and can be one of open, aborted or closed.
+// VotingStatus is the state of a voting session and can be one of open, canceled or closed.
```
Planlos5000: Also change the comment to aborted

Planlos5000 (changes requested):

Planlos5000:
The e2e tests need an update because the voting update now requires the status

server/src/api/votings_test.go
```diff
@@ -94,26 +94,64 @@ func (suite *VotingTestSuite) TestCloseVoting() {
...
 			req.Req = logger.InitTestLoggerRequest(req.Request())
 			req.AddToContext(identifiers.BoardIdentifier, boardId).
 				AddToContext(identifiers.VotingIdentifier, votingId)
 			rr := httptest.NewRecorder()
 
 			notesMock.EXPECT().GetAll(mock.Anything, boardId).Return([]*notes.Note{}, nil)
 
-			votingMock.EXPECT().Close(mock.Anything, votingId, boardId, []votings.Note(nil)).
+			votingMock.EXPECT().Update(mock.Anything, votingId, boardId, votings.Closed, []votings.Note(nil)).
 				Return(&votings.Voting{Status: votings.Closed}, tt.err)
```
Planlos5000: Use the `.Once()` function here if you want to check, that the mock is only called one time. Then you should be able to remove the `votingMock.AssertNumberOfCalls(suite.T(), "Update", 1)` call.

We also don't need the `AssertExpectations` call since this is done for us by the default cleanup function of the mock

server/src/api/votings_test.go
```diff
@@ -94,26 +94,64 @@ func (suite *VotingTestSuite) TestCloseVoting() {
...
+			notesMock.EXPECT().GetAll(mock.Anything, boardId).Return([]*notes.Note{}, nil)
+
+			votingMock.EXPECT().Update(mock.Anything, votingId, boardId, votings.Aborted, []votings.Note(nil)).
+				Return(&votings.Voting{Status: votings.Aborted}, tt.err)
+
+			rr := httptest.NewRecorder()
+			s.updateVoting(rr, req.Request())
+			suite.Equal(tt.expectedCode, rr.Result().StatusCode)
+			votingMock.AssertExpectations(suite.T())
+			votingMock.AssertNumberOfCalls(suite.T(), "Update", 1)
```
Planlos5000: Same as above: Use the `.Once()` method

server/src/votings/voting_status.go
```diff
@@ -12,6 +12,9 @@ const (
 	// Open is the state for an open voting session, meaning that votes are allowed.
 	Open VotingStatus = "OPEN"
 
+	// Aborted represents an aborted voting session (the DB enum uses ABORTED)
```
Planlos5000: The comment for the db is here unnecessary

mating02: but the other states have comments too despite being self explanatory

Planlos5000 (approved):
LGTM

PR #6465, opened 2026-08-07
feat: cancel voting (frontend)

### LINKED ISSUE #6396: Cancel a voting

mating02:
## Description
I implemented a new button right beneath the stop voting button that uses the backend cancel feature to stop the voting without evaluation. 

## Changelog
- added cancelVoting feature in thunks.ts
- extended changeVotingStatus function in votings.ts
- added cancel Voting button in VotingDialog.tsx

## Checklist

## Visual Changes
Screenshot 2026-08-10 at 14 16 12
Screenshot 2026-08-10 at 14 17 17

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/src/api/votings.ts
+++ b/src/api/votings.ts
@@ -1,5 +1,5 @@
-import {CreateVotingRequest, Voting} from "store/features/votings/types";
-import {buildUrl} from "./index";
+import { CreateVotingRequest, Voting } from "store/features/votings/types";
+import { buildUrl } from "./index";
 
 export const VotingAPI = {
   /**
@@ -23,7 +23,7 @@ export const VotingAPI = {
 
       throw new Error(`get votings request resulted in response with status ${response.status}`);
     } catch (error) {
-      throw new Error(`unable to get votings`, {cause: error});
+      throw new Error(`unable to get votings`, { cause: error });
     }
   },
 
@@ -49,7 +49,7 @@ export const VotingAPI = {
 
       throw new Error(`get voting request resulted in response with status ${response.status}`);
     } catch (error) {
-      throw new Error(`unable to get voting`, {cause: error});
+      throw new Error(`unable to get voting`, { cause: error });
     }
   },
 
@@ -75,7 +75,7 @@ export const VotingAPI = {
 
       throw new Error(`create voting request resulted in response with status ${response.status}`);
     } catch (error) {
-      throw new Error(`unable to create voting`, {cause: error});
+      throw new Error(`unable to create voting`, { cause: error });
     }
   },
 
@@ -87,13 +87,18 @@ export const VotingAPI = {
    *
    * @returns updated voting
    */
-  changeVotingStatus: async (boardId: string, votingId: string): Promise<Voting> => {
+  changeVotingStatus: async (board: string, voting: string, status?: string): Promise<Voting> => {
     try {
-      const url = buildUrl(`./boards/${boardId}/votings/${votingId}`);
+      const url = buildUrl(`./boards/${board}/votings/${voting}`);
       const response = await fetch(url, {
         method: "PUT",
         credentials: "include",
       });
+
+      if (typeof status !== "undefined") {
+        response.body = JSON.stringify({ status });
+        response.headers = { "Content-Type": "application/json" };
+      }
 
       if (response.status === 200) {
         return (await response.json()) as Voting;
@@ -101,7 +106,7 @@ export const VotingAPI = {
 
       throw new Error(`change voting status request resulted in response with status ${response.status}`);
     } catch (error) {
-      throw new Error(`unable to change voting status`, {cause: error});
+      throw new Error(`unable to change voting status`, { cause: error });
     }
   },
 };
--- a/src/api/votings.ts
+++ b/src/api/votings.ts
@@ -1,4 +1,4 @@
-import { CreateVotingRequest, Voting } from "store/features/votings/types";
+import { CreateVotingRequest, VotingStatus } from "store/features/votings/types";
 import { buildUrl } from "./index";
 
 export const VotingAPI = {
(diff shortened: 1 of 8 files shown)


src/api/votings.ts
```diff
@@ -27,12 +27,19 @@ export const VotingAPI = {
...
         credentials: "include",
-      });
+      };
+
+      if (typeof status !== "undefined") {
+        options.body = JSON.stringify({status});
+        options.headers = {"Content-Type": "application/json"};
+      }
+
+      const response = await fetch(`${SERVER_HTTP_URL}/boards/${board}/votings/${voting}`, options);
```
Planlos5000: I think the status should always be set and should not be able to undefined or null. Also use the `VotingStatus` type

src/components/VotingDialog/VotingDialog.tsx
```diff
@@ -46,12 +46,23 @@ export const VotingDialog = () => {
     navigate("..");
   };
 
+  const abortVoting = () => {
+    dispatch(cancelVoting(voting!));
+    navigate("..");
+  }
```
Planlos5000: We should be consistent with the naming. Is it now abort or cancel?

src/components/VotingDialog/VotingDialog.tsx
```diff
@@ -46,12 +46,23 @@ export const VotingDialog = () => {
...
       {voting ? (
-        <button className="voting-dialog__start-button" data-testid="voting-dialog__stop-button" onClick={() => stopVoting()}>
-          <label>{t("VoteConfigurationButton.stopVoting")}</label>
-        </button>
+        <>
+          <button className="voting-dialog__start-button" data-testid="voting-dialog__stop-button" onClick={() => stopVoting()}>
+            <label>{t("VoteConfigurationButton.stopVoting")}</label>
+          </button>
+
+          <button className="voting-dialog__start-button voting-dialog__cancel-button" data-testid="voting-dialog__cancel-button" onClick={() => abort_voting()}>
```
Schwehn42: should be "abort" instead of "cancel"

src/components/VotingDialog/VotingDialog.tsx
```diff
@@ -46,12 +46,23 @@ export const VotingDialog = () => {
     navigate("..");
   };
 
+  const abort_voting = () => {
```
Schwehn42: snake case‽ this ain't python

src/api/votings.ts
```diff
@@ -23,24 +23,28 @@ export const VotingAPI = {
 
       throw new Error(`create voting request resulted in response with status ${response.status}`);
     } catch (error) {
-      throw new Error(`unable to create voting`, {cause: error});
+      throw new Error(`unable to create voting`, { cause: error });
     }
   },
 
-  changeVotingStatus: async (board: string, voting: string) => {
+  changeVotingStatus: async (board: string, voting: string, status?: VotingStatus) => {
```
Schwehn42: `status` can be non-nullable

src/api/votings.ts
```diff
@@ -23,24 +23,28 @@ export const VotingAPI = {
...
     try {
-      const response = await fetch(`${SERVER_HTTP_URL}/boards/${board}/votings/${voting}`, {
+      const options: RequestInit = {
```
Schwehn42: I'd prefer the options inline inside the fetch rather than an extra object, since we do it that way in all other places (not to say this is bad, I just want it to be the same if possible)

src/api/votings.ts
```diff
@@ -23,24 +23,28 @@ export const VotingAPI = {
...
         method: "PUT",
         credentials: "include",
-      });
+        body: JSON.stringify({ status: status }),
```
Schwehn42: properties with the same name and value can be shortened like this: `{status}`

src/store/features/votings/thunks.ts
```diff
@@ -1,26 +1,37 @@
...
   );
 });
 
-export const closeVoting = createAsyncThunk<void, string, {state: ApplicationState}>("votings/closeVoting", async (payload, {dispatch, getState}) => {
+export const closeVoting = createAsyncThunk<void, string, { state: ApplicationState }>("votings/closeVoting", async (payload, { dispatch, getState }) => {
   const boardId = getState().board.data!.id;
 
   await retryable(
-    () => API.changeVotingStatus(boardId, payload),
+    () => API.changeVotingStatus(boardId, payload, "CLOSED"),
```
Schwehn42, line 21: should be "ABORT(ED)" here also

Schwehn42 (changes requested):

src/components/VotingDialog/VotingDialog.tsx
mating02: I can't also name the arrow function abortVoting though, so cuz of naming conflicts I used snake case

src/store/features/votings/thunks.ts
mating02: No because then the whole voting is aborted instead of closed although it should only be closed to end the vote successfully. ABORTED status is set in the abortVoting function

Schwehn42: ohh you're right my bad 🙏 read it as cancel

src/components/VotingDialog/VotingDialog.tsx
Schwehn42: you could rename it to something like `abortVotingCallback` or rename the import above to `abortVoting as abortVotingDispatch` or something like that

src/api/votings.ts
```diff
@@ -87,21 +87,23 @@ export const VotingAPI = {
...
-  changeVotingStatus: async (boardId: string, votingId: string): Promise<Voting> => {
+  changeVotingStatus: async (board: string, voting: string, status: VotingStatus): Promise<Voting> => {
     try {
-      const url = buildUrl(`./boards/${boardId}/votings/${votingId}`);
+      const url = buildUrl(`./boards/${board}/votings/${voting}`);
       const response = await fetch(url, {
         method: "PUT",
         credentials: "include",
+        body: JSON.stringify({ status }),
+        headers: { "Content-Type": "application/json" },
```
Planlos5000: Why do we here need the headers field?

Schwehn42 (approved):
lgtm

PR #6471, opened 2026-08-07
feat: upgrade user

### LINKED ISSUE #6470: Add posibilty to upgrade a user
Currently you cannot "upgrade" an anonymous user a user that is logged in with an oidc provider.
The  feature should allow an anonymous user to login with an oidc provider without loosing the sessions of the anonymous user.

Planlos5000:
## Description

This allows an anonymous user to upgrade the user to an oidc user. This will either upgrade the anonymous account if the id from the oidc provider did not exist or it will merge the anonymous account into the exisiting oidc account.
## Changelog

- Add user upgrade methods
- Merge anonymous user if the oidc account already exists
- Add tests for user upgrade methods
- Change login to upgrade an existing anonymous user
- Use mockery to generate mocks for auth package
## Checklist

### DIFF
--- a/server/src/api/login.go
+++ b/server/src/api/login.go
@@ -123,6 +123,7 @@ func (s *Server) beginAuthProviderVerification(w http.ResponseWriter, r *http.Re
 //	@Description	Verify the auth provider call and create or update a user. Redirect to the page provider with the state
 //	@Tags			auth
 //	@Accept			json
+//	@Param			Cookie	header	string					false	"jwt token to for upgrading an existing user"
 //	@Param			user	body	AnonymousSignUpRequest	true	"user to create"
 //	@Produce		json
 //	@Header			303	{string}	Cookie		"jwt token to sign in"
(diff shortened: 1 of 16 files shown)


Planlos5000:
Here the frontend needs to be designed

jmoehler:
sonarqube is not happy yet :)

https://sonarcloud.io/summary/new_code?id=scrumlr_scrumlr&pullRequest=6471

server/src/users/otel.go
```diff
@@ -22,7 +22,7 @@ var anonymousUserCreatedCounter, _ = meter.Int64Counter(
 )
 
 var appleUserCreatedCounter, _ = meter.Int64Counter(
-	"scrumlr.users.aplle.created.counter",
+	"scrumlr.users.appel.created.counter",
```
jmoehler, line 25: should probably be apple not appel

server/src/users/database.go
```diff
@@ -71,6 +71,130 @@ func (db *DB) UpdateUser(ctx context.Context, update DatabaseUserUpdate) (Databa
...
+			Exec(ctx)
+		if err != nil {
+			return err
+		}
+
+		_, err = tx.NewUpdate().
+			Table("board_sessions").
+			Set("\"user\" = ?", mergeInto).
+			Where("\"user\" = ?", merge).
+			Where("board NOT IN (?)",
```
jmoehler, line 182: If both users already have a board_sessions row for the same board, the old user’s row is left behind. Is that intentional? That could create problems right?

server/src/users/database.go
```diff
@@ -71,6 +71,130 @@ func (db *DB) UpdateUser(ctx context.Context, update DatabaseUserUpdate) (Databa
...
+			user.ID, id, name, avatarUrl,
+		).Exec(ctx)
+
+		return err
+	})
+
+	return user, err
+}
+
+func (db *DB) MergeUser(ctx context.Context, mergeInto uuid.UUID, merge uuid.UUID) (DatabaseUser, error) {
```
jmoehler, line 129: what about board_session_requests? they are not merged currently

server/src/api/login.go
```diff
@@ -162,12 +163,34 @@ func (s *Server) verifyAuthProviderCallback(w http.ResponseWriter, r *http.Reque
...
 
 	var internalUser *users.User
-	internalUser, err = s.users.Create(ctx, userInfo.Ident, userInfo.Name, userInfo.AvatarURL, provider)
-	if err != nil {
-		otel.RecordErrorSpan(span, err, new("failed to create user"))
-		w.WriteHeader(http.StatusInternalServerError)
-		log.Errorw("could not create user", "err", err)
-		return
+
+	id, err := s.auth.ExtractUserId(r)
```
jmoehler, line 167: If a browser still has a valid app JWT for a user who was deleted, the new login callback now reads the user ID from that cookie and tries to load the local user first. When that lookup returns not found, the callback returns HTTP 500 instead of falling back to the previous provider-based create or reuse flow. This leaves the user in a non-recoverable login state unless they clear the cookie (or wait for it to expire).

server/src/users/otel.go
```diff
@@ -53,7 +53,49 @@ var microsoftUserCreatedCounter, _ = meter.Int64Counter(
...
+)
+
+var userUpgradedCounter, _ = meter.Int64Counter(
+	"scrumlr.users.upgraded.counter",
+	metric.WithDescription("Number of upgraded users"),
+	metric.WithUnit("users"),
+)
+
+var appleUserUpgradedCounter, _ = meter.Int64Counter(
+	"scrumlr.users.appel.upgraded.counter",
```
jmoehler, line 67: apple not appel

server/src/auth/auth.go
```diff
@@ -270,6 +271,35 @@ func (a *AuthConfiguration) ExtractUserInformation(accountType common.AccountTyp
 	return result, nil
 }
 
+func (a *AuthConfiguration) ExtractUserId(r *http.Request) (uuid.UUID, error) {
+	var token jwt.Token
+	var err error
+	if a.unsafeAuth != nil {
+		token, err = jwtauth.VerifyRequest(a.unsafeAuth, r, jwtauth.TokenFromCookie)
+		if err != nil {
+			return uuid.UUID{}, err
```
jmoehler, line 280: In ExtractUserId, when verification with unsafeAuth fails, the code returns immediately, so it never reaches the auth verification path.

In Verifier, the logic is different: it tries unsafeAuth first, and if that fails, it falls back to auth.

jmoehler (changes requested):
still needs frontend part, but you mentioned that already.

PR #6483, opened 2026-08-10
feat: upgrade urfavre cli to v3

Planlos5000:
## Description

Upgrade the urfavre cli package to v3.
## Changelog

- Upgrade package
- Adjust tests
## Checklist


server/src/go.mod
mating02, line 1: Why are the indirect modules removed here? What was there purpose?

Planlos5000: I updated the urfavre/cli package and removed the old version. I would assume, since i did not change the go.mod file by hand, that those packages are not needed anymore by urfavre/cli v3

server/src/main.go
```diff
@@ -31,337 +33,467 @@ import (
...
-			}),
-			altsrc.NewStringFlag(&cli.StringFlag{
-				Name:     "auth-google-client-id",
-				EnvVars:  []string{"SCRUMLR_AUTH_GOOGLE_CLIENT_ID"},
+			},
+			&cli.StringFlag{
+				Name: "auth-google-client-id",
+				Sources: cli.NewValueSourceChain(
+					cli.EnvVar("SCRUMLR_AUTH_GOOGLE_CLIENT_ID"),
+					toml.TOML("auth-callback-host", altsrc.NewStringPtrSourcer(&tomlconfigFile)),
```
mating02: looks like a copy and paste error, why auth-callback-host here? I think it should be auth-google-client-id

server/src/main.go
```diff
@@ -31,337 +33,467 @@ import (
...
 				Required: false,
 				Value:    "INFO",
-			}),
-			altsrc.NewBoolFlag(&cli.BoolFlag{
-				Name:  "disable-check-origin",
+			},
+			&cli.BoolFlag{
+				Name: "disable-check-origin",
+				Sources: cli.NewValueSourceChain(
+					toml.TOML("disable-check-origin", altsrc.NewStringPtrSourcer(&tomlconfigFile)),
```
mating02, line 420: why no environment variable support for disable-check-origin here?

server/src/main.go
```diff
@@ -31,337 +33,467 @@ import (
...
 				Usage:    "allows anonymous clients to view their history",
 				Required: false,
 				Value:    false,
-			}),
-			altsrc.NewBoolFlag(&cli.BoolFlag{
-				Name:     "auth-enable-experimental-file-system-store",
-				EnvVars:  []string{"SCRUMLR_ENABLE_EXPERIMENTAL_AUTH_FILE_SYSTEM_STORE"},
+			},
+			&cli.BoolFlag{
+				Name: "auth-enable-experimental-file-system-store",
```
mating02, line 216: this name here seems to be wrong. in the toml file, it is auth-experimental-file-system-store

server/src/main.go
```diff
@@ -31,337 +33,467 @@ import (
...
+				Sources: cli.NewValueSourceChain(
+					cli.EnvVar("SCRUMLR_ENABLE_EXPERIMENTAL_AUTH_FILE_SYSTEM_STORE"),
+					toml.TOML("auth-enable-experimental-file-system-store", altsrc.NewStringPtrSourcer(&tomlconfigFile)),
```
mating02, line 219: also here, i think the enable in between needs to be removed or the name in the toml needs to be changed

server/config_example.toml
```diff
@@ -94,6 +109,15 @@ auth-oidc-user-ident-scope = ""
...
+session-secret = ""
+
+# Set the OTel grpc endpoint for sending logs, traces and metrics
+otel-grpc = ""
+
+# Set the OTel http endpoint for sending logs, traces and metrics
+otel-http = ""
+
 # Enable or disable verbose logging.
 verbose = true
```
mating02: where is this flag used?

mating02 (changes requested):

server/src/main.go
Planlos5000: Because the user should set it explicit, because it has some security implications

server/src/main.go
Planlos5000: fixed that

server/src/main.go
Planlos5000: I have changed the name in the toml file since the env var is also called `SCRUMLR_ENABLE_EXPERIMENTAL_AUTH_FILE_SYSTEM_STORE`

server/config_example.toml
Planlos5000: Verbose flag was replaced with the log level. It was deprecated with version [4.0.0](https://github.com/inovex/scrumlr.io/releases/tag/v4.0.0) and was then removed with version [4.5.0](https://github.com/inovex/scrumlr.io/releases/tag/v4.5.0). This was just never removed from the toml file

mating02 (approved):

PR #6490, opened 2026-08-11
fix: avoid avatar update when opening profile

### LINKED ISSUE #5577: HTTP request when opening profile
### The problem

When clicking on the profile icon we instantly send HTTP request that triggers a PARTICIPANT_UPDATED event even though nothing happened

https://github.com/user-attachments/assets/03f4dfe7-ff68-4d7a-bef7-999465a70746

The issue seem to be a frontend problem. When removing line 42 in AvatarSettings.tsx
```  useEffect(() => {
    dispatch(
      editSelf({
        auth: {...self, avatar: properties},
        applyOptimistically: true,
      })
    );
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [properties]);``` 
the issue is no longer present but i am also not able to change the profile avatar anymore

### Browser

firefox

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by almacbe**
## Description

Fixes #5577

`AvatarSettings` treated its initial effect as a user change and persisted the current avatar whenever the profile opened. It now dispatches `editSelf` only when the local avatar is semantically different from the persisted avatar.

This keeps the intentional migration for older avatars without `backgroundColor`, because the completed avatar still differs from the stored value and is persisted once. Regression tests cover initial mounting, real and duplicate shuffle results, optimistic updates, legacy avatars, and missing avatars.

## Changelog

- Skip avatar updates when the local and persisted avatar values are deeply equal.
- Keep optimistic persistence for real avatar changes and legacy avatar migration.
- Add behavioral coverage around the `API.editUser` boundary.

## Checklist

## (Optional) Visual Changes

None. This change does not alter the UI.

### CODE REVIEW

File: src/components/SettingsDialog/Components/__tests__/AvatarSettings.test.tsx [hunk 1]
```diff
@@ -0,0 +1,93 @@
...
+
+  test("persists an old avatar without a background color exactly once", async () => {
+    const {backgroundColor: _backgroundColor, ...oldAvatar} = completeAvatar;
+    const {auth, store} = renderAvatarSettings(oldAvatar as AvataaarProps);
+
+    await waitFor(() => expect(API.editUser).toHaveBeenCalledTimes(1));
+    const migratedAvatar = {...oldAvatar, backgroundColor: completeAvatar.backgroundColor};
+    expect(API.editUser).toHaveBeenCalledWith({...auth, avatar: migratedAvatar});
+    expect(store.getState().auth.user?.avatar).toEqual(migratedAvatar);
+    expect(API.editUser).toHaveBeenCalledTimes(1);
```
Schwehn42: is the duplicate `toHaveBeenCalled` check necessary?

almacbe: Good point. The final assertion is redundant because the preceding waitFor already verifies exactly one call after the optimistic update. I removed the equivalent duplicates from the neighboring tests as well.

Schwehn42 (approved):
LGTM, thanks!

PR #6521, opened 2026-08-18
fix: keep banned participants visible after reload, auto-rejoin on unban

jmoehler:
As we didn't hear back from https://github.com/Phai2o on https://github.com/inovex/scrumlr.io/pull/6127, we took it in our own hands to finish the mr.

This is now the updated mr, containing the original changes puls updated snapshots and rebase on current main.


jmoehler:
@jmoehler: once this is approved, close https://github.com/inovex/scrumlr.io/pull/6127

src/components/RejectionPage/RejectionPage.tsx
```diff
@@ -28,15 +80,23 @@ export const RejectionPage = ({status}: RejectionPageProps) => {
...
-              window.location.pathname = "/";
-            }}
-          >
-            {t("RejectionPage.button")}
-          </button>
+          <div className="rejection-page__description">{isBanned ? t("RejectionPage.banned") : t("RejectionPage.description")}</div>
+          {isBanned && <div className="rejection-page__hint">{t("RejectionPage.bannedRetryHint")}</div>}
+          <div className="rejection-page__button-group">
+            {isBanned && (
+              <button className="rejection-page__return-button rejection-page__return-button--retry" onClick={tryRejoin} disabled={isRetrying}>
```
Schwehn42: I think we should use our Button component here instead of relying on custom styles

src/i18n/de/translation.json
```diff
@@ -860,6 +860,9 @@
     "title": "Ups!",
     "description": "Es sieht so aus, als ob die Türen zu dieser Sitzung vorerst geschlossen bleiben. Wir freuen uns über dein Interesse, aber leider wurde dein Antrag auf Teilnahme abgelehnt.",
     "banned": "Es tut uns Leid, aber es sieht so aus, als ob du von dieser Sitzung ausgeschlossen wurdest.",
+    "bannedRetryHint": "Falls du wieder zugelassen wurdest, kannst du erneut beitreten. Wir prüfen das auch automatisch alle 30 Sekunden.",
```
Schwehn42: might as well pass the 30 as interpolation variable, because why not

jmoehler: done

src/components/RejectionPage/RejectionPage.tsx
```diff
@@ -11,8 +17,72 @@ type RejectionPageProps = {
...
+    return {count: intervalMs / HOUR_IN_MS, unit: "hours" as const};
+  }
+
+  if (intervalMs % MINUTE_IN_MS === 0) {
+    return {count: intervalMs / MINUTE_IN_MS, unit: "minutes" as const};
+  }
+
+  return {count: Math.max(1, Math.round(intervalMs / SECOND_IN_MS)), unit: "seconds" as const};
+};
+
```
Schwehn42, line 21: i get that you are trying to generalize this but is this really necessary? and even if we needed it we have a datetime util helper

Schwehn42: I mean I get why you did it but I just meant we could pass the 30 as an interpolation variable to i18n to not have magic constants in the translation files, not as a sophisticated time parser:D

jmoehler: so rolling it back to just pass a number to i18n?

Schwehn42: yes please

jmoehler: done

src/api/participant.ts
```diff
@@ -106,8 +106,9 @@ export const ParticipantsAPI = {
     // - user tried to join a board where they are currently banned
     // TODO: doing this with proper error codes would be better, especially the check for the banned status
     if (response.status === 403 || response.status === 400) {
-      const body = (await response.json()) as {status: string; error: string};
-      if (body.error === "participant is currently banned from this session") {
+      const body = (await response.json()) as {status?: string; error?: string};
+      const errorMessage = body.error?.toLowerCase() ?? "";
+      if (errorMessage.includes("participant is currently banned from this session") || errorMessage.includes("banned from this session")) {
```
Schwehn42, line 111: looks like the second term `errorMessage.includes("banned from this session")` should be sufficient  since the first term implies the second

src/api/__tests__/participant.test.ts
```diff
@@ -0,0 +1,56 @@
+import {vi} from "vitest";
+import * as config from "config";
+import {ParticipantsAPI} from "../participant";
+
+describe("ParticipantsAPI", () => {
+  beforeEach(() => {
+    vi.spyOn(config, "SERVER_HTTP_URL", "get").mockReturnValue("http://localhost:8080");
```
Schwehn42, line 7: since you stub the fetch itself the `beforeEach` block is not needed.

I guess the alternative would be to use [mock service workers](https://github.com/mswjs/msw) which is a while new layer but would give us the ability to mock the whole api to our liking. we could discuss this though it wouldnt be without effort

src/api/participant.ts
jmoehler: true :)

PR #6522, opened 2026-08-18
fix: prevent joining previous board after logout

### LINKED ISSUE #6058: Autojoining board on login
### The problem

We have a feature that redirects the user to the last path they were on before logging in.

The problem is that when you log out and then create a new user, you are automatically redirected to the previous board and thus automatically join it.

### Browser

Firefox

Steps to reproduce:
1. Create user
2. Create board
3. Log out of the user
4. Instantly create new user
5. Get redirected to old board

### Screenshots

_No response_

### Additional context

_No response_

almacbe:
## Description

Fixes #6058.

Logout navigation previously happened before the server-side logout request had completed. This allowed a new authentication to overlap with the previous session cleanup and reuse navigation state associated with the previous user's board.

The logout flow now waits for the server response and performs a hard navigation to the login page. The authenticated state is preserved until that navigation happens. A login without an explicit destination now redirects to the template page and replaces the login entry in browser history.

## Changelog

- Wait for the logout request before navigating.
- Centralize post-logout navigation in the authentication thunk.
- Preserve authentication state until the hard navigation.
- Redirect logins without an explicit origin to the template page.
- Add unit coverage for the asynchronous logout flow.
- Add an end-to-end regression test with two consecutive anonymous users.

## Checklist

- [ ] The light- and dark-theme are both supported and tested (not applicable: no visual changes)
- [ ] The design was implemented and is responsive for all devices and screen sizes (not applicable: no visual changes)
- [ ] The application was tested in the most commonly used browsers (tested locally and with Cypress Electron)


src/store/features/auth/__tests__/thunks.test.ts
jmoehler, line 1: Please add a test to check the failure path

jmoehler (dismissed):
lgtm, only the one test missing. 

@Schwehn42, could you take over from here? thx

src/components/SettingsDialog/SettingsDialog.tsx
```diff
@@ -115,7 +115,6 @@ export const SettingsDialog = (props: SettingsDialogProps) => {
 
   const handleLogout = () => {
     dispatch(signOut());
-    navigate("/login");
```
Schwehn42, line 119: if the problem is the navigation happening before the logout can complete, can't we just wrap the navigate in a `.then` callback of the dispatch or am I missing something?

src/routes/LoginBoard/LoginBoard.tsx
```diff
@@ -45,12 +47,12 @@ export const LoginBoard = () => {
...
   const [showAnonymousContent, setShowAnonymousContent] = useState(false);
 
-  const redirectPath = (location.state as LocationState)?.from?.pathname || "/";
+  const redirectPath = (location.state as LocationState)?.from?.pathname || DEFAULT_REDIRECT_PATH;
 
   async function handleAnonymousLogin() {
     if (!displayName) return; // Basic safety check
     await Auth.signInAnonymously(displayName);
-    navigate(redirectPath);
+    navigate(redirectPath, {replace: true});
```
Schwehn42, line 55: just noting that this code must be very old, usually we would never call Auth directly but a dispatch thunk with callback

src/store/features/auth/reducer.ts
```diff
@@ -12,9 +12,6 @@ export const authReducer = createReducer(initialState, (builder) =>
       // note that directly reassigning the whole state won't work though!
       state.user = action.payload;
     })
-    .addCase(signOut.fulfilled, (state) => {
-      state.user = undefined;
-    })
```
Schwehn42, line 17: explain why isn't this needed anymore?

src/routes/LoginBoard/LoginBoard.tsx
almacbe: Makes sense. Refactoring this to use a thunk should go into a separate PR. I’ll keep it as it is in this one.

src/store/features/auth/reducer.ts
almacbe: This is no longer needed because `window.location.replace("/login")` performs a full-page navigation. Once it completes, the current application and its in-memory Redux store are discarded.

If we clear the user in `signOut.fulfilled`, `RequireAuthentication` could render before the browser finishes navigating, while we are still on the board route. It would then redirect to `/login` and save the current board in `location.state.from`, which could send the next user back to the previous board.

Therefore, we keep the user in Redux until the hard redirect completes.

src/store/features/auth/__tests__/thunks.test.ts
almacbe: Added the failure-path test, covering the rejected action, no redirect, preserved user state, and retry toast.

src/components/SettingsDialog/SettingsDialog.tsx
almacbe: A `.then()` would work for the first attempt, but not for the retry.

`createAsyncThunk` resolves the dispatch promise for both success and failure, so an unconditional `.then()` would also navigate after an error unless we use `.unwrap()` or check for `fulfilled`.

More importantly, the retry button creates a new `dispatch(signOut())`. A callback attached in `SettingsDialog` only belongs to the first attempt, so it would not run after a successful retry.

I added a [test for this exact flow](https://github.com/inovex/scrumlr.io/pull/6522/commits/07eee092d595c8b2d67109fd7634c8272a838b85): the first request fails without redirecting, the toast retries with a new `signOut()`, and the successful retry performs the redirect.

Schwehn42 (approved):
lgtm

PR #6524, opened 2026-08-18
feat: add loading indicator to history and templates

Schwehn42:
## Description

Users with many templates or especially boards don't get a visual indicator while they are being fetched (not having pagination doesn't help the cause).  
In fact, on the history page, the user might falsely assume they have no boards at all since the empty boards text is being displayed.

This PR not only adds a simple loading indicator to both Templates and History, but also uses a small hook to prevent flickering when the delay is marginal.

## Changelog

- `Templates`, `History`: add loading indicator while thunk fetch hasn't completed yet
- add new hook `useDelayedLoading`, which essentially debounces the loading value so only after a certain loading time (200ms as of now) the indicator is actually rendered.

## Checklist


Planlos5000 (approved):
LGTM

PR #6544, opened 2026-08-24
Refactor (SonarQube): constants instead of literals

### LINKED ISSUE #6549: SQ: define constants instead of string literals
Referring to issue #6099 , this issue is dedicated to the replacement of duplicated string literals by constants as pointed out by SonarQube analysis.

mating02:
Closes #6549 

## Description
Made refactoring changes to the codebase where necessary according to SonarQube analysis. For this Pull Request, these changes consist of replacing duplicated string literals with constants across different files.

## Changelog
- replaced repeating string literals with constants

## Checklist


server/e2e-tests/client.go
```diff
@@ -20,6 +20,13 @@ import (
 	"scrumlr.io/server/votings"
 )
 
+const boardsPath = "/boards"
```
Planlos5000: The `boardsPath` variable is now only used in the `boardpath` function. Why then make it a const?

server/e2e-tests/client.go
```diff
@@ -20,6 +20,13 @@ import (
...
+const votesPath = "/votes"
+
+func boardPath(id uuid.UUID, dir string) string {
```
Planlos5000: I would argue that there is a better name for this function

server/e2e-tests/client.go
```diff
@@ -20,6 +20,13 @@ import (
 	"scrumlr.io/server/votings"
 )
 
+const boardsPath = "/boards"
+const votesPath = "/votes"
```
Planlos5000: `votesPath` is used 3 times. I think this okay to *not* make it a const

server/src/api/boards.go
```diff
@@ -27,6 +27,7 @@ import (
 	"scrumlr.io/server/logger"
 )
 
+const decodeFailureMessageBoard = "failed to decode body"
```
Planlos5000: Define the constant only once for the package and not in every file.
This string is present in more that the 3 files that are currently touched (`boards.go`, `notes.go`, `votes.go`). I think we should also replace them in the other files

server/src/boards/service.go
```diff
@@ -33,6 +33,12 @@ import (
 	"scrumlr.io/server/votings"
 )
 
+const getBoardFailureMessage = "failed to get board"
+const getColumnsFailureMessage = "failed to get columns"
+const getNotesFailureMessage = "failed to get notes"
+const passphraseByPassphraseMessage = "passphrase should not be set for policies except 'BY_PASSPHRASE'"
```
Planlos5000: I would argue that there is a better name for this variable

server/src/sessions/api.go
```diff
@@ -16,6 +16,9 @@ import (
 	"scrumlr.io/server/logger"
 )
 
+const failurePassingUserIdMessage = "failed to parse user id"
```
Planlos5000: `parse` and `passing` are two different things. This should be `failureParsingUserIdMessage`

server/src/main.go
```diff
@@ -25,6 +25,26 @@ import (
...
+const authAzureADTenantIDFlag = "auth-azure-ad-tenant-id"
+const authAzureADClientIDFlag = "auth-azure-ad-client-id"
+const authAzureADClientSecretFlag = "auth-azure-ad-client-secret"
+const authAppleClientIDFlag = "auth-apple-client-id"
+const authAppleClientSecretFlag = "auth-apple-client-secret"
+const authOIDCClientIDFlag = "auth-oidc-client-id"
+const authOIDCClientSecretFlag = "auth-oidc-client-secret"
+const authOIDCDiscoveryURLFlag = "auth-oidc-discovery-url"
+const authOIDCUserIdentScopeFlag = "auth-oidc-user-ident-scope"
+const authOIDCUserNameScopeFlag = "auth-oidc-user-name-scope"
```
Planlos5000: I don't think that this is a good idea, because these names are used for the flags that can be added to the go command when scrumlr is started. For that reason i think it would be better, when the strings are directly where the rest of the flag is defined.
And if we want to define the strings as const i would say either all or no

Planlos5000 (changes requested):

server/e2e-tests/client.go
mating02: actually it also appears two times outside that function, one time in CreateBoard() and another time in GetBoards()

server/e2e-tests/client.go
mating02: yeah but 3 times is enough for sonarcube to complain hahah

server/src/sessions/api.go
mating02: true, was a spelling mistake

Planlos5000 (approved):

PR #6551, opened 2026-08-25
refactor (SQ): refactored joinBoard function using subfunctions

### LINKED ISSUE #6550: SQ: refactor joinBoard method to reduce cognitive complexity
the joinBoard method in server/src/api/boards.go is too complex and thus needs to be refactored.

mating02:
Closes #6550 

## Description
Cognitive Complexity of the joinBoard function was too high, so I reduced it by using dedicated subfunctions and eliminating unnecessary else blocks and code repetition

## Changelog
refactored joinBoard function in server/src/api/boards.go

## Checklist

### DIFF
--- a/server/src/api/boards.go
+++ b/server/src/api/boards.go
@@ -1,12 +1,16 @@
 package api
 
 import (
+	"context"
 	"encoding/csv"
 	"errors"
 	"fmt"
 	"net/http"
 
 	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/zap"
+	"scrumlr.io/server/columns"
 	"scrumlr.io/server/hash"
 	"scrumlr.io/server/otel"
 	"scrumlr.io/server/role"
@@ -245,6 +249,7 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
 		}
 
 		http.Redirect(w, r, s.buildRelativeURL(fmt.Sprintf(boardParticipantsPath, board, user)), http.StatusSeeOther)
+		s.handleExistingSession(w, r, ctx, span, board, user)
 		return
 	}
 
@@ -262,9 +267,24 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
 			common.Throw(w, r, mapError(err))
 			return
 		}
+	switch b.AccessPolicy {
+	case boards.Public:
+		s.joinPublicBoard(w, r, ctx, span, board, user)
+	case boards.ByPassphrase:
+		s.joinBoardByPassphrase(w, r, ctx, span, log, board, user, b)
+	case boards.ByInvite:
+		s.joinBoardByInvite(w, ctx, span, board, user)
+	default:
+		w.WriteHeader(http.StatusBadRequest)
+	}
+}
 
-		w.Header().Set("Location", s.buildRelativeURL(fmt.Sprintf(boardParticipantsPath, board, user)))
-		w.WriteHeader(http.StatusCreated)
+func (s *Server) handleExistingSession(w http.ResponseWriter, r *http.Request, ctx context.Context, span trace.Span, board, user uuid.UUID) {
+	banned, err := s.sessions.IsParticipantBanned(ctx, board, user)
+	if err != nil {
+		span.SetStatus(codes.Error, "failed to check if participant is banned")
+		span.RecordError(err)
+		common.Throw(w, r, mapError(err))
 		return
 	}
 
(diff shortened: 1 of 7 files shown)


server/src/api/boards.go
```diff
@@ -269,91 +256,101 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
...
+	case boards.ByInvite:
+		s.joinBoardByInvite(w, ctx, span, board, user)
+	default:
+		w.WriteHeader(http.StatusBadRequest)
+	}
+}
 
-		w.Header().Set("Location", s.buildRelativeURL(fmt.Sprintf(boardParticipantsPath, board, user)))
-		w.WriteHeader(http.StatusCreated)
+func (s *Server) handleExistingSession(w http.ResponseWriter, r *http.Request, ctx context.Context, span trace.Span, board, user uuid.UUID) {
```
Planlos5000: Do not pass the span. Create a new span with the context.
The context should be the first paramter of a fuction.
Do we need to pass the request and the writer here? Would it be possible to return an error and handle the error in the calling function?

server/src/api/boards.go
```diff
@@ -269,91 +256,101 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
...
-		encodedPassphrase := hash.NewHashSha512().HashBySalt(body.Passphrase, *b.Salt)
-		if encodedPassphrase == *b.Passphrase {
-			_, err := s.sessions.Create(ctx, sessions.BoardSessionCreateRequest{Board: board, User: user, Role: role.ParticipantRole})
-			if err != nil {
-				span.SetStatus(codes.Error, "failed to create session")
-				span.RecordError(err)
-				common.Throw(w, r, mapError(err))
-				return
-			}
+func (s *Server) joinPublicBoard(w http.ResponseWriter, r *http.Request, ctx context.Context, span trace.Span, board, user uuid.UUID) {
```
Planlos5000: context should be the first paramater
don't pass the span as an argument. create a new one
Do we need the http request and the writer here or can we just return an error and handle the it in the calling function

server/src/api/boards.go
```diff
@@ -269,91 +256,101 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
...
+}
 
-		} else {
-			err := errors.New("wrong passphrase")
-			span.SetStatus(codes.Error, "wrong passphrase provided")
-			span.RecordError(err)
-			common.Throw(w, r, common.BadRequestError(err))
-			return
-		}
+func (s *Server) joinBoardByPassphrase(w http.ResponseWriter, r *http.Request, ctx context.Context, span trace.Span, log *zap.SugaredLogger, board, user uuid.UUID, b *boards.Board) {
```
Planlos5000: context should be the first paramater
dont pass the logger and span. create a new one if needed
Do we need the http request and writer or can we throw an error and handle it in the calling function?

server/src/api/boards.go
```diff
@@ -269,91 +256,101 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
...
+		common.Throw(w, r, common.BadRequestError(err))
+		return
+	}
 
-		_, err = s.sessionRequests.Create(ctx, board, user)
-		if err != nil {
+	s.joinPublicBoard(w, r, ctx, span, board, user)
+}
+
+func (s *Server) joinBoardByInvite(w http.ResponseWriter, ctx context.Context, span trace.Span, board, user uuid.UUID) {
```
Planlos5000: context should be the first paramater
don't pass the span as an argument. create a new one
Do we need the http request and the writer here or can we just return an error and handle the it in the calling function

Planlos5000 (changes requested):

server/src/api/boards.go
```diff
@@ -268,92 +265,25 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
 		common.Throw(w, r, mapError(err))
 		return
 	}
```
Planlos5000, line 267: I would say we can also move this part into the join method

server/src/api/boards.go
Planlos5000: Could you rename the `board` variable to `boardId` and then later the `b` variable to `board`. I would think this would be a better naming

server/src/api/boards.go
```diff
@@ -227,117 +224,36 @@ func (s *Server) joinBoard(w http.ResponseWriter, r *http.Request) {
...
+	var joinRequest boards.JoinBoardRequest
 	if b.AccessPolicy == boards.ByPassphrase {
-		var body boards.JoinBoardRequest
-		err := render.Decode(r, &body)
-		if err != nil {
-			otel.RecordErrorSpan(span, err, new("failed to decode body"))
-			log.Errorw("Unable to decode body", "err", err)
+		if err := render.Decode(r, &joinRequest); err != nil {
+			span.SetStatus(codes.Error, "failed to decode body")
+			span.RecordError(err)
```
Planlos5000: Use the `otel.recordErrorSpan` method

Planlos5000 (approved):

PR #6574, opened 2026-08-28
refactor: open telemetry

### LINKED ISSUE #6569: Update otel configuration and usage
Currently the otel configuration happens in the initialize package. We want an own package for otel and with that pacage some helper functions e.g. to set and record errors to otel

Planlos5000:
## Description

This PR adds a new otel package that has the otel setup. This will also add a method to record an error.
The PR also adds tests for the otel setup and the new method to record errors.
## Changelog

- Add otel package
- Add tests for otel package
- Use the `RecordErrorSpan` method

## Checklist


server/src/api/notes.go
```diff
@@ -90,8 +88,7 @@ func (s *Server) getNote(w http.ResponseWriter, r *http.Request) {
 
 	note, err := s.notes.Get(ctx, id)
 	if err != nil {
-		span.SetStatus(codes.Error, "failed to get node")
-		span.RecordError(err)
+		otel.RecordErrorSpan(span, err, new("failed to get node"))
```
mating02: i just realized, i think it should be note here instead of "node"

server/src/boards/service.go
```diff
@@ -409,22 +384,19 @@ func (service *Service) Update(ctx context.Context, body BoardUpdateRequest) (*B
 		case ByInvite, Public:
 			if body.Passphrase != nil {
 				err := errors.New("passphrase should not be set for policies except 'BY_PASSPHRASE'")
-				span.SetStatus(codes.Error, "passphrase should not be set for policies except 'BY_PASSPHRASE'")
-				span.RecordError(err)
+				otel.RecordErrorSpan(span, err, nil)
```
mating02, line 387: nice solution, directly using the error message

server/src/otel/logging_provider.go
```diff
@@ -0,0 +1,50 @@
...
+
+func newGrpcLoggerProvider(ctx context.Context, grpcConnection *grpc.ClientConn) (*log.LoggerProvider, error) {
+	logExporter, err := otlploggrpc.New(ctx, otlploggrpc.WithGRPCConn(grpcConnection))
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := newResource()
+	if err != nil {
+		return nil, err
```
mating02, line 20: In this case, documentation suggests using the Shutdown(ctx) function in case the Exporter is no longer needed, which is true for this case because resource initialization failed but the exporter is still created and running 
you could also move the newResource line to the top to achieve that i think

server/src/otel/logging_provider.go
```diff
@@ -0,0 +1,50 @@
...
+
+func newHttpLoggerProvider(ctx context.Context, httpEndpoint string) (*log.LoggerProvider, error) {
+	logExporter, err := otlploghttp.New(ctx, otlploghttp.WithEndpoint(httpEndpoint))
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := newResource()
+	if err != nil {
+		return nil, err
```
mating02, line 41: also here

server/src/otel/logging_provider.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	loggerProvider := log.NewLoggerProvider(
+		log.WithResource(res),
+		log.WithProcessor(
+			log.NewBatchProcessor(logExporter),
+		),
+	)
+	return loggerProvider, nil
```
mating02, line 30: also, the lines starting from res, err declaration are exactly the same for both functions -> code duplication
Do you think it's better if we transfer those lines to a single function which is called by both of the functions here?

server/src/otel/meter_provider.go
mating02, line 1: the same things i commented about on logging_provider.go apply here as well

server/src/otel/otel.go
```diff
@@ -0,0 +1,153 @@
...
+	)
+}
+
+func newResource() (*resource.Resource, error) {
+	return resource.Merge(
+		resource.Default(),
+		resource.NewWithAttributes(
+			semconv.SchemaURL,
+			semconv.ServiceName("scrumlr"),
+			semconv.ServiceVersion("5.3.1"),
```
mating02, line 151: hardcoding version numbers will be hard to maintain tho, is there maybe a better solution to this?

server/src/otel/otel.go
```diff
@@ -0,0 +1,153 @@
...
+	propagator := newPropagator()
+	otel.SetTextMapPropagator(propagator)
+
+	if options.grpcEndpoint != nil {
+		grpcConnection, ex := initGrpcConnection(*options.grpcEndpoint)
+		if ex != nil {
+			handleErr(ex)
+			return
+		}
+
```
mating02, line 53: why isn't grpcConnection.Close() appended to the shutdown functions?

server/src/otel/trace_provider.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
+	"go.opentelemetry.io/otel/sdk/trace"
+	"google.golang.org/grpc"
+)
+
+func newGrpcTraceProvider(ctx context.Context, connection *grpc.ClientConn) (*trace.TracerProvider, error) {
+	traceExporter, err := otlptracegrpc.New(ctx, otlptracegrpc.WithGRPCConn(connection))
```
mating02: again here, either shutdown or newResource() first

mating02 (changes requested):

server/src/otel/otel.go
Planlos5000: Because it is not possibel. The shutdown functions have the signature `Shutdown(ctx context.Context)` and the close function from the grpc connection has the signature `Close()`.

mating02 (approved):

PR #6582, opened 2026-08-31
fix: server panic in user delete api

Planlos5000:
## Description

This pr fixes a bug in the middleware for the user delete endpoint and adds some tests for the middleware and the delete endpoint.
When a delete request was made with a valid token and a different valid id in the url the server panics because the error that is recorded in the span is nil.
## Changelog

- Fix nil pointer reference
- Add tests for the `isAccountOwner` middleware
- Add tests for the delete api

## Checklist

### DIFF
--- a/server/src/users/api_test.go
+++ b/server/src/users/api_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"testing"
@@ -19,7 +20,7 @@ import (
 	"scrumlr.io/server/technical_helper"
 )
 
-func TestGetUser_api(t *testing.T) {
+func TestApiGetUser(t *testing.T) {
 	userId := uuid.New()
 
 	mockUserService := NewMockUserService(t)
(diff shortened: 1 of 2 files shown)


server/src/users/api_test.go
```diff
@@ -426,3 +475,52 @@ func Test_AnonymousCustomTemplateCreationContext_NotAllowed(t *testing.T) {
 	assert.Equal(t, http.StatusForbidden, rr.Result().StatusCode)
 	assert.Error(t, common.ForbiddenError(errors.New("not authorized to create custom templates anonymous")))
 }
+
+func Test_isAccountOwner(t *testing.T) {
```
mateo-ivc: I would prefer to maintain a single naming scheme

server/src/users/api_test.go
```diff
@@ -426,3 +475,52 @@ func Test_AnonymousCustomTemplateCreationContext_NotAllowed(t *testing.T) {
...
+	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+
+	userApi.isAccountOwner(next).ServeHTTP(rr, req.Request())
+
+	assert.Equal(t, http.StatusOK, rr.Result().StatusCode)
+}
+
+func Test_isAccountOwner_differentIds(t *testing.T) {
```
mateo-ivc: same as above

mateo-ivc (approved):

PR #6606, opened 2026-09-01
feat: Add rate limit environment variables

Planlos5000:
## Description

This PR adds environment variables to adjust the rate limits for the backend api.
It adds the env vars

- SCRUMLR_JOIN_RATE_LIMIT
- SCRUMLR_TEMPLATE_RATE_LIMIT

## Changelog

- Add rate limit env vars
- Documnet env vars
## Checklist

- [x] I have performed a self-review of my own code
- [x] I have written and understand every part of this contribution myself - if AI tools were used, I have thoroughly reviewed and verified all changes
- [x] I have commented my code, particularly in hard-to-understand areas

PR #6607, opened 2026-09-01
test: more cypress tests

Schwehn42:
## Description

adds various tests to cover our bases for some more of the most common interactions with scrumlr:

### Board
- create board
- mark as done
- raise hand
- add board reaction
- start a timer
- start a voting

### Columns
- create a column
- delete a column
- edit a column (its color) 

### Notes
- create a note
- edit a note
- delete a note
- react on a note

## Changelog

- adds the above mentioned cypress suites and tests
- adds commands for some repeated actions
- various class name and test id changes and additions to make element selection easier
- update snapshots due to above changes
- increase `SCRUMLR_JOIN_RATE_LIMIT` (see #6606) for the server compose deployment which the cypress action uses, and due to its rapid execution of tests requires a higher limit

## Checklist


cypress/support/commands.ts
```diff
@@ -89,3 +102,16 @@ Cypress.Commands.add("selectMiniMenu", (cyData: string, itemLabel: string)=>{
...
+
+// create a note and return the subject
+Cypress.Commands.add("createNote", (text: string)=>{
+  cy
+    .get<HTMLInputElement>(".note-input__input")
+    .first()
+    .type(text)
+    .press("Enter")
+
+  // note: since notes pop up on top, we assume the first note we grab is the one we just created
```
jmoehler, line 114: this sounds like create note is not deterministic. should we perhaps fix that?

cypress/e2e/board-spec.cy.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  it("should create board from the initial setup", () => {
+    cy
+      .url()
+      .should("not.include", "/boards/templates") // not on templates anymore
+      .should("include", "/board")
+  })
+
+  /*
+  TODO: add tests for different board use cases:
```
jmoehler, line 16: is it intentional that these tests were not added yet?

cypress/e2e/notes-spec.cy.ts
```diff
@@ -0,0 +1,63 @@
+/*
+* TODO: add tests for different notes use cases:
```
jmoehler, line 2: is it intentional that these tests were not added yet?

cypress/e2e/board-spec.cy.ts
Schwehn42: yes, I just wanted to get the most important features out of the way first. Though you could argue that logout is tested in `logout-spec.cy.ts`.

cypress/e2e/notes-spec.cy.ts
Schwehn42: yes, similar to board not the complete feature list is being tested atm, as stacking (drag and drop) is especially tricky to do. But we should strive to test all major features one day

cypress/support/commands.ts
Schwehn42: I'll try

cypress/e2e/board-spec.cy.ts
jmoehler: ok so not a blocker for the mr I think, was just confused :)

cypress/e2e/notes-spec.cy.ts
jmoehler: see above

cypress/support/commands.ts
jmoehler: perfect, do you want to try in this mr or in a separate? perhaps it deserves its own branch!? If so I would approve this one. What do you think?

Schwehn42: agreed

PR #6642, opened 2026-09-08
refactor: note image url check

### LINKED ISSUE #6598: image fetch is executed on each note input change
### The problem

https://github.com/inovex/scrumlr.io/pull/6529 changed some logic concerning the `isImageUrl` helper util.

this leads to currently sending a fetch request on every note input change, where most will obviously bounce back failing.

Note: High priority because this definitely needs to get fixed before the next release, or alternatively revert https://github.com/inovex/scrumlr.io/pull/6529 first

 

### Browser

All

- start typing a note in the input (e.g. "abc")
- monitor network request
- result: three separate network requests for `https://a`, `https://ab`, `https://abc`

### Screenshots

_No response_

### Additional context

I think there are some strategies than can be employed to mitigate and minimize the amount of network requests being called:

1. fetch only if the input is a valid URL, 
2. debounce fetch (if the user hasn't changed the input for, say, 300ms)
3. don't get the whole image just to read the mime type, use [HTTP HEAD request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD)
4. rethink whether this fetch is needed in the first place or the url itself is sufficient, but I reckon not all image urls always have an adequate extension

Schwehn42:
## Description

fixes https://github.com/inovex/scrumlr.io/issues/6598

currently, fetches are needlessly executed just to check whether a url is a valid image.  
For example, simply putting "a" into a note would check if "https://a" is an image, where we can confidently say it won't.

this PR does various improvements:

1. debounce: don't check while the user is typing, but after a small delay
2. AbortController: fetch allows for signal to cancel the operation, e.g. if the user start typing again
3. optimistic check: well-formed with image suffix (e.g. .png) are considered valid images
4. fallback fetch: for url with no extension, such as https://avatars.githubusercontent.com/u/9919, a fetch is actually executed
   1. only HEADers are fetched
   2. has to be 200 ok
   3. has to be image mime type

## Changelog

- `useImageChecker`
  - add debounce 
  - add abort controller signal
- `utils/url`
  - `normalizeAndParseUrl`: try to create a well-formed URL object
  - isValidWebUrl: checks if a url if valid in the sense of a web url, e.g. https://a is a valid url but not a web url
- `utils/images`
  - `isImageUrl`: parse and check for image url, improve fetch
- add unit tests 

## Checklist


Planlos5000 (approved):
lgtm

PR #6653, opened 2026-09-09
fix: Hide settings for temporary board columns

### LINKED ISSUE #5359: Issues when having multiple temporary columns
Adding a column from a column that is also temporary causes some issues.

Steps to reproduce:
1. Create a new board
2. Create a new column and write something for the title but without submitting it
3. For the temporary column, open the actions menu and create another new column
4. Focus the title field of the first added column again
5. Enter some name and click on the submit button
6. Observe the bug

mvanhorn:
## Description

`renderSettings` in `ColumnDetails.tsx` already returned `null` for non-moderators. This extends that guard to temporary columns:

```ts
if (!isModerator || props.isTemporary) return null;
```

A temporary column is one the user has started but not yet saved. Its settings menu contains "add column", so opening it on an unsaved column creates a second draft while the first is still unsaved, which is the nested-draft path in #5359. Hiding the menu until the column is persisted removes that path.

The draft editor and its Save/Cancel mini-menu are untouched, so an unsaved column can still be named, described, saved, or cancelled. Persisted columns keep their settings in both view and edit modes.

`isTemporary` already flows `Column` -> `ColumnHeader` -> `ColumnDetails`, so no new wiring was needed.

Fixes #5359

## Checklist

- [ ] The light- and dark-theme are both supported and tested (not run; this change removes a menu and adds no styling)
- [ ] The design was implemented and is responsive for all devices and screen sizes (not run)
- [ ] The application was tested in the most commonly used browsers (not run; verified via the unit suite only)

## Verification

`yarn test`: 84 files, 709 passed, 6 skipped. The `ColumnDetails` edit-mode snapshot is updated; the only delta is two React `useId` values (`_r_0_` -> `_r_j_`), which shifted because the removed settings subtree changed the id counter.

AI was used for assistance.

### DIFF
--- a/src/components/Column/ColumnDetails/ColumnDetails.tsx
+++ b/src/components/Column/ColumnDetails/ColumnDetails.tsx
@@ -298,7 +298,7 @@ export const ColumnDetails = (props: ColumnDetailsProps) => {
   const renderDescription = () => (props.mode === "edit" ? editableDescription() : viewableDescription());
 
   const renderSettings = () => {
-    if (!isModerator) return null;
+    if (!isModerator || props.isTemporary) return null;
 
     if (openSettings) {
       return (
--- a/src/components/Column/ColumnDetails/__tests__/ColumnDetails.test.tsx
+++ b/src/components/Column/ColumnDetails/__tests__/ColumnDetails.test.tsx
@@ -4,7 +4,9 @@ import {ParticipantWithUser, ParticipantRole} from "store/features";
 import getTestApplicationState from "utils/test/getTestApplicationState";
 import getTestStore from "utils/test/getTestStore";
 import {Provider} from "react-redux";
-import {fireEvent} from "@testing-library/react";
+import {fireEvent, waitFor} from "@testing-library/react";
+import {API} from "api";
+import {TEMPORARY_COLUMN_ID} from "constants/misc";
 
 describe("ColumnDetails", () => {
   const renderColumnDetails = (overrideProps?: Partial<ColumnDetailsProps>, userRole: ParticipantRole = "OWNER") => {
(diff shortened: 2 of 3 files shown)


Schwehn42 (changes requested):
hey Matt, appreciate the PR! 

looks pretty good to me all in all, as a minor nitpick it would be awesome if you could update the failing test due to a snapshot (`yarn run test --update`).

Thanks!

Planlos5000:
AI assisted is a little understatement
https://github.com/inovex/scrumlr.io/blob/main/CONTRIBUTING.md#use-of-ai-and-llm-tools

mvanhorn:
@Schwehn42 snapshot updated, thanks. The only delta is two React `useId` values in the edit-mode snapshot, which shifted because removing the settings subtree changed the id counter. Full suite is green: 84 files, 709 passed.

@Planlos5000 that's fair, and I've corrected the PR description. Three checklist boxes were ticked that shouldn't have been: cross-browser, responsive, and light/dark theme testing. I didn't do any of those. I've unticked them and said what I did run instead.

On the change itself, so you can judge it directly rather than take my word: `renderSettings` already early-returned `null` for non-moderators, and this adds `|| props.isTemporary` to that same guard. The settings menu is where "add column" lives, so opening it on an unsaved column is what produces a second draft while the first is still unsaved, which is the path in #5359's repro. The Save/Cancel mini-menu is a separate branch further down, so cancelling and saving an unsaved column both still work. `isTemporary` was already threaded through `Column` -> `ColumnHeader` -> `ColumnDetails`.

Happy to answer anything else about it, or to close this if you would rather handle it yourselves.

PR #6661, opened 2026-09-10
feat: scrumlr cli

### LINKED ISSUE #5637: Update database documentation
It might be beneficial to update the database.md documentation with instructions on 
1. How to run database migrations, rollback, etc.
2. How to inspect the database

### LINKED ISSUE #6655: Add cli tool to migrate database
Currently the database is migrated automatically with the start of the backend. This behaviour should stay as it is. This way does not support to apply migrations down.

To support migrations down we could write a little cli helper that lets you apply migrations either up or down.

Planlos5000:
## Description

Add two commands to the scrumlr cli to test the health of the server and to migrate the database.
The health check that was a separate binary before is now integrated into the scrumlr cli. It works like the health check before and can be use with the sub command `healthcheck`.
For database migrations there is now the sub command `database migrate` with the sub commands `up`, `down` and `version`. 
## Changelog

- Add health check as command
- Add database migration commands
- Remove health check binary
## Checklist


server/src/cmd/database/database.go
```diff
@@ -0,0 +1,222 @@
...
+					{
+						Name:      "version",
+						Usage:     "Get the current version of the database migration",
+						UsageText: "scrumlr database migrate version [global options]",
+						Action:    getVersion,
+						Category:  "migrations",
+					},
+					{
+						Name:      "up",
+						Usage:     "Apply all migrations up to the latest if to specified with the --version flag",
```
mating02, line 42: this sentence is grammatically incorrect. Better: Apply all migrations up to the latest, or up to the specified --version

server/src/cmd/database/database.go
```diff
@@ -0,0 +1,222 @@
...
+		if err != nil {
+			return err
+		}
+
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	_, err = fmt.Fprintf(cmd.Writer, "database at migration version %d, all migration applied successfully %t", version, !migrationFailed)
```
mating02, line 93: i think for the Fprintf lines, it is good to include a newline character "\n" at the end of the message string so it does not look weird in stdout

server/src/cmd/database/database.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func migrateUp(ctx context.Context, cmd *cli.Command) error {
+	m, err := getDatabase(ctx, cmd)
```
mating02, line 102: the migration instance and the db never get closed after the function which leads to a leak if the migration instance and the db are not intended to be left open.
I would suggest including  "defer m.Close()" here and in the other function to close both the database and the migration instance.

server/src/cmd/database/database_test.go
```diff
@@ -0,0 +1,587 @@
...
+	"fmt"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/urfave/cli/v3"
+	"scrumlr.io/server/initialize"
+)
+
+const currentVersion uint = 28
```
mating02, line 14: hardcoding the current version of the migration makes this test file hard to maintain. I would suggest using a function that gives you the current version using the file system (the following function is AI generated but I edited it and tested it and it should work for our structure nonetheless):
```

func getLatestMigrationVersion(t *testing.T) uint {
    // Read the embedded migrations directory
    entries, err := fs.ReadDir(initialize.Migrations, "migrations/sql")
    assert.NoError(t, err, "failed to read migrations directory")

    var maxVersion uint = 0

    for _, entry := range entries {
        // Skip subdirectories if any exist
        if entry.IsDir() {
            continue
        }
        name := entry.Name()
        parts := strings.SplitN(name, "_", 2)
        
        // Skip files that don't have an underscore
        if len(parts) < 2 {
            continue
        }

        // Convert the string prefix (e.g., "11") into a number
        version, err := strconv.ParseUint(parts[0], 10, 32)
        
        // If it successfully parsed as a number, and it's the highest one so far, save it
        if err == nil && uint(version) > maxVersion {
            maxVersion = uint(version)
        }
    }

    return maxVersion
}
```

Planlos5000: Since we don't write that many database migrations, i think it is okay to have the current number hardcoded. This is easier to understand than the code above. The tests will fail and show that the number should be adjusted.

Schwehn42: To chime in I do think he's got a point though, magic constants which need to be updated require the awareness by the PR author or reviewer. 

Which I agree is probably not asking too much, but maybe we could consider options to env-ify the current version by utilizing `migrate version` like here: https://stackoverflow.com/q/69068781 ?

Schwehn42: but then again if you say the tests will fail otherwise it just might be obvious enough :D

PR #6711, opened 2026-09-21
Refactor: Refactor template pkg api

### LINKED ISSUE #6656: Refactor: refactor template api packages
With the current state of the refactoring it should be possible to move the api for the board templates as well as the column templates to their packages.

- [ ] Move board temaplate api to the `boardtemplates` package
- [ ] Rewrite the api tests for the board template api
- [ ] Add tests for the board template router
- [ ] Move column template api to the `columntemplates` package
- [ ] Rewrite the api tests for the column templates api
- [ ] Add tests for the column template router

mating02:
Closes #6656 

## Description
The boardtemplate and columntemplate api endpoints were moved from the api package to their own respective package. The api tests were also rewritten and new tests were added for the router to ensure its functionality. Additionally, the mapError function was moved to the common package since it is needed for the api functions in the columntemplate and boardtemplate package and in my opinion this was the best solution when also considering that we want to get rid of the api package in general at some point.

## Changelog
- [x] Moved board temaplate api to the `boardtemplates` package
- [x] Rewrote the api tests for the board template api
- [x] Added tests for the board template router
- [x] Moved column template api to the `columntemplates` package
- [x] Rewrote the api tests for the column templates api
- [x] Added tests for the column template router

## Checklist


server/src/common/error.go
```diff
@@ -60,3 +63,31 @@ func Throw(w http.ResponseWriter, r *http.Request, err error) {
 	}
 	_ = render.Render(w, r, InternalServerError)
 }
+
+// mapError translates domain and database errors to HTTP API Errors.
+func MapError(err error) error {
```
Planlos5000, line 68: This does not help to get rid of the `common` package. The errors from the template packages should be mapped/handled in these packages.

server/src/api/context.go
```diff
@@ -398,6 +399,9 @@ func (s *Server) BoardTemplateRateLimiter(next http.Handler) http.Handler {
 func (s *Server) ColumnTemplateContext(next http.Handler) http.Handler {
```
Planlos5000, line 399: Also move the middleware/context function for the template apis to the packages

server/src/boardtemplates/api.go
```diff
@@ -13,3 +19,197 @@ type BoardTemplateService interface {
...
+	service BoardTemplateService
+}
+
+func NewBoardTemplateApi(service BoardTemplateService) BoardTemplateApi {
+	api := new(API)
+	api.service = service
+	return api
+}
+
+//var tracer trace.Tracer = otel.Tracer("scrumlr.io/server/api")
```
Planlos5000: Remove this commented tracer here

server/src/boardtemplates/api_test.go
```diff
@@ -0,0 +1,250 @@
...
+	} else {
+		bodyBytes, err := json.Marshal(body)
+		if err != nil {
+			panic(err)
+		}
+		bodyReader = bytes.NewReader(bodyBytes)
+	}
+
+	return httptest.NewRecorder(), technical_helper.NewTestRequestBuilder(method, path, bodyReader)
+}
```
Planlos5000, line 43: There is already a helper function for this in the `technical_helper` package

server/src/boardtemplates/router.go
```diff
@@ -0,0 +1,35 @@
...
+	router.Put("/{id}", r.boardTemplateAPI.UpdateBoardTemplate)
+	router.Delete("/{id}", r.boardTemplateAPI.DeleteBoardTemplate)
+	return router
+}
+
+func NewBoardTemplateRouter(boardTemplateApi BoardTemplateApi) *Router {
+	r := new(Router)
+	r.boardTemplateAPI = boardTemplateApi
+	return r
+}
```
Planlos5000: `New` function should be the first function in the file

server/src/columntemplates/api_test.go
```diff
@@ -0,0 +1,247 @@
...
+	} else {
+		bodyBytes, err := json.Marshal(body)
+		if err != nil {
+			panic(err)
+		}
+		bodyReader = bytes.NewReader(bodyBytes)
+	}
+
+	return httptest.NewRecorder(), technical_helper.NewTestRequestBuilder(method, path, bodyReader)
+}
```
Planlos5000: There is already a function in the `technical_helper` package for that

server/src/columntemplates/router.go
```diff
@@ -0,0 +1,35 @@
...
+	router.Put("/{id}", r.columnTemplateAPI.UpdateColumnTemplate)
+	router.Delete("/{id}", r.columnTemplateAPI.DeleteColumnTemplate)
+	return router
+}
+
+func NewColumnTemplateRouter(columnTemplateApi ColumnTemplateApi) *Router {
+	r := new(Router)
+	r.columnTemplateAPI = columnTemplateApi
+	return r
+}
```
Planlos5000: `New` function should be the first in the file

server/src/boardtemplates/router_test.go
Planlos5000, line 1: Wouldn't it be better for debugging purposes to write each test in its own function? Same for the column template router tests

server/src/common/error.go
mating02: but we need to have the MapError function in a shared package, otherwise the same MapError function has to be repeated over multiple packages. Maybe see this more of a temporary solution and if we get rid of the common package we can move that MapError function to a shared error package that every other package can import

Planlos5000: I am not a big fan of temporary solutions because they are here to stay

PR #6713, opened 2026-09-21
feat: added automatic version updating to workflows

### LINKED ISSUE #6712: feat: New Version Tag automation
Try around and see if there is a good way to automate new version tag creation.

Ideas: 
- search and replace script as part of the workflow
- global env variable that is passed down

mating02:
Closes #6712 

## Description
In order to not having to manually change the current app version in 12 different files, I have added a new workflow file to allow the update of every hardcoded app version string to be replaced with the new version that can be inputted before a new release.

## Changelog
- added new bump-version.yml file to enable automatic semver updates

## Checklist


.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Find and replace version strings
+        env:
+          NEW_VERSION: ${{ inputs.new_version }}
+        run: |
+          OLD_VERSION=$(jq -r '.version' package.json)
+
+          echo "Bumping version from $OLD_VERSION to $NEW_VERSION"
+
+          files=(
+            "package.json"
```
Planlos5000: This will not work with the `package.json` because there could be another package with the same version as scrumlr which would result in updating this package as well even if this version does not exist.

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,61 @@
...
+            "deployment/docker/.env.example"
+            "deployment/docker/docker-compose.yml"
+            "deployment/helm/scrumlr/Chart.yaml"
+            "deployment/helm/scrumlr/tests/backend/deployment_test.yaml"
+            "deployment/helm/scrumlr/tests/frontend/deployment_test.yaml"
+            "docs/yarn.lock"
+            "server/src/main.go"
+            "server/src/otel/otel.go"
+            "server/src/swagger/docs.go"
```
Planlos5000: The files in `server/src/swagger` are all generated except the `swagger.go` file and should not be modified. These files should be generated with the `make swagger` command or the underling commands.

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,61 @@
...
+          echo "Bumping version from $OLD_VERSION to $NEW_VERSION"
+
+          files=(
+            "package.json"
+            "deployment/docker/.env.example"
+            "deployment/docker/docker-compose.yml"
+            "deployment/helm/scrumlr/Chart.yaml"
+            "deployment/helm/scrumlr/tests/backend/deployment_test.yaml"
+            "deployment/helm/scrumlr/tests/frontend/deployment_test.yaml"
+            "docs/yarn.lock"
```
Planlos5000: Why is this here? This file has nothing to do with the release?

Planlos5000 (changes requested):

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,60 @@
...
+            "deployment/helm/scrumlr/tests/frontend/deployment_test.yaml"
+            "server/src/main.go"
+            "server/src/otel/otel.go"
+          )
+
+          for file in "${files[@]}"; do
+            sed -i "s/$OLD_VERSION/$NEW_VERSION/g" "$file"
+          done
+
+          cd server && make swagger
```
Planlos5000, line 45: This will also not work. This flow is executed on an ubuntu where no go and `swaggo/swag` is installed

mating02: do you think it is better then to remove the swagger part out of this workflow entirely and just do the make swagger command in the terminal?

